PR #12, opened 2016-08-31
Add an optional hash param to @EpoxyAttribute

gpeal:
You can now annotate @EpoxyAttribute(hash=false) to skip that attribute
for the equals and hash function. This is useful for fields that can
change on each bind but don't affect the underlying data such as a
click listener that gets created on each bind call.


gpeal:
@elihart

epoxy-processor/src/main/java/com/airbnb/epoxy/EpoxyProcessor.java
```diff
@@ -385,6 +390,10 @@ private MethodSpec buildHashCode(ClassToGenerateInfo helperClass) {
     }
 
     for (AttributeInfo attributeInfo : helperClass.getAttributeInfo()) {
+      if (!attributeInfo.useInHash()) {
```
elihart: You should also include this check in line 386 since the temp variable won't be needed if that attribute isn't included in the hash

elihart:
lgtm after the one comment is addressed

Issue #17, opened 2016-08-31
Don't let a model change its ID once it has been bound


elihart:
This can help prevent a tricky bug where a model id is changed without an accompanying notify call.

Additionally, this would allow us to optimize the diff algorithm since we can check models list size for changes to know if there were removals or additions. 

The downside is it would prevent reusing models in something like a model pool for memory optimization. We don't currently do that anywhere, and models are generally fairly small so unless somebody had thousands of them and recreated them often that shouldn't be a big issue.

The pros seem to outweigh the cons.

Thoughts @gpeal ?

elihart:
https://github.com/airbnb/epoxy/pull/19

PR #19, opened 2016-09-02
Don't allow a model id to change after it has been added to an adapter

elihart:
https://github.com/airbnb/epoxy/issues/17

@gpeal


gpeal:
Awesome! Let's hope this saves many hours of debugging in the future 👍

PR #22, opened 2016-09-07
Use correct layout in EpoxyModel hashcode

elihart:
@gpeal 

Before hashcode and equals were using the layout field only, but since the `getDefaultLayout` could change its return value we need to use `getLayout()` instead.


gpeal:
👍

PR #23, opened 2016-09-07
Fix differ test

elihart:
@gpeal

This test never worked properly since the sublist was cleared before its contents were added to the other list.


gpeal:
👍

PR #28, opened 2016-09-09
Support binding to a view holder

elihart:
@gpeal 

Let me know if you have better naming ideas. Still need to update the readme.

### DIFF
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyHolder.java
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyHolder.java
@@ -12,8 +12,7 @@ public abstract class EpoxyHolder {
    * opportunity to find views by id, and do any other initialization you need. This is called only
    * once for the lifetime of the class.
    *
-   * @param itemView The view this holder should hold. Created by inflating the layout provided by
-   *                 {@link EpoxyModelWithHolder#getLayout()}
+   * @param itemView A view inflated from the layout provided by {@link EpoxyModelWithHolder#getLayout()}
    */
   protected abstract void bindView(View itemView);
 }
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyHolder.java
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyHolder.java
@@ -1,0 +1,19 @@
+package com.airbnb.epoxy;
+
+import android.view.View;
+
+/**
+ * Used in conjunction with {@link com.airbnb.epoxy.EpoxyModelWithHolder} to provide a view holder
+ * pattern when binding to a model.
+ */
+public abstract class EpoxyHolder {
+  /**
+   * Called when this holder is created, with the view that it should hold. You can use this
+   * opportunity to find views by id, and do any other initialization you need. This is called only
+   * once for the lifetime of the class.
+   *
+   * @param itemView The view this holder should hold. Created by inflating the layout provided by
+   *                 {@link EpoxyModelWithHolder#getLayout()}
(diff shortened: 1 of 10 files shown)


epoxy-sample/src/main/java/com/airbnb/epoxy/models/BaseEpoxyHolder.java
```diff
@@ -0,0 +1,20 @@
...
+
+import android.support.annotation.CallSuper;
+import android.view.View;
+
+import com.airbnb.epoxy.EpoxyHolder;
+
+import butterknife.ButterKnife;
+
+/**
+ * Creating a base holder class allows us to leverage Butterknife's view binding for all subclasses.
```
mmdango: nit: ButterKnife

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyHolder.java
```diff
@@ -0,0 +1,19 @@
...
+ * Used in conjunction with {@link com.airbnb.epoxy.EpoxyModelWithHolder} to provide a view holder
+ * pattern when binding to a model.
+ */
+public abstract class EpoxyHolder {
+  /**
+   * Called when this holder is created, with the view that it should hold. You can use this
+   * opportunity to find views by id, and do any other initialization you need. This is called only
+   * once for the lifetime of the class.
+   *
+   * @param itemView The view this holder should hold. Created by inflating the layout provided by
```
gpeal: How about something like `The view inflated from EpoxyModelWithHolder#getLayout()`

elihart: good idea

elihart:
Updated, please take a look @mmdango @gpeal

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyViewHolder.java
```diff
@@ -20,21 +21,30 @@ public EpoxyViewHolder(ViewGroup parent, @LayoutRes int layoutId) {
...
 
+    if (epoxyHolder == null && model instanceof EpoxyModelWithHolder) {
+      epoxyHolder = ((EpoxyModelWithHolder) model).createNewHolder();
+      epoxyHolder.bindView(itemView);
+    }
+
     if (payloads.isEmpty()) {
       // noinspection unchecked
-      model.bind(itemView);
+      model.bind(objectToBind());
```
gpeal: Is there a reasonable way to clean up the rawtype EpoxyModel above and here?

elihart: No, unfortunately. As far as I know at least. It would mean typing the View holder with the view type, which isn't possible because the view types are all dynamic.

It doesn't really matter though since this is safe and not going to be the wrong type, and it's hidden in the library so the casting isn't exposed to the user.

gpeal: You could do it with an if instanceof check then call the correctly typed bind method with the view holder or view right?

elihart: Not sure we're on the same page, let's sync in person.

This is already guaranteed to be the correctly typed bind method.

gpeal: I know that this code would work, it would just be nice to have the type safety for clarity.

gpeal:
👍

PR #31, opened 2016-09-12
Remove object pools in differ

elihart:
I previously added object pools for the ModelState and UpdateOp objects which are used in the diffing algorithm.

A ModelState object is created for each model in the adapter, and an UpdateOp was created for each separate notify action. I did this to prevent memory thrashing, since if you have lots of models in your adapter and run the differ a lot it would create many of these objects each time.

I spent more time profiling this and found the pools to not be worth while. The objects are too small, and created too efficiently, to really make a difference pooling. In fact, the overhead of pooling slows down the diff quite a bit.

Removing the pools gives about a 14% speed improvement. It adds a little bit more memory overhead, but in tests with the sample app it seems like 10x more memory and object allocation is done just to handle the drawing and recycler view updates when a diff happens.

I seem to have prematurely optimized here. Removing the pools simplifies the code and will make it faster.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,7 +4,7 @@ buildscript {
     jcenter()
   }
   dependencies {
-    classpath 'com.android.tools.build:gradle:2.2.0-rc1'
+    classpath 'com.android.tools.build:gradle:2.2.0-rc2'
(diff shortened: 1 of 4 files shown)


gpeal:
👍

PR #32, opened 2016-09-12
Don't allow notifyDataSetChanged when diffing is enabled

elihart:
If diffing is enabled it will now throw if you call `notifyDataSetChanged`.

There's no reason to do `notifyDataSetChanged` if you are using diffing, and it is probably a bug/accident if you do since it may be easy to mix it up with `notifyModelsChanged`. 

This should prevent that and simplify the diffing code from having to handle the `notifyDataSetChanged` case

The notify tests had to be cleaned up to handle this

### DIFF
--- a/epoxy-sample/src/main/java/com/airbnb/epoxy/MainActivity.java
+++ b/epoxy-sample/src/main/java/com/airbnb/epoxy/MainActivity.java
@@ -38,6 +38,11 @@ public class MainActivity extends AppCompatActivity {
     recyclerView.setHasFixedSize(true);
     recyclerView.addItemDecoration(new VerticalGridCardSpacingDecoration());
     recyclerView.setAdapter(adapter);
+
+    // Many color models are shown on screen at once. The default recycled view pool size is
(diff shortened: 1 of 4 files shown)


epoxy-sample/src/main/java/com/airbnb/epoxy/MainActivity.java
```diff
@@ -38,6 +38,11 @@ protected void onCreate(Bundle savedInstanceState) {
     recyclerView.setHasFixedSize(true);
     recyclerView.addItemDecoration(new VerticalGridCardSpacingDecoration());
     recyclerView.setAdapter(adapter);
+
+    // Many color models are shown on screen at once. The default recycled view pool size is
+    // only 5, so we manually set the pool size to avoid constantly creating new views when
+    // the colors are randomized
+    recyclerView.getRecycledViewPool().setMaxRecycledViews(R.layout.model_color, 50);
```
elihart: Also through in this optimization. It will make the sample run better, and is also a nice trick to show.

gpeal:
👍

PR #35, opened 2016-09-14
[EpoxyModel] Use getLayout() method instead of layout field in toStri…

elihart:
@gpeal 

the field won't be set if the model is using the default layout. Using the getter is the correct way to access the current layout.


seanabraham:
Whoa first pull request acceptance I've seen on github!

PR #39, opened 2016-09-16
Prepare for release 1.3.0

elihart:
@gpeal @seanabraham

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 1.3.0 (09/15/2016)
+
+* Add support for using the view holder pattern with models. See the readme for more information.
+* Throw an exception if `EpoxyAdapter#notifyDataSetChanged()` is called when diffing is enabled. It doesn't make sense to allow this alongside diffing, and calling this is most likely to be an accidental mixup with `notifyModelsChanged()`.
+* Some performance improvements with the diffing algorithm.
+
 # 1.2.0 (09/07/2016)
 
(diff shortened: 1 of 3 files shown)


gradle.properties
```diff
@@ -1,5 +1,5 @@
 VERSION_CODE=1
```
seanabraham: I think you'll want to bump the version code too, no?

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 1.3.0 (09/15/2016)
```
seanabraham: You could potentially mention the NPE fix? Not sure how relevant it is.

gradle.properties
elihart: I've followed Felipe's example in other open source libraries and don't see this done. I'm not actually sure what this version code does but it doesn't seem used. I"ll talk to felipe

CHANGELOG.md
elihart: I would be incredibly surprised if anyone ran into it, which is why I didn't put it. more likely to confuse people since it's a bit complicated. I could add it though.

seanabraham: Got it, sounds like it doesn't need to be added!

gradle.properties
seanabraham: Hm, okay. I've always bumped it when bumping versions but mostly out of ceremony and not necessarily knowing what it does (although I always thought it might have mattered for some maven dependency related reasons).

Issue #36, opened 2016-09-15
Don't generate setter for final @EpoxyAttribute fields


elihart:

elihart:
https://github.com/airbnb/epoxy/pull/41

PR #41, opened 2016-09-16
#36 Don't generate setter for final @EpoxyAttribute fields

geralt-encore:
https://github.com/airbnb/epoxy/issues/36

### DIFF
--- a/epoxy-processor/src/main/java/com/airbnb/epoxy/EpoxyProcessor.java
+++ b/epoxy-processor/src/main/java/com/airbnb/epoxy/EpoxyProcessor.java
@@ -126,9 +126,10 @@ public class EpoxyProcessor extends AbstractProcessor {
     String name = attribute.getSimpleName().toString();
     TypeName type = TypeName.get(attribute.asType());
     boolean hasSuper = hasSuperMethod(classElement, name);
+    boolean hasFinalModifier = attribute.getModifiers().contains(FINAL);
     helperClass.addAttribute(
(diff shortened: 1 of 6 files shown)


epoxy-processor/src/main/java/com/airbnb/epoxy/EpoxyProcessor.java
```diff
@@ -160,7 +161,7 @@ private void validateAccessibleViaGeneratedCode(Element attribute) throws
 
     // Verify method modifiers.
     Set<Modifier> modifiers = attribute.getModifiers();
-    if (modifiers.contains(PRIVATE) || modifiers.contains(STATIC) || modifiers.contains(FINAL)) {
+    if (modifiers.contains(PRIVATE) || modifiers.contains(STATIC)) {
       throwError(
           "%s annotations must not be on private, final, or static fields. (class: %s, field: %s)",
```
elihart: update the error message to not include final

geralt-encore: done

elihart:
looks good to me, but not sure why the tests are failing. if you can fix them I'll merge, otherwise when I have time later I can pull the branch and figure out why the tests are failing

geralt-encore:
`testModelWithFinalClassFails` was wrong from the start, because in contained static field annotated with `@EpoxyAttribute` and failing because of it, but not because the class was actually final.

elihart:
Ah, my bad. Thanks for finding the issue and fixing! Looks great.

Issue #25, opened 2016-09-08
Package private field with @EpoxyAttribute annotation breaks when the class is extended in a different package


elihart:
Can either not include those in the generated class or throw an error at generation time

geralt-encore:
I can do it. Which option is better in your opinion?

elihart:
@geralt-encore That would be great!

I've been thinking about the best option, doesn't seem like a clear cut winner.

So far I'm thinking that we should take the route of not including package private attributes in the generated class if the class they came from is in a different package. If the field is package private then the developer should have had the intention that subclasses can't access them anyway.

 I'm a little worried about confusion arising from this if people see their annotated fields being ignored by the generator, but it should be a very uncommon case and we can add details in the readme.

Additionally, we can change the generated equals method to cast the object to the generated class instead of the base class so that `protected` fields in a different package can be accessed.

Does that make sense?

geralt-encore:
Totally makes sense

geralt-encore:
To be sure that we are on the same page could you please provide the simplest possible example of the problem?

elihart:
Sure.

Say we have `ClassA extends EpoxyModel` in package `exampleA`. Then we extend that class with `ClassB extends ClassA` in a different package `exampleB`.

The processor looks for fields with @EpoxyAttributes in all super classes. So @EpoxyAttribute fields on ClassA  will be included in the generated class for ClassB.

However, if those fields on ClassA are package private, ClassB can't access them. Since ClassB's generated class is in the same package as ClassB the generated class won't be able to access the fields as well.

I'm hoping we can detect that problem and just not include those fields.

There is also a problem with protected fields on ClassA, since the generated code for ClassB creates an equals method that casts the object to ClassB, in which case ClassB_ doesn't have access to the protected fields on ClassA.

 If we change this to cast the object to `ClassB_` then the generated code will have proper access.

Hopefully that makes sense :P

elihart:
https://github.com/airbnb/epoxy/pull/47

PR #44, opened 2016-09-22
 Ignore parents package private fields if parent is in a different package

geralt-encore:
https://github.com/airbnb/epoxy/issues/25

The only problem is that I wasn't able to write a test for that case because I couldn't figure out how to compile two resources at once since one of them is dependent on another and can't compile without him. If you have an idea or suggestion how to do it I will happily add it.


geralt-encore:
Should I rollback build tools back to 24.0.1?

FabianTerhorst:
Just Update the travis file.

elihart:
looks good at first glance, thanks for doing this! I'll need some more time to look in detail later, but I'll try to get to it soon.

Good point about writing a test for this, I'm not sure how to do it, but I'll think on it.

geralt-encore:
Shoot me if you have any questions!

epoxy-processor/src/main/java/com/airbnb/epoxy/EpoxyProcessor.java
```diff
@@ -127,9 +130,19 @@ private void processAttribute(Element attribute,
...
+            attribute.getAnnotation(EpoxyAttribute.class), hasSuper, hasFinalModifier,
+            packagePrivate));
+  }
+
+  /**
+   * Checks if the given field has package-private visibility
+   */
+  private boolean isFieldPackagePrivate(Element attribute) {
+    Set<Modifier> modifiers = attribute.getModifiers();
+    return !modifiers.contains(PUBLIC) && !modifiers.contains(PROTECTED);
```
elihart: Check for `!Private` too? I know we enforce a field can't be private elsewhere, but in the context of just this method it seems like we should also check it for this method to be correct

epoxy-processor/src/main/java/com/airbnb/epoxy/EpoxyProcessor.java
```diff
@@ -245,12 +258,29 @@ private void updateClassesForInheritance(
 
       for (Entry<TypeElement, ClassToGenerateInfo> otherEntry : otherClasses.entrySet()) {
         if (isSubtype(entry.getKey().asType(), otherEntry.getKey().asType())) {
-          entry.getValue().addAttributes(otherEntry.getValue().getAttributeInfo());
+          if (belongToTheSamePackage(entry.getKey(), otherEntry.getKey())) {
+            entry.getValue().addAttributes(otherEntry.getValue().getAttributeInfo());
+          } else {
+            for (AttributeInfo attribute : otherEntry.getValue().getAttributeInfo()) {
```
elihart: A comment here about why we do this would probably be helpful, just going back over this code that I haven't looked at in a few months I'm already having a hard time remembering what it does :P

If you prefer you can leave it for now and I can comments later after merging

epoxy-processor/src/main/java/com/airbnb/epoxy/EpoxyProcessor.java
```diff
@@ -337,8 +367,8 @@ private MethodSpec buildEquals(ClassToGenerateInfo helperClass) {
         .beginControlFlow("if (!super.equals(o))")
         .addStatement("return false")
         .endControlFlow()
-        .addStatement("$T that = ($T) o", helperClass.getOriginalClassNameWithoutType(),
-            helperClass.getOriginalClassNameWithoutType());
+        .addStatement("$T that = ($T) o", helperClass.getGeneratedName(),
```
elihart: Thanks for fixing this too, and for updating all the tests to reflect it

epoxy-sample/build.gradle
```diff
@@ -3,7 +3,7 @@ apply plugin: 'android-apt'
 
 android {
   compileSdkVersion 24
-  buildToolsVersion "24.0.1"
+  buildToolsVersion "24.0.2"
```
elihart: One of the less fun tasks I need to do is consolidate all dependency versions into one place, so when we need to update one of them we only do it once. I like the way butterknife does it, for example.

If you're interested that would be awesome. I need to make an issue about it

elihart:
Looks great, I like how you broke things down nicely with helper methods and comments. Very clean.

It would just be nice to have a test for this case, if I can't think of a way to make that work I'll merge it.

elihart:
Think I've got a way to test this :) I'll merge this and add the tests in another PR

epoxy-processor/src/main/java/com/airbnb/epoxy/EpoxyProcessor.java
geralt-encore: Makes sense.

epoxy-processor/src/main/java/com/airbnb/epoxy/EpoxyProcessor.java
geralt-encore: Would you prefer another PR for these fixes since this one has been already merged?

epoxy-sample/build.gradle
geralt-encore: Sure, I can do it. Should be pretty easy.

geralt-encore:
I am really interested about how are you going to test it, because I didn't find any suitable solution using current setup.

elihart:
https://github.com/airbnb/epoxy/pull/50 just worked around it and didn't use the processor :P

Issue #43, opened 2016-09-20
Add class based annotation so EpoxyModels with attributes can still get a generated class


elihart:
Helpful for models that extend other models with Epoxy attributes, where the subclass doesn't need to add any attributes

geralt-encore:
So basically it should be allowed only for models that extend other models with attributes?

elihart:
Yes, but that is just the use cases I have seen. It may be that somebody may want a generated class for a model that doesn't have any attributes on super classes. 

Not sure why that would be needed, but I don't see any errors that would come from that case so I don't see any reason to prevent that usage.

elihart:
#47

PR #47, opened 2016-09-23
Add class based annotation

geralt-encore:
https://github.com/airbnb/epoxy/issues/43

Not sure about the name of annotation,  but it could be easily changed if needed.
Also, we are lacking right now validation for EpoxyClass annotation in case someone annotates not  a class but interface or enum with it. Do we need it?
And sorry for long names in tests - didn't come up with something smarter :D


epoxy-model/src/main/java/com/airbnb/epoxy/EpoxyClass.java
```diff
@@ -0,0 +1,16 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Used to annotate {@link EpoxyModel} classes in order to generate a
+ * subclass of that model with getters, setters, equals, and hashcode for the annotated fields.
+ */
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.CLASS)
+public @interface EpoxyClass {
```
elihart: We could maybe go with `EpoxyModelClass` to be more clear, since any class in this library could be considered an Epoxy class

epoxy-processor/src/main/java/com/airbnb/epoxy/EpoxyProcessor.java
```diff
@@ -93,12 +94,15 @@ public SourceVersion getSupportedSourceVersion() {
...
-    for (Element attribute : roundEnv.getElementsAnnotatedWith(EpoxyAttribute.class)) {
-      try {
+    try {
+      for (Element attribute : roundEnv.getElementsAnnotatedWith(EpoxyAttribute.class)) {
         processAttribute(attribute, modelClassMap);
-      } catch (EpoxyProcessorException e) {
-        writeError(e);
       }
+      for (Element clazz : roundEnv.getElementsAnnotatedWith(EpoxyClass.class)) {
+        processClass((TypeElement) clazz, modelClassMap);
```
elihart: I'd be fine with calling `getOrCreateTargetClass(modelClassMap, classElement)` directly here and getting rid of the `processClass` method, since that method is only called once and is just the one line. The naming adds some readability, but it also adds an extra step to follow

elihart:
Nice! you're cranking these PR's out.

It's nice that we can mostly reuse `getOrCreateTargetClass`. That method actually checks 
```
if (!isSubtype(classElement.asType(), epoxyModelType)) {
      throwError("Class with %s annotations must extend %s (%s)",
          EpoxyAttribute.class.getSimpleName(), epoxyModelType,
          classElement.getSimpleName());
    }
```

so it will enforce that the annotated object extends EpoxyModel, which should prevent this being used on an enum or interface. Maybe we can include a test for that?

Long names are fine in testing :P thanks for the good test coverage.

Everything looks pretty good, but let's think about the annotation name a little bit before finalizing and merging this.

epoxy-model/src/main/java/com/airbnb/epoxy/EpoxyClass.java
geralt-encore: done

epoxy-processor/src/main/java/com/airbnb/epoxy/EpoxyProcessor.java
geralt-encore: and done

geralt-encore:
You right, I missed that part. I can add tests for this case, but not sure if it's really needed or not.

elihart:
@geralt-encore I don't think we need explicit tests since I believe there is already a test that checks for extending epoxy model.

I appreciate the help! I'm curious if you are using Epoxy in your app, or what your motivation is for helping? 

A reason I ask is it would be great to get feedback on how other people use this, since I only really have my team's experience and use cases to go off of for what features and issues there are.

One thing I'm thinking about is that it can be easy to use the base class instead of the generated class (`MyModel` instead of `MyModel_`) and you would lose the correct hashcode implementation there and diffing wouldn't work.

To prevent that we could have a convention of making base models abstract to prevent accidental usage. The processor currently avoids generating classes for abstract models since `getDefaultLayout` might not be implemented and if so the generated class would not compile.

One thing we could do is generate models for an abstract class if it has the `EpoxyModelClass` annotation on it. If we go that route it may make sense to rename the annotation something like `@GenerateModel` to indicate more broadly that a model should be generated for that class.

Sorry for the long ramblings but it is helpful to bounce ideas off of external users.

geralt-encore:
Yeah, me too. These tests are kinda overkill, I guess.

Unfortunately, I haven't used Epoxy in a real app yet, but I just really like a concept of the library and see how many problems from my experience it actually solves. I just want to contribute to the interesting and useful project, so, basically, it's my motivation.

I see your reasoning, but I am not sure, that convention is a better choice since it so easy to break. Also, it's a common practice to use `MyModel_` instead of `MyModel` in libraries where annotations processing involved, so users should be aware of it.

elihart:
lgtm, I'll merge.

Good points. We just have a large team and anything we can do to prevents accidents or confusion is better. The tests are a pain, but hopefully they'll help prevent regressions later on if more features are added or other people contribute.

Hope you get try actually try out the library in an app sometime! I'm sure it would give you some ideas about features to add or other things to improve.

PR #50, opened 2016-09-24
Add test for classes in different package

elihart:
@geralt-encore 

Adds a test for https://github.com/airbnb/epoxy/pull/44

Instead of using a processor this just uses a normal test and actual model class.

I also realized that in the equals method we changed the type to be the generated class in the cast, but not in the instance of check, which could cause crashes.

### DIFF
--- a/epoxy-processor/src/main/java/com/airbnb/epoxy/EpoxyProcessor.java
+++ b/epoxy-processor/src/main/java/com/airbnb/epoxy/EpoxyProcessor.java
@@ -146,7 +146,9 @@ public class EpoxyProcessor extends AbstractProcessor {
    */
   private boolean isFieldPackagePrivate(Element attribute) {
     Set<Modifier> modifiers = attribute.getModifiers();
-    return !modifiers.contains(PUBLIC) && !modifiers.contains(PROTECTED);
+    return !modifiers.contains(PUBLIC)
+        && !modifiers.contains(PROTECTED)
+        && !modifiers.contains(PRIVATE);
   }
 
   /**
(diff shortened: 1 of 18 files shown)


epoxy-processor/src/main/java/com/airbnb/epoxy/EpoxyProcessor.java
```diff
@@ -252,23 +254,33 @@ private ClassName getGeneratedClassName(TypeElement classElement) {
    * Check each model for super classes that also have attributes. For each super class with
    * attributes we add those attributes to the attributes of the generated class, so that a
    * generated class contains all the attributes of its super classes combined.
+   * <p>
+   * One caveat is that if a sub class is in a different package than its super class we can't
+   * include attributes that are package private, otherwise the generated class won't compile.
    */
   private void updateClassesForInheritance(
```
elihart: This original method was already dense and hard to read, adding the package stuff on top of it made it denser, so I cleaned it up a little bit.

epoxy-processor/src/main/java/com/airbnb/epoxy/EpoxyProcessor.java
```diff
@@ -365,7 +381,7 @@ private MethodSpec buildEquals(ClassToGenerateInfo helperClass) {
         .beginControlFlow("if (o == this)")
         .addStatement("return true")
         .endControlFlow()
-        .beginControlFlow("if (!(o instanceof $T))", helperClass.getOriginalClassNameWithoutType())
+        .beginControlFlow("if (!(o instanceof $T))", helperClass.getGeneratedName())
```
elihart: Use the generated name in the instance of, so when we cast below this we won't get a class cast exception

epoxy-processor/src/test/resources/BasicModelWithAttribute_.java
```diff
@@ -56,7 +56,7 @@ public boolean equals(Object o) {
     if (o == this) {
       return true;
     }
-    if (!(o instanceof BasicModelWithAttribute)) {
+    if (!(o instanceof BasicModelWithAttribute_)) {
```
elihart: Thank god for intellijs `replace structurally`

geralt-encore:
👍

PR #52, opened 2016-09-25
Custom arraylist that reports model changes

elihart:
This creates a ModelList class that extends ArrayList and notifies a listener whenever something is added or removed from the list. A listener in the diff helper uses this information to optimize diffing.

This makes the differ quite a bit faster if no changes have been done to the list, or if just adds or just removes were done, since moves don't need to be checked. This will also be more memory efficient since it isn't creating new state objects for each model.

With changes:
list_test_changes

Before changes:
list_test_master

### DIFF
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/DiffHelper.java
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/DiffHelper.java
@@ -2,9 +2,6 @@
 package com.airbnb.epoxy;
 
 import android.support.annotation.Nullable;
-import android.support.v7.util.DiffUtil;
-import android.support.v7.util.DiffUtil.Callback;
-import android.support.v7.util.DiffUtil.DiffResult;
 import android.support.v7.widget.RecyclerView;
 
 import java.util.ArrayList;
@@ -17,7 +14,6 @@ import java.util.Map;
  * Helper to track changes in the models list.
  */
 class DiffHelper {
-  private static final boolean USE_DIFF_UTIL = false;
   private ArrayList<ModelState> oldStateList = new ArrayList<>();
   // Using a HashMap instead of a LongSparseArray to
   // have faster look up times at the expense of memory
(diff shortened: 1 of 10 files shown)


epoxy-adapter/src/main/java/com/airbnb/epoxy/DiffHelper.java
```diff
@@ -17,18 +14,29 @@
  * Helper to track changes in the models list.
  */
 class DiffHelper {
-  private static final boolean USE_DIFF_UTIL = false;
```
elihart: I also removed the optional usage of the support library diff util since it is definitely slower now, and keeping it around made the code messier and more complex

epoxy-adapter/src/main/java/com/airbnb/epoxy/DifferModelListObserver.java
```diff
@@ -0,0 +1,38 @@
...
+ * Listens for changes to the model array list in the adapter, storing and batching insertions and
+ * removals to that list. This knowledge helps us be more efficient with the differ, avoiding diff
+ * computations when not necessary and allowing us to take some shortcuts.
+ * <p>
+ * This isn't a silver bullet, as users could clear the list and add back the same objects resulting
+ * in no net change, but this would record a batch removal and batch insert. Thus when we record
+ * both removals and insertions we still need to run a full diff. However, if the list doesn't
+ * change, or only changes due to insertions or removals, we can skip doing a full diff.
+ */
+class DifferModelListObserver extends UpdateOpHelper implements ModelListObserver {
```
elihart: This leverages the previous work I did on the `UpdateOpHelper`, which makes batching list changes easy.

epoxy-adapter/src/main/java/com/airbnb/epoxy/ModelList.java
```diff
@@ -0,0 +1,510 @@
...
+import java.util.AbstractList;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.ConcurrentModificationException;
+import java.util.Iterator;
+import java.util.List;
+import java.util.ListIterator;
+import java.util.NoSuchElementException;
+
+public class ModelList extends ArrayList<EpoxyModel<?>> {
```
elihart: I should add javadoc to this class

elihart: and make it package private

PR #53, opened 2016-09-27
Eli/annotation library fix

elihart:
The annotation processor and `EpoxyModel` was depending on the support library annotations library for the @LayoutRes annotation. This only worked because that library was mistakenly publicly released on maven, but normally java libraries should not have access to the android support libraries. I had to remove that dependency, and so moved `EpoxyModel` to the adapter module, and stubbed out dependencies for EpoxyModel and @LayoutRes in the processor.

Then I had to move all the processor tests to a new android module so they could also depend on the EpoxyModel in order for the test code to compile.

I ran into issues with the resources not being picked up when running tests (https://code.google.com/p/android/issues/detail?id=64887) and by using https://github.com/nenick/AndroidStudioAndRobolectric/blob/master/app/build.workaround-missing-resource.gradle I got this working. Running processor tests from android studio now works properly as an added benefit :) @felipecsl fyi including that build script is a nice way to get the resources working, may want to consider it for other projects.


elihart:
@niccorder we should be able to update and merge your gradle clean up after this :)

epoxy-processor/src/main/java/com/airbnb/epoxy/EpoxyProcessor.java
```diff
@@ -65,7 +65,9 @@
 @AutoService(Processor.class)
 public class EpoxyProcessor extends AbstractProcessor {
   private static final String GENERATED_CLASS_NAME_SUFFIX = "_";
-  private static TypeMirror epoxyModelType;
+  private static final String EPOXY_MODEL_TYPE = "com.airbnb.epoxy.EpoxyModel<?>";
```
elihart: Instead of referencing the model class and annotation directly I stubbed them out.

epoxy-adapter/build.gradle
```diff
@@ -15,7 +15,7 @@ dependencies {
   compile "com.android.support:recyclerview-v7:24.2.1"
   compile "com.android.support:design:24.2.1"
   compile "com.android.support:appcompat-v7:24.2.1"
-  compile project(":epoxy-model")
+  compile project(':epoxy-annotations')
```
elihart: I also renamed the model module to `annotations`, since all it has is annotations now that the EpoxyModel was moved to `epoxy-adapter` module

niccorder:
LGTM! This was for sure an odd issue, but I'm happy the fix was a little refactor 👍

PR #54, opened 2016-10-01
Simplify build.workaround-missing-resource.gradle

geralt-encore:
Since flavors aren't really needed for library project this workaround could be simplified.


elihart:
Great. simple is better :) thanks!

Issue #37, opened 2016-09-15
Have generated classes duplicate any methods from super classes that return the class type


elihart:
To make chaining easier.

@gpeal

geralt-encore:
What about these improvements that we were talking in PR comments?

elihart:
There were two things:

1) If there is just one param in a method then name it the same as the method name

2) Also look for methods  on super classes beyond the first parent. We  will have to check for duplicates here, as well as package private methods for super classes in different packages. Then we could remove the custom code to build the id, show, etc methods.

And now that I think about it it seems like we should also capture annotations on either the method or the parameters.

geralt-encore:
I was able to look for methods on super classes beyond the first parent and eliminate custom code to build the id, layout, show and hide methods. The only problem right now is that all tests are failing, because of `@LayoutRes` annotation which is missing because we ignore annotations for methods and method parameters. I am not sure how to proceed next since I haven't known yet how to deal with that case and it's kinda out of scope for the PR, but obviously it can't be merged in current state because tests are failing.

elihart:
That's great that you have that part working! The annotations are tricky though.

I've been looking into them and think I know how to do it. I'll put up a PR which we can merge first, then you should be able to add your changes without break tests. Sound good?

geralt-encore:
Sounds good! Let me know if you need help with it.

elihart:
Unfortunately my idea didn't work. I noticed at run time that the method element is a `com.sun.tools.javac.code.Symbol.MethodSymbol` and that contains all the annotation information, and even the parameter name. As far I can tell though we can't cast it to that since it's an implementation of the compiler and not an accessible package. We could maybe somehow do something hacky to work around that but I think that's too ugly.

`typeMirror.getAnnotationMirrors()` is available starting in java 8. I believe this is the only reliable way we could get the annotations and unfortunately we're stuck below java 8.

We could drop the annotations or not include methods in super classes. I don't think dropping annotations is the worst thing. Maybe we could drop annotations on most methods, but manually add the `layout` method from the EpoxyModel class that has one.

elihart:
Awesome, all done!

PR #56, opened 2016-10-02
#37 Duplicate methods returning class type

geralt-encore:
https://github.com/airbnb/epoxy/issues/37

### DIFF
--- a/epoxy-processor/src/main/java/com/airbnb/epoxy/ClassToGenerateInfo.java
+++ b/epoxy-processor/src/main/java/com/airbnb/epoxy/ClassToGenerateInfo.java
@@ -9,7 +9,9 @@ import com.squareup.javapoet.TypeVariableName;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
 
@@ -33,6 +35,7 @@ public class ClassToGenerateInfo {
   private final Set<AttributeInfo> attributeInfo = new HashSet<>();
   private final List<TypeVariableName> typeVariableNames = new ArrayList<>();
   private final List<ConstructorInfo> constructors = new ArrayList<>();
+  private final List<MethodInfo> methodsReturningClassType = new ArrayList<>();
 
   public ClassToGenerateInfo(TypeElement originalClassName, ClassName generatedClassName,
       boolean isOriginalClassAbstract) {
@@ -50,7 +53,22 @@ public class ClassToGenerateInfo {
           && !subElement.getModifiers().contains(Modifier.PRIVATE)) {
         List<? extends TypeMirror> params = subElement.asType().accept(CONSTRUCTOR_VISITOR, null);
         constructors
-            .add(new ConstructorInfo(subElement.getModifiers(), buildConstructorParamList(params)));
+            .add(new ConstructorInfo(subElement.getModifiers(), buildParamList(params)));
+      }
+    }
+
+    // Get information about methods returning class type of the original class so we can
+    // duplicate them in the generated class for chaining purposes
+    for (Element subElement : originalClassName.getEnclosedElements()) {
+      if (subElement.getKind() == ElementKind.METHOD
+          && ((ExecutableType) subElement.asType()).getReturnType()
+            .equals(originalClassName.asType())
+          && !subElement.getModifiers().contains(Modifier.PRIVATE)
+          && !subElement.getModifiers().contains(Modifier.STATIC)) {
+        List<? extends TypeMirror> params = ((ExecutableType) subElement.asType())
+            .getParameterTypes();
+        methodsReturningClassType.add(new MethodInfo(subElement.getSimpleName().toString(),
+            subElement.getModifiers(), buildParamList(params)));
       }
     }
 
@@ -66,7 +84,7 @@ public class ClassToGenerateInfo {
     this.isOriginalClassAbstract = isOriginalClassAbstract;
   }
 
-  private List<ParameterSpec> buildConstructorParamList(List<? extends TypeMirror> params) {
+  private List<ParameterSpec> buildParamList(List<? extends TypeMirror> params) {
     List<ParameterSpec> result = new ArrayList<>();
 
     // We don't know the name of the variable, just the type. So just use generic parameter names
@@ -86,10 +104,12 @@ public class ClassToGenerateInfo {
       };
 
   public void addAttribute(AttributeInfo attributeInfo) {
+    removeMethodIfDuplicatedBySetter(Collections.singletonList(attributeInfo));
     this.attributeInfo.add(attributeInfo);
   }
 
   public void addAttributes(Collection<AttributeInfo> attributeInfo) {
+    removeMethodIfDuplicatedBySetter(attributeInfo);
     this.attributeInfo.addAll(attributeInfo);
   }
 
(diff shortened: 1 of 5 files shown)


epoxy-processortest/src/test/resources/ModelReturningClassType.java
```diff
@@ -0,0 +1,14 @@
+package com.airbnb.epoxy;
+
+public class ModelReturningClassType extends EpoxyModel<Object> {
+  @EpoxyAttribute int value;
```
elihart: maybe add another field and setter for it to test the `removeMethodIfDuplicatedBySetter` code?

epoxy-processor/src/main/java/com/airbnb/epoxy/ClassToGenerateInfo.java
```diff
@@ -125,6 +149,18 @@ public TypeName getParameterizedGeneratedName() {
     return parameterizedClassName;
   }
 
+  private void removeMethodIfDuplicatedBySetter(Collection<AttributeInfo> attributeInfos) {
+    for (AttributeInfo attributeInfo : attributeInfos) {
+      Iterator<MethodInfo> iterator = methodsReturningClassType.iterator();
+      while (iterator.hasNext()) {
+        MethodInfo methodInfo = iterator.next();
+        if (methodInfo.name.equals(attributeInfo.getName())) {
```
elihart: This doesn't check for overloaded methods, which is probably ok for now, but would be nice to have eventually. We would have to check if the `methodInfo` object has just one param, and that the type of that param matches `attributeInfo.type`, before we remove it.

epoxy-processor/src/main/java/com/airbnb/epoxy/ClassToGenerateInfo.java
```diff
@@ -86,10 +104,12 @@ public ClassToGenerateInfo(TypeElement originalClassName, ClassName generatedCla
       };
 
   public void addAttribute(AttributeInfo attributeInfo) {
+    removeMethodIfDuplicatedBySetter(Collections.singletonList(attributeInfo));
```
elihart: might as well do `addAttributes(Collections.singletonList(attributeInfo))` to remove this duplication

epoxy-processor/src/main/java/com/airbnb/epoxy/ClassToGenerateInfo.java
```diff
@@ -66,7 +84,7 @@ public ClassToGenerateInfo(TypeElement originalClassName, ClassName generatedCla
     this.isOriginalClassAbstract = isOriginalClassAbstract;
   }
 
-  private List<ParameterSpec> buildConstructorParamList(List<? extends TypeMirror> params) {
+  private List<ParameterSpec> buildParamList(List<? extends TypeMirror> params) {
```
elihart: It's nice that we can reuse this method, but I'm thinking if there is only one param we should go with the convention of naming that param after the method. This method was made for constructors so we can't do that here so well. Any ideas on a nice way to have this custom naming for the methods without having to duplicate all this?

epoxy-processor/src/main/java/com/airbnb/epoxy/ClassToGenerateInfo.java
```diff
@@ -50,7 +53,22 @@ public ClassToGenerateInfo(TypeElement originalClassName, ClassName generatedCla
...
+    for (Element subElement : originalClassName.getEnclosedElements()) {
+      if (subElement.getKind() == ElementKind.METHOD
+          && ((ExecutableType) subElement.asType()).getReturnType()
+            .equals(originalClassName.asType())
+          && !subElement.getModifiers().contains(Modifier.PRIVATE)
+          && !subElement.getModifiers().contains(Modifier.STATIC)) {
+        List<? extends TypeMirror> params = ((ExecutableType) subElement.asType())
+            .getParameterTypes();
+        methodsReturningClassType.add(new MethodInfo(subElement.getSimpleName().toString(),
+            subElement.getModifiers(), buildParamList(params)));
```
elihart: could pull out `subElement.getModifiers()` to a variable since it is called a few times

geralt-encore:
Thanks for the review! I'll fix your comments tomorrow. Regarding improvements you have mentioned in the first comment, I'd rather leave this PR as it is and add improvements in next PRs. Would be nice to have tasks or subtasks in order not to miss something and be sure that we understand each other correctly.

elihart:
Sounds like a great approach, getting this in soon and iterating will be nice 👍

epoxy-processor/src/main/java/com/airbnb/epoxy/ClassToGenerateInfo.java
geralt-encore: The only idea that comes to my mind is to have a boolean flag to be able to distinguish between method call for constructor and method, but it looks kinda ugly.

epoxy-processor/src/main/java/com/airbnb/epoxy/ClassToGenerateInfo.java
geralt-encore: Thanks! I haven't thought about it, but it totally makes sense. Definitely will fix it.

epoxy-processor/src/main/java/com/airbnb/epoxy/ClassToGenerateInfo.java
```diff
@@ -134,4 +170,16 @@ public ConstructorInfo(Set<Modifier> modifiers, List<ParameterSpec> params) {
       this.params = params;
     }
   }
+
+  public static class MethodInfo {
```
geralt-encore: Then we also can use `MethodSpec` instead of `ConstructorInfo`?

geralt-encore: I dunno, don't think it's a good idea because we are building actual methods later, so we need this intermediate data, so we don't mix this logic with code generation logic.

epoxy-processor/src/main/java/com/airbnb/epoxy/ClassToGenerateInfo.java
geralt-encore: 👍

epoxy-processortest/src/test/resources/ModelReturningClassType.java
geralt-encore: 👍

geralt-encore: Actually, there is no need for it because we have `testModelWithSuper` which initially was failing in that case. That's how I discovered this case :D

epoxy-processor/src/main/java/com/airbnb/epoxy/EpoxyProcessor.java
```diff
@@ -400,6 +402,37 @@ private void generateClassForModel(ClassToGenerateInfo info) throws IOException
...
+      Builder builder = MethodSpec.methodBuilder(methodInfo.name)
+          .addModifiers(methodInfo.modifiers)
+          .addParameters(methodInfo.params)
+          .addAnnotation(Override.class)
+          .returns(info.getGeneratedName());
+
+      StringBuilder statementBuilder = new StringBuilder(String.format("super.%s(",
+          methodInfo.name));
+      boolean first = true;
+      for (ParameterSpec param : methodInfo.params) {
```
elihart: maybe we can pull this out into a `join(List<String>, delimiter)` since it is done both here and in the constructor method.

geralt-encore: Not sure that I understood your suggestion, but I've made it a bit different.

elihart: Ah this is good. I was thinking more of a generic join method like `StringUtils.join`, but your way is better since we have a special usage case. Thanks!

PR #59, opened 2016-10-10
#37 Look up for super methods

geralt-encore:
https://github.com/airbnb/epoxy/issues/37

### DIFF
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModel.java
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModel.java
@@ -140,13 +140,13 @@ public abstract class EpoxyModel<T> {
     return this;
   }
 
-  public EpoxyModel<T> hide() {
-    shown = false;
+  public EpoxyModel<T> show(boolean show) {
+    shown = show;
     return this;
   }
 
-  public EpoxyModel<T> show(boolean show) {
-    shown = show;
+  public EpoxyModel<T> hide() {
+    shown = false;
     return this;
   }
 
--- a/epoxy-processor/src/main/java/com/airbnb/epoxy/ClassToGenerateInfo.java
+++ b/epoxy-processor/src/main/java/com/airbnb/epoxy/ClassToGenerateInfo.java
@@ -12,6 +12,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
 
@@ -21,9 +22,11 @@ import javax.lang.model.element.Modifier;
 import javax.lang.model.element.TypeElement;
 import javax.lang.model.element.TypeParameterElement;
 import javax.lang.model.type.ExecutableType;
+import javax.lang.model.type.TypeKind;
 import javax.lang.model.type.TypeMirror;
 import javax.lang.model.type.TypeVisitor;
 import javax.lang.model.util.SimpleTypeVisitor6;
+import javax.lang.model.util.Types;
 
 public class ClassToGenerateInfo {
 
@@ -35,10 +38,12 @@ public class ClassToGenerateInfo {
   private final Set<AttributeInfo> attributeInfo = new HashSet<>();
   private final List<TypeVariableName> typeVariableNames = new ArrayList<>();
   private final List<ConstructorInfo> constructors = new ArrayList<>();
-  private final List<MethodInfo> methodsReturningClassType = new ArrayList<>();
+  private final Set<MethodInfo> methodsReturningClassType = new LinkedHashSet<>();
+  private final Types typeUtils;
 
-  public ClassToGenerateInfo(TypeElement originalClassName, ClassName generatedClassName,
-      boolean isOriginalClassAbstract) {
+  public ClassToGenerateInfo(Types typeUtils, TypeElement originalClassName,
+      ClassName generatedClassName, boolean isOriginalClassAbstract) {
+    this.typeUtils = typeUtils;
     this.originalClassName = ParameterizedTypeName.get(originalClassName.asType());
     this.originalClassNameWithoutType = ClassName.get(originalClassName);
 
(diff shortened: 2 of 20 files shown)


elihart:
Thanks for putting this up. I'm a bit busy these days unfortunately but I'll try to look soon

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModel.java
```diff
@@ -140,13 +140,13 @@ public int getSpanSize(int totalSpanCount, int position, int itemCount) {
     return this;
   }
 
-  public EpoxyModel<T> hide() {
-    shown = false;
+  public EpoxyModel<T> show(boolean show) {
```
elihart: nothing really changed here?

epoxy-processor/src/main/java/com/airbnb/epoxy/ClassToGenerateInfo.java
```diff
@@ -81,26 +90,39 @@ private void collectOriginalClassConstructors(TypeElement originalClass) {
...
-              Collections.singletonList(parameterSpec)));
-        } else {
-          methodsReturningClassType.add(new MethodInfo(methodName, modifiers,
-              buildParamList(params)));
+    TypeElement clazz = originalClass;
+    while (clazz.getSuperclass().getKind() != TypeKind.NONE) {
+      for (Element subElement : clazz.getEnclosedElements()) {
+        Set<Modifier> modifiers = subElement.getModifiers();
+        if (subElement.getKind() == ElementKind.METHOD
+            && !modifiers.contains(Modifier.PRIVATE)
```
elihart: oh, we should add not `final` too

epoxy-processor/src/main/java/com/airbnb/epoxy/ClassToGenerateInfo.java
```diff
@@ -81,26 +90,39 @@ private void collectOriginalClassConstructors(TypeElement originalClass) {
...
-          TypeMirror param = params.get(0);
-          ParameterSpec parameterSpec = ParameterSpec.builder(TypeName.get(param),
-              methodName).build();
-          methodsReturningClassType.add(new MethodInfo(methodName, modifiers,
-              Collections.singletonList(parameterSpec)));
-        } else {
-          methodsReturningClassType.add(new MethodInfo(methodName, modifiers,
-              buildParamList(params)));
+    TypeElement clazz = originalClass;
+    while (clazz.getSuperclass().getKind() != TypeKind.NONE) {
```
elihart: we really only need to look until the base EpoxyModel. maybe reuse the `isEpoxyModel(clazz.asType())` method from the Processor class (move it into a Util class).

epoxy-processor/src/main/java/com/airbnb/epoxy/ClassToGenerateInfo.java
```diff
@@ -81,26 +90,39 @@ private void collectOriginalClassConstructors(TypeElement originalClass) {
...
+          TypeMirror methodReturnType = ((ExecutableType) subElement.asType()).getReturnType();
+          if (methodReturnType.equals(clazz.asType())
+              || typeUtils.isSubtype(clazz.asType(), methodReturnType)) {
+            List<? extends TypeMirror> params = ((ExecutableType) subElement.asType())
+                .getParameterTypes();
+            String methodName = subElement.getSimpleName().toString();
+            if (params.size() == 1) {
+              TypeMirror param = params.get(0);
+              ParameterSpec parameterSpec;
+              if (methodName.equals(LAYOUT_METHOD) && param.getKind() == TypeKind.INT) {
```
elihart: adding a brief comment here about why we need to handle layout specially would probably be helpful in the future

epoxy-processor/src/main/java/com/airbnb/epoxy/ClassToGenerateInfo.java
```diff
@@ -203,5 +225,33 @@ public MethodInfo(String name, Set<Modifier> modifiers, List<ParameterSpec> para
       this.modifiers = modifiers;
       this.params = params;
     }
+
+    @Override
+    public boolean equals(Object o) {
```
elihart: great! I like this approach of using equals to prevent duplicates. Since this class is getting large what do you think of making `MethodInfo`  top level class? We could also maybe move `collectMethodsReturningClassType` into that class as a static method to build a list of methods based on a class type

epoxy-processor/src/main/java/com/airbnb/epoxy/EpoxyProcessor.java
```diff
@@ -578,61 +575,6 @@ private MethodSpec generateSetter(ClassToGenerateInfo helperClass, AttributeInfo
...
-   * generated class type for use in chaining.
-   */
-  private List<MethodSpec> buildDefaultSetters(ClassToGenerateInfo helperClass) {
-    List<MethodSpec> result = new ArrayList<>();
-
-    result.add(MethodSpec.methodBuilder("id")
-        .addModifiers(Modifier.PUBLIC)
-        .returns(helperClass.getParameterizedGeneratedName())
-        .addAnnotation(Override.class)
-        .addParameter(long.class, "id")
```
elihart: Great to see this all get deleted!

epoxy-processortest/src/test/resources/ModelWithAllFieldTypes_.java
```diff
@@ -218,8 +218,8 @@ public ModelWithAllFieldTypes_ id(long id) {
   }
 
   @Override
-  public ModelWithAllFieldTypes_ layout(@LayoutRes int layoutRes) {
-    super.layout(layoutRes);
+  public ModelWithAllFieldTypes_ layout(@LayoutRes int layout) {
```
elihart: Thanks for fixing the tests with this name change

epoxy-processor/src/main/java/com/airbnb/epoxy/ClassToGenerateInfo.java
```diff
@@ -81,26 +90,39 @@ private void collectOriginalClassConstructors(TypeElement originalClass) {
...
+            String methodName = subElement.getSimpleName().toString();
+            if (params.size() == 1) {
+              TypeMirror param = params.get(0);
+              ParameterSpec parameterSpec;
+              if (methodName.equals(LAYOUT_METHOD) && param.getKind() == TypeKind.INT) {
+                parameterSpec = ParameterSpec.builder(int.class, methodName)
+                    .addAnnotation(LAYOUT_RES_ANNOTATION)
+                    .build();
+              } else {
+                parameterSpec = ParameterSpec.builder(TypeName.get(param), methodName).build();
```
elihart: This is an edge case, but are parameterized params handled correctly? Like if we have a `List<String>` param does the generated method include the type? No biggie if not, we can add it as a follow up

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModel.java
geralt-encore: I just swap `show` and `hide` methods so they appears in a right order.

epoxy-processor/src/main/java/com/airbnb/epoxy/ClassToGenerateInfo.java
geralt-encore: But it's, basically the same because `TypeKind.NONE` means `Object` and `EpoxyModel` don't have a parent except it.

epoxy-processor/src/main/java/com/airbnb/epoxy/ClassToGenerateInfo.java
geralt-encore: True

epoxy-processor/src/main/java/com/airbnb/epoxy/ClassToGenerateInfo.java
geralt-encore: done

epoxy-processor/src/main/java/com/airbnb/epoxy/ClassToGenerateInfo.java
geralt-encore: Looks like it crashes so I have to investigate it

epoxy-processor/src/main/java/com/airbnb/epoxy/ClassToGenerateInfo.java
geralt-encore: I like that idea, but then we probably should do the same with `ConstructorInfo` since they are essentially the same, but this one is rather small.

epoxy-processor/src/main/java/com/airbnb/epoxy/ClassToGenerateInfo.java
geralt-encore: No, actually, it works fine. Added test case for it

epoxy-processor/src/main/java/com/airbnb/epoxy/ClassToGenerateInfo.java
elihart: Constructor info is very basic though so they don't necessarily need to be treated the same way. it could be a clean up task in the future though

elihart:
Thanks for the fix ups, great stuff!

Issue #64, opened 2016-10-14
Generate toString method on model classes


elihart:
I'm thinking this could be useful for debugging, at times where a model is logged and its toString implementation gives details about it. If we generate an implementation that lists all the attributes it could make logs more useful.

Low priority.

elihart:
Added with https://github.com/airbnb/epoxy/pull/65. 

We can follow up to also include the `super.toString()` result, since super classes (like the base EpoxyModel) may also add information that is useful in the string output.

geralt-encore:
The thing with `super.toString()` that it looks like it is impossible to do without reflection if you are not an owner of the class and you precisely doesn't know what fields are available from super classes. We can add fields annotated with `EpoxyAttribute` from super classes, but then we could have some inconsistency with fields without annotation aren't included into `toString()`. So overall I am not sure if it is a good idea to include `super.toString()`.

elihart:
I'm not proposing using reflection to generate the `super.toString()`, just adding a call to it like 
```
@Override
  public String toString() {
    return "TestModel{" +
        "publicValue=" + publicValue +
        ", protectedValue=" + protectedValue +
        ", packagePrivateValue=" + packagePrivateValue +
        "} " + super.toString();
  }
```

When you generate a toString method you can select a template, and there is a template in Android Studio called `String concat and super.toString()` that generates this format. I was just thinking of something simple like that

geralt-encore:
Looks pretty simple. I'll check out this template and adjust current implementation to it.

elihart:
thanks!

elihart:
https://github.com/airbnb/epoxy/pull/68

PR #65, opened 2016-10-15
#64 Generate toString method

geralt-encore:
https://github.com/airbnb/epoxy/issues/64


elihart:
Looks pretty good overall! Only thing is it would be useful to also include the `super.toString()` result, since super classes (like the base EpoxyModel) may also add information

geralt-encore:
I was thinking about it initially as well, but I took Intellij Idea's template as a reference and it doesn't contain super call.

elihart:
Hm, I'm not sure why intellij wouldn't include that in their template. I think what you have is a great start so I'll merge it for now and keep the super call as a potential add on in the future.

Thanks!

Issue #67, opened 2016-10-18
Allow for subclasses in other modules in processor


elihart:
In the annotation processor we look for annotated fields on super classes and bundle them with sub classes. This relies on the super classes being included in the processor - we don't walk up the inheritance tree and explicitly look for super classes with annotations.

This creates a problem where if the super class is in a different module it's annotated fields will not be included in generated subclasses in other modules.

Solution should be to walk up the inheritance tree for each model class found in the processor.

geralt-encore:
Can you elaborate more on this or provide an example? Seems like I don't get it completely =(

elihart:
Sure, say Model A is in module A and Model B extends Model A and is in module B.

When the annotation processor runs it gets annotated elements for a single module at a time, so it will process module B separately from module A.

Since Model B extends Model A it may inherit @EpoxyAttribute annotated fields from Model A. Right now the processor groups inherited fields only if the super class is being processed at the same time as the sub class. This is because it goes through all models with annotations in that module, and then looks at which classes are subclasses of another and adds inherited annotations if so.

This approach doesn't work if the models are processed separately when they are in separate modules.

elihart:
I'll try to work on this myself today

PR #71, opened 2016-10-21
Fix processor to include attributes from super class models in other modules

elihart:
https://github.com/airbnb/epoxy/issues/67


geralt-encore:
How common is that usecase? I am just interested how you come up with an understanding that it's a problem that needs to be addressed.
And do you think that introducing whole new module just to test one non-standard usecase is a good idea?

elihart:
I'm sure it's pretty rare. However, our team recently switched to using Buck as a build system, and so we are splitting our app into many modules. That's the only reason I noticed.

We have a base model we use that adds some airbnb specific support, so this base class has an annotated field. I noticed that models that extend it in our new modules don't work properly because of the module setup.

Few people would probably have this issue, but it's an issue for our team.

geralt-encore:
Thanks for the detailed response!
Just out of curiosity, are you using plain Buck or OkBuck? Is really worth it in comparison with Gradle? Sorry for offtopic.

elihart:
No worries about the off topic, its fun to talk about.

We're using OkBuck. It made it much easier to switch. Uber did the same thing recently and made some improvements to OkBuck and we were able to talk with them and follow how they did it, which helped a lot.

For a large app it seems to be very worth it. Gradle incremental builds were 1.5 - 2 minutes for us, and cleans were 5-7. Plus incremental builds weren't always reliable so you'd have to clean often.

We didn't notice a huge speed increase with Buck right away since we were still a single module project, but once we started splitting features into modules you notice a large speed increase. A  module for a decent sized feature builds in 5-10 seconds. And if you change multiple modules they build in parallel.

Resource changes still trigger a whole AAPT build though, so changing an xml file in a module results in a 1.5 minute build or so.

Buck has recently been adding support for Kotlin which will be great in the future. Using something like Anko instead of xml would help to get around the slow AAPT processing.

There are other caveats, like we're still figuring out our testing setup with Buck, but overall it's been a fairly smooth switch and very nice. Most of the work is in splitting up the app into modules, which can be a huge pain at first.

Once you have many modules Android Studio runs very slowly - gradle just doesn't handle it well. We instead are switching to plain Intellij with the Android plugin. This handles the modules much better and is faster.

geralt-encore:
Thanks for sharing! Very interesting experience =)

Issue #69, opened 2016-10-19
Generate a method to reset fields back to defaults


elihart:
a method like `model.reset()` could set all annotated fields back to their defaults. This is useful if a model is reused or changed and you don't want any previously set values carrying over to the new usage.

geralt-encore:
You mean type defaults or values assigned by a user during declaration? The first should be easy and I dunno about the second one.

elihart:
the first one, type defaults. definitely should be easy, just a little grunt work to enumerate them for each primitive type.

What do you think of this idea? I don't want to head down a path where the generated models include too many helper methods since it could bloat them and be too specific, but a single general method like this seems worthwhile

geralt-encore:
I understand your concerns, but this method looks pretty reasonable to have. I'll try to do it today.

PR #72, opened 2016-10-22
#69 Generate reset method

geralt-encore:
https://github.com/airbnb/epoxy/issues/69

### DIFF
--- a/epoxy-processor/src/main/java/com/airbnb/epoxy/EpoxyProcessor.java
+++ b/epoxy-processor/src/main/java/com/airbnb/epoxy/EpoxyProcessor.java
@@ -49,6 +49,7 @@ import static com.squareup.javapoet.TypeName.FLOAT;
 import static com.squareup.javapoet.TypeName.INT;
 import static com.squareup.javapoet.TypeName.LONG;
 import static com.squareup.javapoet.TypeName.SHORT;
+import static com.squareup.javapoet.TypeName.VOID;
 import static javax.lang.model.element.ElementKind.CLASS;
 import static javax.lang.model.element.Modifier.FINAL;
 import static javax.lang.model.element.Modifier.PRIVATE;
@@ -366,6 +367,7 @@ public class EpoxyProcessor extends AbstractProcessor {
         .addMethods(generateConstructors(info))
         .addMethods(generateSettersAndGetters(info))
         .addMethods(generateMethodsReturningClassType(info))
+        .addMethod(generateReset(info))
         .addMethod(generateEquals(info))
         .addMethod(generateHashCode(info))
         .addMethod(generateToString(info))
@@ -601,6 +603,36 @@ public class EpoxyProcessor extends AbstractProcessor {
     return builder
         .addStatement("return this")
         .build();
+  }
+
+  private MethodSpec generateReset(ClassToGenerateInfo helperClass) {
+    Builder builder = MethodSpec.methodBuilder("reset")
+        .addModifiers(Modifier.PUBLIC)
+        .returns(VOID);
+
+    for (AttributeInfo attributeInfo : helperClass.getAttributeInfo()) {
+      if (!attributeInfo.hasFinalModifier()) {
+        String attributeName = attributeInfo.getName();
+        TypeName attributeType = attributeInfo.getType();
+        if (attributeType == BOOLEAN) {
+          builder.addStatement("this.$L = false", attributeName);
+        } else if (attributeType == BYTE || attributeType == SHORT || attributeType == INT) {
+          builder.addStatement("this.$L = 0", attributeName);
+        } else if (attributeType == LONG) {
+          builder.addStatement("this.$L = 0L", attributeName);
+        } else if (attributeType == FLOAT) {
+          builder.addStatement("this.$L = 0.0f", attributeName);
+        } else if (attributeType == DOUBLE) {
+          builder.addStatement("this.$L = 0.0d", attributeName);
+        } else if (attributeType == CHAR) {
+          builder.addStatement("this.$L = '\\u0000'", attributeName);
+        } else {
+          builder.addStatement("this.$L = null", attributeName);
+        }
+      }
+    }
+
+    return builder.build();
   }
 
   private void writeError(Exception e) {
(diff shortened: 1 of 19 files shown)


epoxy-processor/src/main/java/com/airbnb/epoxy/EpoxyProcessor.java
```diff
@@ -603,6 +605,36 @@ private MethodSpec generateSetter(ClassToGenerateInfo helperClass, AttributeInfo
...
+        } else if (attributeType == BYTE || attributeType == SHORT || attributeType == INT) {
+          builder.addStatement("this.$L = 0", attributeName);
+        } else if (attributeType == LONG) {
+          builder.addStatement("this.$L = 0L", attributeName);
+        } else if (attributeType == FLOAT) {
+          builder.addStatement("this.$L = 0.0f", attributeName);
+        } else if (attributeType == DOUBLE) {
+          builder.addStatement("this.$L = 0.0d", attributeName);
+        } else if (attributeType == CHAR) {
+          builder.addStatement("this.$L = '\\u0000'", attributeName);
```
elihart: I didn't realize that char's default is `'\u0000', cool. From http://stackoverflow.com/questions/9909333/whats-the-default-value-of-char and http://docs.oracle.com/javase/tutorial/java/nutsandbolts/datatypes.html it seems this is equivalent to `0`. How about assigning it to 0 instead since that might be the better understood default value, and we could combine it with the byte, short, int block to simplify.

epoxy-processor/src/main/java/com/airbnb/epoxy/EpoxyProcessor.java
```diff
@@ -603,6 +605,36 @@ private MethodSpec generateSetter(ClassToGenerateInfo helperClass, AttributeInfo
...
+    for (AttributeInfo attributeInfo : helperClass.getAttributeInfo()) {
+      if (!attributeInfo.hasFinalModifier()) {
+        String attributeName = attributeInfo.getName();
+        TypeName attributeType = attributeInfo.getType();
+        if (attributeType == BOOLEAN) {
+          builder.addStatement("this.$L = false", attributeName);
+        } else if (attributeType == BYTE || attributeType == SHORT || attributeType == INT) {
+          builder.addStatement("this.$L = 0", attributeName);
+        } else if (attributeType == LONG) {
+          builder.addStatement("this.$L = 0L", attributeName);
```
elihart: Instead of duplicating `builder.addStatement("this.$L = X", attributeName);` we could have each block just state its value
```
String value;
if (attributeType == BOOLEAN) {
   value = "false";
}...
builder.addStatement("this.$L = " + value, attributeName);
```

To clean it up even more there could be a method `getDefaultValue`
```
private String getDefaultValue(TypeName attributeType) {
if (attributeType == BOOLEAN) {
   return "false";
} else ...

return null;
}
```
This could make this a bit easier to read, and reduce the chance of having any typos in all the repeated `"this.$L = 0L"` blocks.

epoxy-processor/src/main/java/com/airbnb/epoxy/EpoxyProcessor.java
```diff
@@ -603,6 +605,36 @@ private MethodSpec generateSetter(ClassToGenerateInfo helperClass, AttributeInfo
         .build();
   }
 
+  private MethodSpec generateReset(ClassToGenerateInfo helperClass) {
+    Builder builder = MethodSpec.methodBuilder("reset")
```
elihart: Can we actually have this return the type of the generated class? That way you could chain it
```
model.reset().value(...).value2(...)
```

This could be done in another PR, but we'd have to do it before releasing a next version

elihart:
Looks great! Just a few suggestions.

I'm thinking we should also have a `reset` method on the base `EpoxyModel`. This would reset layout to default as well as the visibility to `shown`. This seems more correct since if you call reset it seems like you'd expect these values to be reset as well.

Another benefit of this is that it reserves the `reset` method name as an EpoxyModel method, so people can expect it without looking at the generated class.

That could be a followup PR.

geralt-encore:
All points are valid. I'll fix them in this PR. Thanks for review!

geralt-encore:
If you want to add `reset` method to the `EpoxyModel` class then it will require workaround for `generateMethodsReturningClassType` since we need to do it manually. Is it ok?

elihart:
Good point.

I think adding an exception to `generateMethodsReturningClassType` and putting the `reset` method on the base class could be the best path forward.

This will make it so if anybody wants to add a custom behavior to `reset` they can override it in their model. Perhaps we can call `super.reset()` last in the generated `reset` method so any custom reset behavior isn't overridden by the generated default values.

Sound ok to you?

geralt-encore:
Yeah, I am totally ok with it. Will do it tomorrow.

elihart:
Thanks!

epoxy-processor/src/main/java/com/airbnb/epoxy/EpoxyProcessor.java
```diff
@@ -603,6 +604,44 @@ private MethodSpec generateSetter(ClassToGenerateInfo helperClass, AttributeInfo
...
+            getDefaultValue(attributeInfo.getType()));
+      }
+    }
+
+    return builder
+        .addStatement("return this")
+        .build();
+  }
+
+  private String getDefaultValue(TypeName attributeType) {
```
elihart: could be a static method

epoxy-processor/src/main/java/com/airbnb/epoxy/EpoxyProcessor.java
```diff
@@ -603,6 +604,44 @@ private MethodSpec generateSetter(ClassToGenerateInfo helperClass, AttributeInfo
...
+    String defaultValue;
```
elihart: no need for the `String defaultValue`, intermediate variable. could just return directly in each if block

epoxy-processor/src/main/java/com/airbnb/epoxy/EpoxyProcessor.java
```diff
@@ -603,6 +604,44 @@ private MethodSpec generateSetter(ClassToGenerateInfo helperClass, AttributeInfo
         .build();
   }
 
+  private MethodSpec generateReset(ClassToGenerateInfo helperClass) {
+    Builder builder = MethodSpec.methodBuilder("reset")
+        .addAnnotation(Override.class)
+        .addModifiers(Modifier.PUBLIC)
+        .returns(helperClass.getParameterizedGeneratedName())
+        .addStatement("super.reset()");
```
elihart: I mentioned before that maybe we should have `super.reset() be called last, so that if a model want's a different default value (like true for a boolean) they could override `reset` and have that method be called last so that our default values don't override there's.

What do you think? I know its a pain to change the tests so I can do this is you want.

geralt-encore: You mean if a model will extend generated class? Not sure that I get your suggestion correctly

PR #73, opened 2016-10-23
Add better examples to [Epoxy - Sample]

mshukla19:
Adds examples to [Epoxy - Sample] and fixes padding between items that have spanCount = totalSpanCount


epoxy-sample/src/main/java/com/airbnb/epoxy/SampleAdapter.java
```diff
@@ -96,11 +113,23 @@ public void onClick(View v) {
     }
   };
 
+  EditTextListener editTextListener = new EditTextListener() {
+    @Override
+    public void onTextEntered(String enteredText) {
+      Log.d(LOG_TAG, enteredText);
```
elihart: Maybe we could make this a little more interesting and the entered text would be centered on any of the colored blocks when `add` is clicked?

If the text is empty we add no text to the blocks. If the text is longer than a certain length we show an error. I think a better example for the error would be to set the error text on the model via the adapter. We should try not to have the error logic in the model and instead just delegate all the click behavior back here.

epoxy-sample/src/main/java/com/airbnb/epoxy/VerticalGridCardSpacingDecoration.java
```diff
@@ -45,6 +45,9 @@ public void getItemOffsets(Rect outRect, View view, RecyclerView parent, State s
       // Only item in row
       outRect.left = outerPadding;
       outRect.right = outerPadding;
+      // Assign top and bottom padding if there is just one item in the row
+      outRect.top = innerPadding;
```
elihart: Sorry, I'm missing the need for this. Top and bottom padding shouldn't need to change if there is only one item vs many. Could you explain it? If you want to add top/bottom padding to the edit text layout you can set it directly on `model_edit_text.xml`. I think views should generally be able to have different vertical padding if they need it

mshukla19: @elihart So in the current sample app, all the buttons don't have a vertical padding between them. I thought it would be better of symmetrical reasons to provide a default padding between all items (no matter the span count) and then if someone wants to increase it, they can do so via the xml

epoxy-sample/src/main/java/com/airbnb/epoxy/SampleAdapter.java
```diff
@@ -32,21 +40,30 @@
         .title(R.string.epoxy)
         .caption(R.string.header_subtitle);
 
+    EditTextModel editTextModel = new EditTextModel_()
+        .editTextListener(editTextListener)
+        .hint(R.string.text_hint);
+
     ButtonModel addButton = new ButtonModel_()
         .text(R.string.button_add)
+        .backgroundColor(ContextCompat.getColor(context, R.color.green))
```
elihart: You don't need to pass a context to the models, `bind` passes a view so just do `view.getContext()`

elihart:
I like the visual improvements, but don't want to merge until the items I identified are done. Closing until someone gets to them.

PR #77, opened 2016-11-15
Update dependencies

geralt-encore:


elihart:
@geralt-encore thanks!

PR #85, opened 2016-12-17
Use actual param names instead of generic paramN

geralt-encore:


epoxy-processor/src/main/java/com/airbnb/epoxy/ClassToGenerateInfo.java
```diff
@@ -102,14 +102,14 @@ private void collectMethodsReturningClassType(TypeElement originalClass) {
...
-                .getParameterTypes();
+            List<? extends VariableElement> params =
+                ((ExecutableElement) subElement).getParameters();
             String methodName = subElement.getSimpleName().toString();
             if (methodName.equals(RESET_METHOD) && params.isEmpty()) {
               continue;
             }
             if (params.size() == 1) {
-              TypeMirror param = params.get(0);
+              TypeMirror param = params.get(0).asType();
```
elihart: Since these are `Element` objects it seems like we can now call `getAnnotationMirrors` to get annotations on them so we could remove this work around for the @LayoutRes annotation.

PR #88, opened 2016-12-19
Cleanup after using actual names instead of paramN

geralt-encore:
Also, I tried to implement your idea about getting rid of specific layout method implementation, but for some reason, if a parameter is annotated then I am getting just generic `arg0` name instead of the actual `layoutRes` parameter name. You can try to investigate with it if you want.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,7 +4,7 @@ buildscript {
     jcenter()
   }
   dependencies {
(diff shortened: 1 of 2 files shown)


elihart:
That's strange about the `arg0` name issue for annotated params. I'd like to look into it, but am away from work the next few weeks. I'll check in out in January. Thanks for the clean up!

PR #89, opened 2016-12-20
Update README.md

wrparrish:
Just a small correction to fix the linking that leads to the 'Using View Holders' section.


elihart:
I just tried this linked and it seems to work fine. The header anchor is `## <a name="view-holders"/>Using View Holders` so linking to it with `#view-holders` works

wrparrish:
Noted!   Looking back at it,  i had recently clicked the viewholder link and then scrolled away,  mistakenly seeing that as a bug.  That's just how it is.   Apologies!

PR #91, opened 2016-12-27
Getting rid of layout method corner case

geralt-encore:
Despite the fact that with this solution we are losing `layout` method parameter name I still think that it is an overall improvement. The logic behind it became considerably simpler and it's always nice to avoid a corner case with more general solution.


elihart:
Great! I agree that this is an improvement. Cleaner code, and it seems more useful to keep the param annotations than the param name.

Thanks!

PR #94, opened 2017-01-06
Add removeAllModels() to EpoxyAdapter

lawrencebautista:
Thanks for the library Eli and AirBnb. I figured a removeAllModels() method would be useful.


elihart:
Looks great! Thanks for the update to tests and the simple adapter as well.

PR #95, opened 2017-01-09
remove apt dependency

eoinfogarty:
As of android gradle 2.2 all functionality of `android-apt` is fully provided they plugin itself via annotationProcessor


elihart:
Great, thanks!

Issue #99, opened 2017-01-27
Annotation processor duplicates vararg parameters incorrectly


elihart:
If there is a method on a model that takes varargs as an argument (`EpoxyModel<?> myMethod(Long... numbers)`) and the processor duplicates the method on a generated subclass the generated method will have an array in the parameters instead of varargs (`EpoxyModel<?> myMethod(Long[] numbers)`)

elihart:
https://github.com/airbnb/epoxy/pull/100

PR #100, opened 2017-01-27
#99 Correct varargs parameters

geralt-encore:
https://github.com/airbnb/epoxy/issues/99


geralt-encore:
Weird... 🤔
Tests are running successfully for me locally both from Android Studio and command line.

elihart:
You're fast!

In the tests I see one file named `ModelReturningClassTypeWithVarargs.java` (notice the lower case args) and the generated filed is `ModelReturningClassTypeWithVarArgs_.java` with an uppercase `Args`. I think that's it?

geralt-encore:
Looks like you were right. Not sure why did it work in a first place.

Issue #96, opened 2017-01-17
Clarify bind/unbind behavior.


awenger:
I observed that `unbind` is not always called before a view is rebound to **another** model (recycled). I assume this is working as intended and not a bug. However I think the documentation on this behavior could be improved. 

Knowing the described behavior I can interpret this information into the documentation:

https://github.com/airbnb/epoxy/blob/c2bdcd29d8d6acce0d47a0a5f7c14b3b4f07d937/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModel.java#L39-L60
https://github.com/airbnb/epoxy/blob/c2bdcd29d8d6acce0d47a0a5f7c14b3b4f07d937/README.md

However without this knowledge this was absolutely not obvious for me from the documentation. Especially because of the naming of the two methods `bind`/`unbind` and the observed behavior in most cases (most of the time `unbind` is called before the view is rebound).

elihart:
You're right, working as intended but still a bit confusing.

I believe that the Epoxy documentation does make sense in the context of RecyclerView. You said `before a view is rebound to another model (recycled)`, but that isn't actually the definition of recycled. Epoxy uses the RecylerView method `onViewRecycled` which documents the method as
```
Called when a view created by this adapter has been recycled.

A view is recycled when a RecyclerView.LayoutManager decides that it no longer needs to be attached to its parent RecyclerView. This can be because it has fallen out of visibility or a set of cached views represented by views still attached to the parent RecyclerView. If an item view has large or expensive data bound to it such as large bitmaps, this may be a good place to release those resources.

RecyclerView calls this method right before clearing ViewHolder's internal data and sending it to RecycledViewPool. This way, if ViewHolder was holding valid information before being recycled, you can call RecyclerView.ViewHolder.getAdapterPosition() to get its adapter position.
``` 

So the view is only recycled when it goes off screen. It is not recycled when it is bound to another model.

The `unbind` documentation is pretty clear that it is called when the view is recycled `Subclasses can override this if their view needs to release resources when it's recycled.`. I could have named this method `onViewRecycled` or something to make that more clear. I liked the symmetry and simplicity of bind/unbind, but it is misleading.

I think it's a bit late now to rename the `unbind` method, but I can update the javadoc and readme to be more explicit. Thanks for bringing this to my attention!

awenger:
Thank you for the clarification about when a View is recycled. 
[...]

PR #103, opened 2017-01-31
Clarify unbind behavior in documentation

elihart:
https://github.com/airbnb/epoxy/issues/96

@awenger The Readme was pretty clear about this behavior, but I tried to make it even more straightforward. I also updated the javadoc on the model.

Do you think this makes sense now?


awenger:
Great update to the documentation! I like that it's now much more explicit. :+1:

PR #105, opened 2017-02-02
Generate model for abstract class if class annotation is present

elihart:
@felipecsl @gpeal @seanabraham 

This forces the annotation processor to generate a model for abstract model classes if they have the `@EpoxyClassModel` attribute. Previously abstract classes never had a model generated for them since we couldn't know if they implemented every necessary method. This annotation approach allows users to tell the processor if they want a model generated.

We may like to use an abstract class for all our models to indicate that they shouldn't be instantiated directly, and to prevent people from accidentally using them instead of the generated class.

In this PR I also add a new feature to automatically generate the `createNewHolder` method on models that use view holders. This method was mostly boilerplate and we can generate it by looking at the type of the class and just returning a new object of that type.

For example, in our sample app we now have the class
```
@EpoxyModelClass
public abstract class ButtonModel extends EpoxyModelWithHolder<ButtonHolder> {
```
This is a small win to reduce boilerplate.

Besides that I did some cleanup - moving util methods to a ProcessorUtils class - and a little other refactoring clean up since the processor is getting large and more complicated.

### DIFF
--- a/epoxy-processor/src/main/java/com/airbnb/epoxy/EpoxyProcessor.java
+++ b/epoxy-processor/src/main/java/com/airbnb/epoxy/EpoxyProcessor.java
@@ -5,7 +5,6 @@ import com.airbnb.epoxy.ClassToGenerateInfo.MethodInfo;
 import com.google.auto.service.AutoService;
 import com.google.common.collect.ImmutableSet;
 import com.squareup.javapoet.ArrayTypeName;
-import com.squareup.javapoet.ClassName;
 import com.squareup.javapoet.JavaFile;
 import com.squareup.javapoet.MethodSpec;
 import com.squareup.javapoet.MethodSpec.Builder;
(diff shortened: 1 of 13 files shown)


epoxy-processor/src/main/java/com/airbnb/epoxy/EpoxyProcessor.java
```diff
@@ -458,6 +402,44 @@ private void generateClassForModel(ClassToGenerateInfo info) throws IOException
     return methods;
   }
 
+  /**
+   * Generates default implementations of certain model methods if the model is abstract and doesn't
+   * implement them.
+   */
+  private Iterable<MethodSpec> generateDefaultMethodImplementations(ClassToGenerateInfo info)
```
gpeal: @elihart What are the pros and cons of creating empty implementation vs having the generated class be abstract?

elihart: @gpeal I can't think of any good cases where you would want the generated class to be abstract. That doesn't seem useful to me, interested to hear if you know a case.

This isn't creating an empty implementation, it's generating a default implementation. In this case the default implementation is very likely what you would have written manually anyway, so it is helping to reduce boilerplate.

PR #109, opened 2017-02-05
Allow default layout res to be specified in class annotation

elihart:
This lets us get rid of even more boilerplate by not writing the `getDefaultLayout` method in models.

### DIFF
--- a/epoxy-annotations/src/main/java/com/airbnb/epoxy/EpoxyModelClass.java
+++ b/epoxy-annotations/src/main/java/com/airbnb/epoxy/EpoxyModelClass.java
@@ -1,16 +1,24 @@
 
 package com.airbnb.epoxy;
 
+import android.support.annotation.LayoutRes;
+
 import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 /**
- * Used to annotate EpoxyModel classes in order to generate a
- * subclass of that model with getters, setters, equals, and hashcode for the annotated fields.
+ * Used to annotate EpoxyModel classes in order to generate a subclass of that model with getters,
+ * setters, equals, and hashcode for the annotated fields, as well as other helper methods and
+ * boilerplate reduction.
  */
 @Target(ElementType.TYPE)
 @Retention(RetentionPolicy.CLASS)
 public @interface EpoxyModelClass {
+  /**
+   * A layout resource that should be used as the default layout for the model. If you set this you
+   * don't have to implement `getDefaultLayout`; it will be generated for you.
+   */
+  @LayoutRes int value() default 0;
 }
--- a/epoxy-annotations/src/main/java/com/airbnb/epoxy/EpoxyModelClass.java
+++ b/epoxy-annotations/src/main/java/com/airbnb/epoxy/EpoxyModelClass.java
@@ -20,5 +20,5 @@ public @interface EpoxyModelClass {
    * A layout resource that should be used as the default layout for the model. If you set this you
    * don't have to implement `getDefaultLayout`; it will be generated for you.
    */
-  @LayoutRes int value() default 0;
+  @LayoutRes int layout() default 0;
 }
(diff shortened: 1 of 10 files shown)


elihart:
@felipecsl @gpeal @seanabraham

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModel.java
```diff
@@ -287,11 +287,11 @@ public void onViewDetachedFromWindow(T view) {
...
   @Override
   public String toString() {
-    return getClass().getSimpleName() + "{" +
-        "id=" + id +
-        ", layout=" + getLayout() +
-        ", shown=" + shown +
-        ", addedToAdapter=" + addedToAdapter +
-        '}';
+    return getClass().getSimpleName() + "{"
+        + "id=" + id
```
gpeal: Is this the correct Airbnb java open source formatting? I can start using this in Lottie if that's the case.

epoxy-annotations/src/main/java/com/airbnb/epoxy/EpoxyModelClass.java
```diff
@@ -1,16 +1,24 @@
...
+ * boilerplate reduction.
  */
 @Target(ElementType.TYPE)
 @Retention(RetentionPolicy.CLASS)
 public @interface EpoxyModelClass {
+  /**
+   * A layout resource that should be used as the default layout for the model. If you set this you
+   * don't have to implement `getDefaultLayout`; it will be generated for you.
+   */
+  @LayoutRes int value() default 0;
```
gpeal: Is it possible to have this be a named annotation property but also be the default?
So, you could do:
`@EpoxyModelClass(R.layout.hello_world)`
or
`@EpoxyModelClass(layoutRes=R.layout.hello_world, futureProperty=true)`
?

elihart: No unfortunately :( If you want to omit the parameter name it has to be called `value` http://stackoverflow.com/questions/11786354/how-i-can-do-java-annotation-like-nameluke-with-no-attribute-inside-parenth

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModel.java
elihart: Yes, this is the formatting I got from Felipe. Although I think I may have changed a few things to my preference. In general we should follow the google java style guide.

epoxy-annotations/src/main/java/com/airbnb/epoxy/EpoxyModelClass.java
gpeal: @elihart what would it look like in the future if you want to add more parameters?

elihart: Hm, the easy option would be `@EpoxyModelClass(value=R.layout.hello_world, futureProperty=true)` which isn't great.

The other option is to rename the param from `value` to `layoutRes` in the future if we other params. That'd be a breaking change we'd have to force on people. Not too bad to upgrade if we provide a structural replace pattern, but also not great.

Another option is to create a whole new annotation. That could work depending on what the goal of the feature is.

Or we just do `@EpoxyModelClass(layoutRes=R.layout.hello_world)` now in the anticipation of adding future params.

That last option is the safest. It's not terrible to require the param name, but it would be a bit annoying. I supposed the param name would make the purpose clearer.

Thanks for bringing this up, hadn't considered it.

gpeal: @elihart Personally I'm a fan because it's the most explicit and is future proof. Luckily IntelliJ should be able to autocomplete the name as well.

elihart: Yeah, I think you're right. I'll change it. Thanks!

PR #112, opened 2017-02-06
Prepare for 1.6.1 release

elihart:


elihart:
@gpeal @felipecsl 

Going to update our app to test version once it is available on Maven. If all works well I'll merge.

CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
...
+* Allow the `createNewHolder` method to be omitted and generated automatically [(#105)](https://github.com/airbnb/epoxy/pull/105)
+* Generate a subclass for abstract model classes if the EpoxyModelClass annotation is present [(#105)](https://github.com/airbnb/epoxy/pull/105)
+* Allow strings as model ids [(#107)](https://github.com/airbnb/epoxy/pull/107)
+* Add instructions to readme for avoiding memory leaks [(#106)](https://github.com/airbnb/epoxy/pull/106)
+* Add model callbacks for view attached/detached from window, and onFailedToRecycleView [(#104)](https://github.com/airbnb/epoxy/pull/104)
+* Improve documentation on model unbind behavior [(#103)](https://github.com/airbnb/epoxy/pull/103)
+* Fix generated methods from super classes that have var args [(#100)](https://github.com/airbnb/epoxy/pull/100)
+* Remove apt dependency [(#95)](https://github.com/airbnb/epoxy/pull/95)
+* Add `removeAllModels` method to EpoxyAdapter [(#94)](https://github.com/airbnb/epoxy/pull/94)
+* Use actual param names when generating methods from super classes [(#85)](https://github.com/airbnb/epoxy/pull/85)
```
felipecsl: I found it useful when the changelog entries include either `Fix:` or `New:` in the beginning of each line, so it's easy to know whether its a new feature/change or bug fix :)
Also the `Releases` tab in Github is pretty dope 👍

elihart: Good idea. And I'm going to mark this as a release in the releases tab once merged 👍

elihart:
Realized there was an issue. Fixed with https://github.com/airbnb/epoxy/pull/113 and then bumped the version to 1.6.1

elihart:
I realized that in multimodule projects doing `@EpoxyModelClass(layout = R.layout.view)` doesn't work properly. I'm using butterknife's plugin to use `R2` so the value is final, but the processor ends up with the incorrect value, because I'm guessing it reference the R value of the module it is processed in, which changed after the modules are merged to get the final R class that the main project uses.

I'm looking into how butterknife handles this. If I can't solve it quickly I can release with the note that multi module projects are not supported

elihart:
Support library projects won't be trivial so I have just added a note to the readme for now so we can get this release out without more waiting.

Issue #118, opened 2017-02-09
Look at super classes for default layout if model does not define it


elihart:

geralt-encore:
You mean look for in parents `@EpoxyModelClass` annotations?

elihart:
https://github.com/airbnb/epoxy/pull/119 :)

PR #119, opened 2017-02-09
Look for default layout in parents if not specified

geralt-encore:
https://github.com/airbnb/epoxy/issues/118
The logic ended up more complex than I expected initially. If you have any ideas how to do it easily - I'll gladly fix it.

### DIFF
--- a/epoxy-processor/src/main/java/com/airbnb/epoxy/EpoxyProcessor.java
+++ b/epoxy-processor/src/main/java/com/airbnb/epoxy/EpoxyProcessor.java
@@ -493,7 +493,7 @@ public class EpoxyProcessor extends AbstractProcessor {
       return;
     }
 
-    EpoxyModelClass annotation = originalClassElement.getAnnotation(EpoxyModelClass.class);
+    EpoxyModelClass annotation = findClassAnnotation(originalClassElement);
     if (annotation == null) {
       throwError("Model must use %s annotation if it does not implement %s. (class: %s)",
           EpoxyModelClass.class,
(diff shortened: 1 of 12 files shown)


elihart:
Nice! It is a bit complex, but I don't think there is much getting around that. I just merged https://github.com/airbnb/epoxy/pull/120 which adds checks around the `annotation.layout()` calls.  To prevent having to do that mess I refactored what you have to only check the layout in one place. I'm also a fan of recursion :P

```
private EpoxyModelClass findClassAnnotationWithLayout(TypeElement classElement)
      throws EpoxyProcessorException {
    if (!isEpoxyModel(classElement)) {
      return null;
    }

    EpoxyModelClass annotation = classElement.getAnnotation(EpoxyModelClass.class);
    if (annotation == null) {
      return null;
    }

    try {
      int layoutRes = annotation.layout();
      if (layoutRes != 0) {
        return annotation;
      }
    } catch (AnnotationTypeMismatchException e) {
      throwError("Invalid layout value in %s annotation. (class: %s). %s: %s",
          EpoxyModelClass.class,
          classElement.getSimpleName(),
          e.getClass().getSimpleName(),
          e.getMessage());
      return null;
    }

    TypeElement superclassElement = getSuperClassElement(typeUtils, classElement);
    EpoxyModelClass annotationOnSuperClass = findClassAnnotationWithLayout(superclassElement);

    // Return the last annotation value we have so the proper error can be thrown if needed
    return annotationOnSuperClass != null ? annotationOnSuperClass : annotation;
  }
```

geralt-encore:
I'll look into your implementation! Just wanted to be sure that I understood the problem and did it correctly.

elihart:
yep, you understood perfectly, thanks for working on this!

PR #125, opened 2017-02-15
Prepare for release 1.7.5

elihart:
@felipecsl @seanabraham @gpeal @ngsilverman


elihart:
Had to bump to 1.7.1 to include a NPE fix in the processor https://github.com/airbnb/epoxy/pull/126

README.md
```diff
@@ -387,6 +388,27 @@ epoxyAdapter.setSpanCount(spanCount);
 layoutManager.setSpanSizeLookup(epoxyAdapter.getSpanSizeLookup());
 ```
 
+## Configuration
+
+Use the `@ModuleEpoxyConfig` annotation to specify module wide configuration options for all models in the module. This is a class annotation which you can put on any class in your module. Using an empty class is a fine pattern.
```
ngsilverman: nit: "module wide" and "in the module" is redundant. You could omit "in the module" and it would still be clear imo.

README.md
```diff
@@ -387,6 +388,27 @@ epoxyAdapter.setSpanCount(spanCount);
...
+  private EpoxyConfig() {
+  }
+}
+```
+
+These configuration options are used by the annotation processor to enforce model requirements or influence how generated models are created.
+
+This is especially useful for larger projects with many models and/or contributors. In these cases having a standardized pattern for models greatly reduces mistakes in creating and using models.
+
+If a config annotation is not found in the module then the default values are used.
```
ngsilverman: Might be worth mentioning what the defaults are.

ngsilverman:
👍

PR #132, opened 2017-02-22
Add diffing adapter to 2.0 branch

elihart:
This is the start of a redesign, to imagine what the next evolution of Epoxy should look like. It aims to address several problems I've noticed with the existing adapter.

1. The `models` list is often misunderstood and data changes are not notified correctly. It is very easy to accidentally notify the wrong change, or mix notify calls with diffing incorrectly. This causes hard to debug issues.
2. It is easy to forget to set an ID on a model. This prevents diffing from working correctly.
3. Keeping model ids constant across adapter instances is an extra step and can be forgotten. The ID must be saved as a constant and applied to the model manually.
4. Using diffing follows the same boilerplate pattern -> `enableDiffing()`, `models.clear()`, `notifyModelsChanged()`. We can eliminate this boilerplate.
5. Modifying the `models` list directly is too flexible. It can be abused, or even when used well in complicated adapters it can be hard to follow what is happening. This is because any callback or method can add/remove/change models and it can be hard to follow where all the changes happen from.
6. Model state can incorrectly carry over. For models that are saved as fields and reused this can be an issue if certain values are set on them and not completely cleared when they are used again.
7. High learning curve. Working with a `models` list directly is powerful and flexible, but it is also a strange paradigm and there are many ways to approach it incorrectly.

I aim to fix all of these issues by creating a new Epoxy adapter that requires diffing and enforces a strict paradigm of how models are created. The models list can not be changed directly, it can only be added to and cannot be directly accessed. It can only be added to in one method - `buildModels`. This method starts with a completely empty state, so every time it is called it should completely rebuild the state of the models based on current data.

This is similar to the React approach. The idea is that `buildModels` is a function that takes a dataset and produces a set of models that should be shown for that state. This should make it easier to understand what the adapter is doing, allow for testing, and greatly reduce the chance for developer error.

Additionally, this introduces a new annotation, `AutoModel`, which can be used to annotate fields on the adapter. Previously models that were static and reusable were saved as final fields. This led to problems 3 and 6 above. The `AutoModel` allows the adapter to set a stable id across adapter instances, as well as automatically create a new instance of the model every time models are rebuilt so that you are guaranteed to start with a consistently empty state. If an id is overwritten, or not otherwise set on other models, we can throw, reducing the chance of problem 2. This doesn't stop people from setting incorrect ids, but the differ will (and already does) prevent duplicate ids.

For an example of this new paradigm, see the change to the SampleAdapter - https://github.com/airbnb/epoxy/pull/132/files#diff-dc704cc94b11359816d96f96d4d9b4ed

The main downside to this is performance. Recreating models and diffing for every change is inefficient compared to modifying models in place and notifying specific changes when you know them. In my profiling the performance drop is negligible (dealing with hundreds of models in the sample app only takes ~1ms to build/diff them) and the boost to developer productivity is well worth the tradeoff. I'm planning to leave the original EpoxyAdapter in the library so it can still be used if the extra performance is needed. Also, the way I've currently set up the new adapter should allow for processing Diffing in the background in the future if we need the performance gain.

I want to make sure that the interface for this new adapter is as easy to use as possible. We should think through design decisions carefully so it is clear how to use this adapter, and that we aren't too limiting in how people can use it. This is our chance to change anything we want about Epoxy!

There are a bunch of aspects to consider, including:

1. How people signal when models should be rebuilt. In my draft I have a `requestModelUpdate` method which signals to the adapter to build models at some time in the future. I think it may be useful to debounce this so developers can feel free to call `requestModelUpdate` multiple times in a row without worrying about performance. In the `TypedDiffAdapter` you can call `setData` on the adapter, instead of `requestModelUpdate`, and that data will be passed on to the `buildModels` method.

2. What the `buildModels` method should be called and how people interact with it. Right now implementations call `add` to add models in the order that should be displayed. I think getting the naming and usability perfect here is really important.

3. Should we allow only generated models? It's more restrictive, but guarantees models have a hashcode generated for them.

4. Should we deprecate/remove the original Epoxy Adapter? Doing that would enforce these better patterns and allow us to clean up a lot of code, but it would make the transition harder and also remove the flexibility and power of that adapter.

5. What should the adapter be named? What about the typed adapter?

6. What should the `AutoModel` annotation be named? Should it add any other functionality?

7. Should we not allow hiding models in this new adapter? It doesn't make sense since they should just not be added in that case and having it as an option could cause confusion.

8. The current implementation makes it hard to get a model's position since there is no list to access. I have ideas on how to generate a click listener that returns the clicked model, the view, and the position in the adapter, or we could just go back to using a list like we used to.

Any other thoughts or concerns?


ngsilverman:
This sounds great. Some thoughts:

1. Could we assume the data is immutable and make `requestModelUpdate` and `setData` one and the same?
2. Consider using an object passed in as a parameter of `buildModels` to add the models. Similar to how `onCreateView(...)` has a `LayoutInflater` parameter and `onCreateOptionsMenu(...)` has a `MenuInflater`. This should help clarify what is supposed to happen within that method.
I wish there was a way to also make the `@AutoModels` a part of the method signature somehow. That's one thing that confused me when I looked at `SampleAdapter`, I wondered where and when the models were being instantiated.
3. I don't think we should enforce generated models. But we should make sure that people understand both options.
4. If the only downside to the new adapter is an insignificant performance loss, and every other legitimate use case is still supported, I would be in favor of removing the old adapter.
5. `AutoEpoxyAdapter`? Something that makes it sound easy and better.
6. Maybe something like `@Instantiate` could help document their behavior at the same time?
7. Agreed. I don't think hiding is needed anymore.

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModel.java
```diff
@@ -184,6 +184,16 @@ public final int getLayout() {
     return this;
   }
 
+  public void addTo(DiffAdapter adapter) {
+    adapter.add(this);
+  }
+
+  public void addIf(boolean condition, DiffAdapter adapter) {
```
ngsilverman: 👍

geralt-encore:
@elihart looks very good! Recently I was doing a fairly complex adapter so I had to work with `models` directly in some places and I didn't look good. I clearly see how it could be improved with your proposal. Let me know if you need any help!

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModel.java
geralt-encore: Would be nice if instead of `boolean` parameter it was a predicate function so you can provide a lambda. Or it could be an overload as well.

elihart: @geralt-encore I like the idea of a predicate as an overload!

elihart:
@geralt-encore Thanks for looking and for the feedback! I think I have most of the coding work under control, but I really appreciate you looking over PR's and commenting so I can get a sense of how this will affect how other people use Epoxy and how the community likes it.

elihart:
From Nick Reynolds:

I really like the changes to Epoxy. It greatly decreases the room to make mistakes and helps ensure consumers get the best of epoxy w.r.t diffing etc. Down side - it does abstract more of the adapter and list view, this means we get further away from "real" Android which has its advantages and disadvantages. Also - I see in the example the callbacks exist in the view model, I'd personally prefer that logic be kept further in the adapter. On the host team we've gained a ton by keeping as much as we can in the adapter so it's compartmental and re-useable across LYS and ML. That's had the added benefit of keeping our Fragments down to the bare minimum of networking and UI interactions

PR #136, opened 2017-03-02
Add options array to EpoxyAttribute annotation

elihart:
This introduces a set of options that can be set as an array on an EpoxyAttribute. It deprecates and replaces the use of hash=false. Instead you would now say `@EpoxyAttribute({DoNotHash})`. it also introduces two new options. AllowMissingHash and NoGetter. These are detailed in the code comments.

There were a few reasons for this. Importantly, something like `AllowMissingHash` needed to be added for the special case of ignoring the package config of failing on attributes without hashcode (like the AutoValue case). `hash=true` does not work in that case because it is the default, and it isn't possible to use Boolean as an annotation param. This options approach makes it easy to add support for this.

Another, I want to allow for other adding more params to the annotation without having a long list -> `EpoxyAttribute(hash = false, setter = false, getter = false)` etc.

I also think this reads a bit nicer.

@gpeal @felipecsl @seanabraham @ngsilverman

### DIFF
--- a/epoxy-annotations/src/main/java/com/airbnb/epoxy/EpoxyAttribute.java
+++ b/epoxy-annotations/src/main/java/com/airbnb/epoxy/EpoxyAttribute.java
@@ -7,25 +7,77 @@ import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 /**
- * Used to annotate fields on EpoxyModel classes in order to generate a
- * subclass of that model with getters, setters, equals, and hashcode for the annotated fields.
+ * Used to annotate fields on EpoxyModel classes in order to generate a subclass of that model with
+ * getters, setters, equals, and hashcode for the annotated fields.
  */
 @Target(ElementType.FIELD)
 @Retention(RetentionPolicy.CLASS)
 public @interface EpoxyAttribute {
   /**
+   * Options that can be included on the attribute to affect how the model's generated class is
+   * created.
+   */
+  enum Option {
+    /**
+     * A getter is generated for this attribute by default. Add this option to prevent a getter from
+     * being generated.
+     */
+    NoGetter,
+    /**
+     * A setter is generated for this attribute by default. Add this option to prevent a setter from
+     * being generated.
+     */
+    NoSetter,
+    /**
+     * By default every attribute's hashCode method is called when determining the model's state.
+     * This option can be used to exclude an attribute's hashCode from contributing to the state.
+     * <p>
+     * This is useful for objects that may change without actually changing the model's state. A
+     * common case is an anonymous click listener that gets recreated with every bind call.
+     * <p>
+     * When this is used, the attribute will affect the model state solely based on whether it is
+     * null or non null.
+     */
+    DoNotHash,
+    /**
+     * This is meant to be used in conjunction with {@link PackageEpoxyConfig#requireHashCode()}.
+     * When that is enabled every attribute must implement hashCode. However, there are some valid
+     * cases where the attribute type does not implement hashCode, but it should still be hashed at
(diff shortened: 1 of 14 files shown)


epoxy-sample/src/main/java/com/airbnb/epoxy/models/ButtonModel.java
```diff
@@ -12,11 +12,13 @@
...
 import butterknife.BindView;
 
+import static com.airbnb.epoxy.EpoxyAttribute.Option.DoNotHash;
+
 /** This model class gives an example of how to use a view holder pattern with your models. */
 @EpoxyModelClass(layout = R.layout.model_button)
 public abstract class ButtonModel extends EpoxyModelWithHolder<ButtonHolder> {
   @EpoxyAttribute @StringRes int text;
-  @EpoxyAttribute(hash = false) OnClickListener clickListener;
+  @EpoxyAttribute({DoNotHash}) OnClickListener clickListener;
```
ngsilverman: nit: you don't need the curly brackets when there's only one item.

epoxy-annotations/src/main/java/com/airbnb/epoxy/EpoxyAttribute.java
```diff
@@ -7,25 +7,77 @@
...
+     * tell the processor to let it pass the hashCode validation.
+     * <p>
+     * An example case is AutoValue classes, where the generated class correctly implements hashCode
+     * at runtime.
+     * <p>
+     * If you use this it is your responsibility to ensure that the object assigned to the attribute
+     * at runtime correctly implements hashCode. If you don't want the attribute to contribute to
+     * model state you should use {@link Option#DoNotHash} instead.
+     */
+    AllowMissingHash
```
ngsilverman: The relation between `PackageEpoxyConfig`'s `requireHashCode` and this might be clearer if instead of an `AllowMissingHash` option this also has a `requireHashCode` attribute. One is global, the other local.

ngsilverman:
@gpeal could be `@EpoxyAttribute(value = { NoGetter, DoNotHash }, otherAttribute = ...)`, but that gets a bit weird. Personally I don't mind the longer list of attributes. In most cases you wouldn't need to specify any.

elihart:
@ngsilverman Ah, thanks for the tip about the single value not needing braces, forgot about that!

I believe what @gpeal is saying is that by making the options default we have to name the param `value`, and doing `@EpoxyAttribute(value = { NoGetter, DoNotHash }, otherAttribute = ...)` for adding other params is confusing since `value` is not a clear param name.

I'm not too worried about that case. I haven't thought of any param cases that can't be booleans. If we really need one later on I think having `value` is acceptable. I'd rather optimize this very common case and be able to do `@EpoxyAttribute(DoNotHash)`, instead of needing `@EpoxyAttribute(options = {DoNotHash})` every single time, instead of trying to future proof for something we don't that we we'll need.

epoxy-annotations/src/main/java/com/airbnb/epoxy/EpoxyAttribute.java
elihart: `requireHashCode` doesn't make as much sense on the attribute, since a hashCode is always required (unless DoNotHash is used). Maybe a better name for the attribute option could be `IgnoreRequireHashCode` so that the name ties back better to the config option. I don't mind it being verbose since it should rarely be used, and I think having a name that is clear is important here. What do you think of `IgnoreRequireHashCode` vs `AllowMissingHash` (or `AllowMissingHashCode`)? I want the name to say "I know this field type doesn't implement hashCode, but it's ok and shouldn't fail the hashCode requirement check because I know the object used at runTime _will_ properly implement hashCode".

PR #138, opened 2017-03-02
Fixed #137

### LINKED ISSUE #137: [OperatorPrecedence] Use grouping parenthesis to make the operator precedence explicit
In Epoxy Processor 
`builder.beginControlFlow("if ($L != null && that.$L == null"
                + " || $L == null && that.$L != null)",`
No Operator Precedence exists
[Reference](http://errorprone.info/bugpattern/OperatorPrecedence)

engincancan:
Used grouping parenthesis to make the operator precedence explicit

### DIFF
--- a/epoxy-processor/src/main/java/com/airbnb/epoxy/EpoxyProcessor.java
+++ b/epoxy-processor/src/main/java/com/airbnb/epoxy/EpoxyProcessor.java
@@ -662,8 +662,7 @@ public class EpoxyProcessor extends AbstractProcessor {
               .endControlFlow();
         }
       } else {
-        builder.beginControlFlow("if (($L != null && that.$L == null)"
-                + " || ($L == null && that.$L != null))",
+        builder.beginControlFlow("if (($L == null) != (that.$L == null))",
             name, name, name, name)
             .addStatement("return false")
             .endControlFlow();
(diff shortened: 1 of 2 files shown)


epoxy-processor/src/main/java/com/airbnb/epoxy/EpoxyProcessor.java
```diff
@@ -662,8 +662,8 @@ private MethodSpec generateEquals(ClassToGenerateInfo helperClass) {
               .endControlFlow();
         }
       } else {
-        builder.beginControlFlow("if ($L != null && that.$L == null"
-                + " || $L == null && that.$L != null)",
+        builder.beginControlFlow("if (($L != null && that.$L == null)"
```
elihart: I believe we could also simplify this to `if(($L == null) != (that.$L == null))` if you're so inclined

engincancan: It is a simply XOR operation. I didn't quite catch that. Thanks for pointing out :) I will make the change and update it.

elihart:
Thanks!

I'm in the middle of a very large refactor and update of the processor (see https://github.com/airbnb/epoxy/pull/139) so I'm not actually going to merge your PR because that will cause me a bunch of merge headaches. I will incorporate your change into the processor update though (doing it now! :) )

Thanks again!

elihart:
Your change is incorporated into https://github.com/airbnb/epoxy/pull/140 :)

https://github.com/airbnb/epoxy/pull/140/files#diff-e7b2767f9be47977fb1b4c944e9919d2R599

PR #140, opened 2017-03-03
Support  onBind/onUnbind listeners on generated models

elihart:
Generated models now include `onBind` and `onUnbind` builder style method that takes a callback listener. When the model hits these lifecycle events the listener will be called with the model and view that was bound.

This looks like
```
new LoaderModel_()
       .onBind((model, view) -> loadMore())
       .addTo(this);
```

The goal is to avoid having one adapter method for handling onModelBound, and instead keeping the logic where the model is created.

This involved quite a bit change to the processor to generate the correct interface, store the listener fields, update hashCode/equals, etc. Also had to update all the tests.

Since I was updating all the tests and figuring out a good way to do this, I realized a better way to do the model click listener handling as well without storing the viewholder and bound object on the model. This seems cleaner and safer.


elihart:
@gpeal @ngsilverman

ngsilverman:
Love it. 👍

Issue #133, opened 2017-02-23
EpoxyAttribute generate problem for Kotlin


minibugdev:
I try to apply `EpoxyAttribute` to field in Kotlin, like below:
``` kotlin
@EpoxyModelClass(layout = R.layout.model_epoxy)
abstract class TestEpoxyModel(@EpoxyAttribute @ColorInt var color: Int) : EpoxyModel<View>() {
	override fun bind(view: View?) {
		view?.setBackgroundColor(color)
	}
}
```

And got error on compile annotation, like below:
```
com.airbnb.epoxy.EpoxyProcessorException: EpoxyAttribute annotations must not be on private or static fields. (class: TestEpoxyModel, field: color)
e: Some error(s) occurred while processing annotations. Please see the error messages above.
```

geralt-encore:
The problem is that in Kotlin backing fields of properties are private which are not allowed by epoxys annotation processor.

@elihart I've solved the same issue in [StorIO](https://github.com/pushtorefresh/storio).
The solution was to allow annotate not only actual fields but methods as well - getters in our case.
It looks like that:
```
@StorIOSQLiteType(table = "tweets") data class Tweet @StorIOSQLiteCreator constructor(@get:StorIOSQLiteColumn(name = "author") val author: String,
                                                  @get:StorIOSQLiteColumn(name = "content") val content: String)
```
I am not sure that it is the best solution, but it was at least the easiest one to integrate into current codebase. As a bonus it also allows to use annotation on AutoValue classes since they can be used on methods.

minibugdev:
@geralt-encore Thank you for solution but its doesn't work both `@get:EpoxyAttribute` and `@field:EpoxyAttribute`.

For my solution, I added `@JvmField @EpoxyAttribute` to the field and it work properly.

geralt-encore:
@minibugdev sorry, I wasn't clear enough. It wasn't a solution for you case - it was a proposal to @elihart how it could be possibly solved in a more kotlin way.
`@JvmField` works as a temporary workaround tho.

elihart:
@geralt-encore Thanks for chiming in. I don't have much experience with Kotlin unfortunately so I'm glad you know what's going on.

I'm glad it can be solved with `JvmField `. Very interesting idea to allow annotating getter methods with `EpoxyAttribute`. I'm not ready to spend the time going down that path myself right now, but would be open to a PR if somebody really wants it.

geralt-encore:
[...]

PR #144, opened 2017-03-05
Allow to annotate private fields based on their getters and setters

geralt-encore:
https://github.com/airbnb/epoxy/issues/133
Well, it took much more that I thought initially. Now it is possible to apply `@EpoxyAttribute` directly on Kotlin properties without `@JvmField`.

There is quite some code duplication with these changes, but I didn't find a better solution. Please, feel free to suggest any improvements, since I am not really happy with it myself.

Also, there is a weird issue that a couple of my tests are failing. I am checking that compilation fails with a specific error, but it contains only final compilation error in generated code without any mentions of the initial error. And I am pretty sure that I logged it correctly since the code I am checking with these tests is pretty trivial. So it would be nice if you could help me to deal with it.


epoxy-processor/src/main/java/com/airbnb/epoxy/AttributeInfo.java
```diff
@@ -142,6 +155,43 @@ private boolean isFieldPackagePrivate(Element attribute) {
   }
 
   /**
+   * Checks if the given private field has getter and setter for access to it
+   */
+  private void findGetterAndSetterForPrivateField(ErrorLogger errorLogger) {
```
elihart: To simplify, what if we assumed/enforced that the getter/setter have the same name as the field. Our generated getter/setter will already share that name so it seems like it would be a good thing for consistency anyway. Does this make sense in kotlin?

I previously made a `ProcessorUtilsgetMethodOnClass` method that you may be able to reuse to simplify this, although you may want to add some modifier checking into that method (to check !static, etc).

epoxy-processor/src/main/java/com/airbnb/epoxy/GeneratedModelWriter.java
```diff
@@ -215,29 +214,51 @@ private static ParameterizedTypeName getModelClickListenerType(ClassToGenerateIn
...
+                + "    public void onClick($T v) {\n"
+                + "    $L.onClick($T.this, object,\n"
+                + "        holder.getAdapterPosition());\n"
+                + "    }\n"
+                + "    public int hashCode() {\n"
+                + "       // Use the hash of the original click listener so we don't change the\n"
+                + "       // value by wrapping it with this anonymous click listener\n"
+                + "       return $L.hashCode();\n"
+                + "    }\n"
+                + "  });\n", attribute.setter(), viewClickListenerType,
```
elihart: is using the setter vs assigning the field directly the only difference? Maybe we can keep the main code block the same, but assign it to a local variable.

Then in an if/else we can either call the super setter or set the super field. That way we could reuse most of the code

epoxy-processor/src/main/java/com/airbnb/epoxy/GeneratedModelWriter.java
```diff
@@ -623,6 +656,41 @@ private void addEqualsLineForType(Builder builder, boolean useObjectHashCode, Ty
     }
   }
 
+  private void addEqualsLineForTypeUsingGetter(Builder builder, boolean useObjectHashCode,
```
elihart: To avoid duplicating hash/equals methods, what if we made a little method like `attributeGetterCode` and `attributeSetterCode`, which would so something like
```
if(isPrivateAttribute){
   return attribute.name + "()"; // Assuming we enforce setter name has to equal field name
} else {
   return attribute.name;
}

Setter might be trickier but maybe possible?

Comment (elihart) [inline epoxy-processor/src/main/java/com/airbnb/epoxy/GeneratedModelWriter.java]: This could also be a method on the attribute info -> `attributeInfo.getAccessorCode()`

I think this approach could clean up the code in all the usages above too

**Comment by elihart** [comment]
Not having the right compilation error is weird. Probably related to my changes to group all errors at the end. I'll try running these tests locally and see if I can debug.

I was thinking recently that we could make a script that updates failing processor tests with the actual output. 

The script could take the filepath of the report, like `repos/epoxy/epoxy-processortest/build/reports/tests/debug/index.html`, find tests that failed because of processed code mismatch (it says something like "actual source" and "expected source", and it could update the file of the expected source to contain the actual code.

This would be very helpful when we change a small thing in the processor that affects all processor tests. We know the actual code is right, but would have to manually update the test code. I know you've had to do that in the past :P, and I 've been wasting a lot of time on it recently with my refactor.

File: epoxy-processor/src/main/java/com/airbnb/epoxy/AttributeInfo.java (hunk 1 above)
Comment (geralt-encore) [inline epoxy-processor/src/main/java/com/airbnb/epoxy/AttributeInfo.java]: The thing is that we can't enforce it since Kotlin generate automatically getters and setters for properties with this get/set naming pattern.

**Comment by elihart** [comment]
Fyi https://github.com/airbnb/epoxy/pull/145 will cause a small merge conflict, but it is just adding a view param to the model click listener in the generated code in tests.

If you prefer I can wait to merge it until after this PR is merged

**Comment by geralt-encore** [comment]
Go ahead! I'll resolve a merge conflict. It's not a big deal.

**Comment by elihart** [comment]
Thanks. I'm gonna try my hand at a ruby script to automatically update processor tests, so maybe don't try changing your tests for now (if you change processor code)

**Comment by elihart** [comment]
I think I have the script working (albeit it's ugly :P)

```
require 'rubygems'
require 'nokogiri'

def updateTestClass(test_class_result)
  page = Nokogiri::HTML(open(test_class_result))

  page.css('pre').each do |preBlock|
    if preBlock.include? "Source declared the same top-level types of an expected source"
      next
    end

    expected_source_file_path = /Expected file: <([^>]*)>/m.match(preBlock).captures[0]

    expected_source_file_path.sub! "build/intermediates/classes/test/debug", "src/test/resources"

    actual_source_match = /Actual Source:[\s]*=*[\s]*(package.*?})[\s]*at com\.google/m.match(preBlock)
    actual_source = actual_source_match.captures[0]

    File.open(expected_source_file_path, "w") do |f|
      f.write actual_source
    end
  end
end

module_name = ARGV[0]
classes_directory_name = module_name + "/build/reports/tests/debug/classes"

Dir.glob(classes_directory_name + "/*.html") do |test_class_result|
  updateTestClass(test_class_result)
end
```

you would first run `./gradlew test`, and if tests fail you can run this script to replace the failing expected source with the actual source.

You just provide the module name of the tests, like:
```ruby TestUpdater.rb epoxy-processortest```

I've only tested briefly. I need to go through and clean it up, but if you want to try it for you own usage to save some time, go for it!

I'm not great with ruby, so if you see opportunity for improvement, please go for it. I'll put this up as a PR later when I have time.

File: epoxy-processor/src/main/java/com/airbnb/epoxy/AttributeInfo.java (hunk 1 above)
Comment (elihart) [inline epoxy-processor/src/main/java/com/airbnb/epoxy/AttributeInfo.java]: ah makes sense. oh well, no big deal

File: epoxy-processor/src/main/java/com/airbnb/epoxy/AttributeInfo.java [hunk 4]
```diff
@@ -142,6 +155,43 @@ private boolean isFieldPackagePrivate(Element attribute) {
...
+    for (Element element : classElement.getEnclosedElements()) {
+      if (element.getKind() == ElementKind.METHOD) {
+        ExecutableElement method = (ExecutableElement) element;
+        String methodName = method.getSimpleName().toString();
+        // check if it is a valid getter
+        if (methodName.equalsIgnoreCase(String.format("get%s", name))
```
elihart: should this not be `equalsIgnoreCase` since with a different case it would be a different method?

epoxy-processor/src/main/java/com/airbnb/epoxy/AttributeInfo.java
```diff
@@ -142,6 +155,43 @@ private boolean isFieldPackagePrivate(Element attribute) {
...
+            || methodName.equalsIgnoreCase(String.format("is%s", name))
```
elihart: ah your tests are failing because your order of operations for the logic here is messed up. You need `( )` around your `||` operation
```
(methodName.equalsIgnoreCase(String.format("get%s", name))
            || methodName.equalsIgnoreCase(String.format("is%s", name)))
```
Tricky!

Btw I recently learned how to debug annotation processing and it is very helpful. I used it to figure this out https://blog.xmartlabs.com/2016/03/28/Debugging-an-Annotator-Processor-in-your-project/

epoxy-processortest/src/test/java/com/airbnb/epoxy/ModelProcessorTest.java
```diff
@@ -498,4 +487,179 @@ public void modelWithViewClickListener() {
...
+    assert_().about(javaSource())
+        .that(model)
+        .processedWith(new EpoxyProcessor())
+        .compilesWithoutError()
+        .and()
+        .generatesSources(generatedModel);
+  }
+
+  @Test
+  public void modelWithViewPrivateClickListener() {
```
elihart: Thanks for adding so many tests!

epoxy-processor/src/main/java/com/airbnb/epoxy/AttributeInfo.java
geralt-encore: wow, thanks! That makes sense now. Yeah, I've seen this article, but I was pretty sure that the code is alright and the issue was in errors handling. Shame on me :P

epoxy-processor/src/main/java/com/airbnb/epoxy/AttributeInfo.java
geralt-encore: You are right, it was just the easiest approach. I'll handle it properly. Thanks for pointing out!

epoxy-processor/src/main/java/com/airbnb/epoxy/GeneratedModelWriter.java
geralt-encore: Yep, just wasn't sure how to proceed with it. I managed to solve it. Will include into next commit.

epoxy-processor/src/main/java/com/airbnb/epoxy/GeneratedModelWriter.java
geralt-encore: I've made it in a different way. `addEqualsLineForTypeUsingGetter` is not needed anymore.

epoxy-processortest/src/test/java/com/airbnb/epoxy/ModelProcessorTest.java
geralt-encore: No problem, I am jus trying to cover every possible case. This is the purpose of tests after all :)

geralt-encore:
Glad that we were able to deal with it so fast :)

jbmlaird:
Can you give me an example of how this is meant to work @geralt-encore ?

Using version 2.1.0 gives me the error mentioned in this thread with my code:

```
@EpoxyModelClass(layout = R.layout.model_text)
abstract class TextModel(@EpoxyAttribute val text: String?) : EpoxyModel<TextView>() {
    override fun bind(view: TextView?) {
        super.bind(view)
        view?.text = text
    }
}
```

geralt-encore:
@jbmlaird `text` has to be `var` instead of `val` since in generated code we need access to both getter and setter and in `val` case only getter exists.

PR #146, opened 2017-03-06
Add ruby script to auto update processor test resources

elihart:
This should make our lives much easier in the future. I should have invested in doing this before I started the recent refactor 😞 

fyi @geralt-encore this seems to work well! I simplified it so it analyzes all modules with no need for a param when you run it. I hope it can make your contributions more painless!

I'm not great with ruby or regex's so I'm open to any suggestions on how to do this better!

### DIFF
--- a/CONTRIBUTING.MD
+++ b/CONTRIBUTING.MD
@@ -1,0 +1,7 @@
+# Epoxy is an Open Source Project
+
+Pull requests are welcome! We'd love help improving this library.
+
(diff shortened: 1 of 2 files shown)


elihart:
@geralt-encore Ah, I pulled down your branch to take a look at your failing tests and realized this script doesn't fix the tests correctly in your case. I'll get a fix up soon so you don't have to mess with fixing the tests from the merge conflicts

elihart:
nvm it works fine :P

geralt-encore:
cool!

PR #147, opened 2017-03-06
Reduce duplication of private getter/setter code

elihart:
@geralt-encore This lets us remove the `if (attribute is private) else...` code generation blocks by having a little method to generate the setter/getter code for each case. What do you think?


geralt-encore:
Nice 👍

PR #151, opened 2017-03-09
Add programmatic view support

elihart:
@gpeal @ngsilverman


elihart:
I will follow up later with a processor change to create a view type when one isn't explicity defined

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModelWithView.java
```diff
@@ -0,0 +1,58 @@
...
+ * based on the generated model's name. This means that all models of that type should be able to
+ * share the same view, but the view won't be shared with models of any other type.
+ * <p>
+ * The generated view type will be negative so that it cannot collide with values from layout
+ * resources, which are used in normal Epoxy models. However, it is possible for generated view
+ * types to collide if models from different libraries are used together, since the model processor
+ * can only guarantee that view types are unique within a library. If you need to use {@link
+ * EpoxyModelWithView} models from different libraries or modules together you can define a view
+ * type manually to avoid this small collision chance. A good way to manually create value is by
+ * creating an R.id. value in an ids resource while, which will guarantee a unique value.
```
ngsilverman:
```
class ViewTypeManager {
  private final Map<Class<? extends EpoxyModel>, int> viewTypeMap = new HashMap<>();

  public int getViewType(EpoxyModel model) {
    Class<? extends EpoxyModel> modelClass = model.getClass();
    if (!viewTypeMap.contains(modelClass)) {
      int viewType = -viewTypeMap.size()-1;
      viewTypeMap.put(modelClass, viewType);
      return viewType;
    } else {
      return viewTypeMap.get(modelClass);
    }
  }
}
```

How about using a singleton to generate unique View types for every model class?

elihart: yep, something like that would definitely work. The main issue is this can't work across multiple modules. They would instantly clash since they all start at -1 and go down. By using a hash we at least have a pretty high chance of not having a collision. The other thing I considered is that a hash of the class name is reproducible, whereas something like this could arbitrarily result in a different value if more model classes are added/deleted/implementation changes or whatever. it seems useful to have a fairly reproducible viewtype so if people do have a collision it is reproducible.

On the other hand, if we did it this way we could basically rule out library model sharing because they would clash. That might be safer than saying it _probably_ won't be a hash collision. I think I can add debug checking for models with a view type collision being used in the same adapter at runtime though, and alert the user to use a manual id in that case, so that takes the risk away from random unknown collisions.

What do you think?

ngsilverman: I should have clarified: My suggestion is to use the singleton at runtime, not during the annotation processing.

elihart: oh wow, brilliant! I think that could be perfect. I had too much tunnel vision on accomplishing this during processing, but your idea is much more simple and elegant

Issue #86, opened 2016-12-17
No support for programmatically generated views


gumil:
No option to use programmatically generated views to be used with the EpoxyAdapter

elihart:
I like the idea of supporting this. I'm thinking that a model that subclasses `EpoxyModel` could support it. Something like
```
abstract class EpoxyModelWithView<T> extends EpoxyModel<T> {

         @Override
         protected int getDefaultLayout() {
             return 0;
         }

         abstract int getItemType();

         abstract T buildView(ViewGroup parent);
     }
```

I don't like the idea of having to manually give an item type. That could hopefully be generated.

I'm also thinking that we could automatically make a `buildView` implementation in the generated class that just returns a new instance of the view.

One other tricky part is that the adapter only has the item type `onCreateViewHolder(ViewGroup parent, int layoutRes)` when creating the viewholder, so we would need some way to link that to the model.

Alternatively, maybe we should separate the view building from the epoxy model, and have a factory interface that builds a view for a given item type.

Do you have an example use case or ideas for what this would look like?

gumil:
For an example usecase, I've been using Anko to create my views that's why I needed to have a way to use views instead of inflating layouts.

What I came up with but not really tested it yet is to have a constructor for EpoxyViewHolder that takes a view.

`  public EpoxyViewHolder(View itemView) {
    super(itemView);
  }`

For the adapter maybe something like this

    private SparseArray<EpoxyModelWithView> views = new SparseArray<>();

    @Override
    public EpoxyViewHolder onCreateViewHolder(ViewGroup parent, int layoutRes) {
        return new EpoxyViewHolder(views.get(layoutRes));
    }
    protected void addModel(EpoxyModelWithView<?> epoxyModelWithView) {
        super.addModel(epoxyModelWithView);
        views.put(epoxyModelWithView.hashCode(), epoxyModelWithView);
    }

I can see that creating the ID manually might cause a lot of careless bugs. Maybe we could make use of hashCode()?

elihart:
using `hashCode()` for item types (maybe hashing the fully qualified name of the model) would be easiest, but I would worry about the chance of collisions. The chance would be exceedingly small, but if you got unlucky with multiple item types it just wouldn't work.

One other question about item types - do you foresee a model needing to programatically change its model type? Maybe the programmatic view it builds can differ and it needs to have its item type change? In that case we can't tie the item type to the model class.

gumil:
[...]

PR #153, opened 2017-03-12
Automatically assign view types for models with view

elihart:
Finishes support for https://github.com/airbnb/epoxy/issues/86

This allows models with programmatic views to not specify a custom view type (leaving a default of 0) and at runtime the adapter will create a view type to use with models of that class.

@niccorder This is MUCH simpler than trying to generate a value at compile time in the processor, and this seems to work nicely. Hope you didn't spend too much time on the gradle plugin approach. Personally I was too focused on the idea of solving this in the annotation processor. Thankfully my teammate realized it could be done this way instead.

I included some other clean up and feature work for the 2.0 branch too, but the applicable change is the `ViewTypeManager` class and it's usage in BaseEpoxyAdapter. EpoxyModelWithView was also updated to support this.

@ngsilverman Thanks for the idea! It is awesome.

### DIFF
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/ViewTypeManager.java
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/ViewTypeManager.java
@@ -1,0 +1,92 @@
+package com.airbnb.epoxy;
+
+import android.support.annotation.VisibleForTesting;
+
+import java.util.HashMap;
+import java.util.Map;
+
+class ViewTypeManager {
+  private static Map<Class, Integer> viewTypeMap;
+  /**
+   * The last model that had its view type looked up. This is stored so in most cases we can quickly
+   * look up what view type belongs to which model.
+   */
+  private EpoxyModel<?> lastModelForViewTypeLookup;
+
+  /**
+   * The type map is static so that models of the same class share the same views across different
+   * adapters. This is useful for view recycling when the adapter instance changes, or when there
+   * are multiple adapters. For testing purposes though it is good to be able to clear the map so we
+   * don't carry over values across tests.
+   */
+  @VisibleForTesting
+  void resetMapForTesting() {
+    if (viewTypeMap != null) {
+      viewTypeMap.clear();
+    }
+  }
+
+  int getViewType(EpoxyModel<?> model) {
+    lastModelForViewTypeLookup = model;
+    return getViewTypeInternal(model);
+  }
+
+  private static int getViewTypeInternal(EpoxyModel<?> model) {
+    int defaultViewType = model.getViewType();
+    if (defaultViewType != 0) {
+      return defaultViewType;
+    }
+
+    // If a model does not specify a view type then we generate a value to use for models of that
+    // class.
+    Class modelClass = model.getClass();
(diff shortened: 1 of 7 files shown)


epoxy-adapter/src/test/java/com/airbnb/epoxy/ViewTypeManagerTest.java
```diff
@@ -0,0 +1,113 @@
...
+
+  @Test
+  public void modelWithLayout() {
+    SimpleEpoxyAdapter adapter = new SimpleEpoxyAdapter();
+    adapter.addModel(new ModelWithViewType());
+    adapter.addModel(new ModelWithViewType2());
+
+    // The view type should be the value declared in the model
+    assertEquals(1, adapter.getItemViewType(0));
+    assertEquals(2, adapter.getItemViewType(1));
```
ngsilverman: Why test `ViewTypeManager` through the adapter instead of directly? These look more like `BaseEpoxyAdapter` tests.

elihart: Yeah, this isn't a perfect architecture in that the adapter and view type manager are tightly coupled, so it's easiest to test through the adapter. I could make the interface more generic, or mock out the adapter, to make this a true unit test of the view type manager, but I don't think that's worth the extra overhead even though that would be better form. These are more like integration tests of the adapter/viewtypemanager usage so we can be sure that it is used correctly in the adapter.

epoxy-adapter/src/main/java/com/airbnb/epoxy/ViewTypeManager.java
```diff
@@ -0,0 +1,92 @@
...
+        new IllegalStateException("Last model did not match expected view type"));
+
+    // To be extra safe in case RecyclerView implementation details change...
+    for (EpoxyModel<?> model : adapter.getCurrentModels()) {
+      if (model.getViewType() == viewType) {
+        return model;
+      }
+    }
+
+    throw new IllegalStateException("Could not find model for view type: " + viewType);
```
elihart: Thanks for the thought! I agree, and I've been trying to be careful about that. The new controller has a `onExceptionSwallowed` method that is called with any exceptions we can recover from. By default it does nothing, but the subclass override to do anything it wants.

In this case I don't think we have much choice but to crash. If we can't find the model then we can't accurately create the right view, which could lead to a crash down the line when the expected view doesn't match. 

I appreciate the feedback though, it's great to here suggestions on architecture as I keep working on this.

elihart:
@niccorder Thanks for looking, and I'm interested to hear your idea about dynamic content.

niccorder:
No worries, anytime!

The TL;DR of dynamic content would be to leverage the fact that Epoxy itself is already a renderer, and it already is sort of expressive. Ideally, the end goal would be to define a DML (interface/JSON/Whatever) that would basically handle eeverything for the user based off their interactivity graph.

I'll go into depth once I have some basic method signatures/architecture for the idea.

It may not fit in with version 2.0 though as you are understandably making this a much more "closed" system. I'll take a more in depth look later tonight

Issue #152, opened 2017-03-09
Qualify layout files in annotation processor


elihart:
To prevent collisions between layout resources of the same value from different libraries. We just need to do the same thing as https://github.com/JakeWharton/butterknife/commit/4ad9f9888c4b62476591488cade3aacda80330ba. Should be straightforward

geralt-encore:
Can you elaborate on this one, please? I got the idea more or less, but just want to be sure that I understand it 100% before starting to implement.

elihart:
Sure @geralt-encore. We want to be able to do `@EpoxyModelClass(R.layout.layout_file)`, but at compile time that annotation param value is just an int, and we need to look up the name of the resource so we can turn it back into just `R.layout.layout_file`. Butterknife does some complex thing to be able to do that which I adopted, and it lives in `ResourceProcessor` in Epoxy.

The problem is that when compiling resources from multiple modules there is a risk of collision. An R file from one module (com.module1.R) and another module (com.module2.R) processed their R files separately and might have a layout resource that has the same int value. In this case we can get a collision since we assume a one way mapping between layout resource value and the resource name. We need to change that to consider both resource value and the package of the R file so we don't get collisions.

The change to butterknife to fix this was made by a coworker of mine and it has worked well for us. I think we just need to change Epoxy's `ResourceProcessor` to copy this

elihart:
fixed with https://github.com/airbnb/epoxy/pull/155

PR #154, opened 2017-03-12
Qualify layout files in annotation processor

geralt-encore:
https://github.com/airbnb/epoxy/issues/152
Not sure how to test is thought. I've seen Butterknife's approach but didn't quite get hot to apply it here.

### DIFF
--- a/epoxy-processor/src/main/java/com/airbnb/epoxy/GeneratedModelWriter.java
+++ b/epoxy-processor/src/main/java/com/airbnb/epoxy/GeneratedModelWriter.java
@@ -441,7 +441,8 @@ class GeneratedModelWriter {
       return;
     }
 
-    AndroidResource layoutResource = resourceProcessor.getResourceForValue(layoutRes);
+    AndroidResource layoutResource = resourceProcessor.getResourceForValue(
+        resourceProcessor.elementToQualifiedId(originalClassElement, layoutRes));
     getDefaultLayoutMethod = getDefaultLayoutMethod.toBuilder()
         .addStatement("return $L", layoutResource.code)
         .build();
--- a/epoxy-processor/src/main/java/com/airbnb/epoxy/QualifiedId.java
+++ b/epoxy-processor/src/main/java/com/airbnb/epoxy/QualifiedId.java
@@ -1,9 +1,5 @@
 package com.airbnb.epoxy;
 
-/**
- * Created by ilyazorin on 12/03/2017.
- */
-
 final class QualifiedId {
 
   final String packageName;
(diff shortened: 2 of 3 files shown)


epoxy-processor/src/main/java/com/airbnb/epoxy/QualifiedId.java
```diff
@@ -0,0 +1,48 @@
+package com.airbnb.epoxy;
+
+/**
+ * Created by ilyazorin on 12/03/2017.
```
elihart: Remove the created by javadoc please. We like relying on git blame instead of code comments since those can become out of date.

geralt-encore: Sorry, totally forgot about that

elihart:
Thanks for working on this! It's definitely a bit confusing, I pulled down your branch to look at it in the IDE. I'll think about how to test it. This looks like a great start though!

epoxy-processor/src/main/java/com/airbnb/epoxy/GeneratedModelWriter.java
```diff
@@ -441,7 +441,8 @@ private void addDefaultLayoutMethodIfNeeded(TypeElement originalClassElement,
       return;
     }
 
-    AndroidResource layoutResource = resourceProcessor.getResourceForValue(layoutRes);
+    AndroidResource layoutResource = resourceProcessor.getResourceForValue(
```
elihart: since this is the only place that `getResourceForValue` and `elementToQualifiedId` are called it seems like we could combine them into one method that returns `AndroidResource` and never expose the qualified id here. possibly just changing the parameters of `getResourceForValue` to `(originalClassElement, layoutRes)`, but maybe there is a better name

elihart:
@geralt-encore I dug into this, and realized its using the wrong package name to qualify the resources. It should use the package name of the model class, not the R class.

I also spent more time trying to understand exactly what the R class parsing it doing (since I only half understood it when copying from butterknife), and realized we can simplify quite a bit. Butterknife has to handle a lot of different annotations and layout parameters in each annotation, but we have a very specialized case of just `EpoxyModelClass` and the layout resource in that. I don't imagine that changing in the future so we can probably simplify the code to make it more understandable.

Mind if I push some commits to this?

geralt-encore:
Sure, feel free to change anything. Unfortunately, I didn't have time to dig into it, so I don't have good enough understanding of what's going on.

elihart:
To be honest I didn't fully understand it either. After spending way too much time finally digging into the details I think I have a much better way to accomplish this -> https://github.com/airbnb/epoxy/pull/155

PR #155, opened 2017-03-13
Refactor resource processing

elihart:
I started playing around with https://github.com/airbnb/epoxy/issues/152 to clean up the resource processor. As I started to understand the intricacies more I realized it was very inefficient and spent _way_ too long trying to figure out how to do it better. I think the result is much easier to understand and should be a lot faster. This maps the fully qualified name of a model class to the layout information used in that model's class annotation.

This eliminates the need to parse the R class file by getting the resource name directly from the annotation parameter value. I'm not sure why butterknife didn't simply do it this way. We only have to deal with layout values, which makes this simpler, but I believe they could still do it this way. I also think that the namespacing that butterknife does via package name isn't perfect since two classes in the same package could use different R classes.

@geralt-encore It was simpler to open up a new PR instead of pushing to your fork. Sorry! Thanks for your help getting this going. Let me know if this seems readable and understandable to you, and if you spot areas to improve naming/documentation.

@BenSchwab Since you're familiar with how butterknife does this maybe you can take a look for me? I diverged quite a bit from how butterknife approaches this, but it seems better. I hope I am not missing a case that this doesn't support though. Am I right that since butterknife namespaces with package name you might still get a collision of different R classes used in classes of the same package?

Fixes https://github.com/airbnb/epoxy/issues/152


geralt-encore:
@elihart this one is far easier to read and understand 👍

geralt-encore:
I have a question if you don't mind. Why do we need to work here with AST directly instead of using regular annotation processing primitives as usual?

BenSchwab:
This looks very clean to me, good work!

I'll summarize our discussion for posterity: I view this mostly as a way to reduce red herrings. I.E. when there is a resource collision, something _is_ wrong with your app, (most likely referencing another module's R file, which can have different R values).

Let's say module `bar` depends on  module `foo`
```
//foo.r
R.id.rogue = 5;

//bar.r
R.id.victim = 5;
R.id.rogue = 6;
```

So in this case `@EpoxyModelClass(foo.R.id.rogue)` would still lead to bad things. (The id we should be using is `R.id.rogue`.)  But, the annotation processor won't crush `R.id.victim`. So `@EpoxyModelClass(R.id.victim)` will now work. This should be helpful to anyone who is debugging this issue, as `@EpoxyModelClass(R.id.victim)` did everything right.

You are right on the butterknife front --  mysterious collisions may still happen if the colliding resources are referenced in the same package.

The solution you have may not work for butterknife because, if I am reading this right, you are mapping `annotatedElementName -> ResourceIDModel`. Butterknife allows List bindings, so there is not a 1:1 mapping between annotatedElement and ResourceIdModel (one list field maps to many r values). But this approach is very clean, and maybe able to be adapted.

Also, thanks for the comments in the code!

elihart:
cc @ngsilverman you could take this resource processor for your usage

elihart:
@geralt-encore Thanks for looking. To be honest I borrowed this from butterknife and haven't wanted to spend more time exploring other ways to do it since it worked initially without much adaptation from butterknife. It does seem we should be able to do this with normal annotation processing, but since it's working already and I've already spent too much time on this I'm not ready to look into it. If you are interested in figuring out how to do it that'd be cool.

@BenSchwab Thanks for the summary! Appreciate you writing that up for future reference. You're right that this isn't immediately adaptable by butterknife, but it doesn't seem that bad to change it to `Map<annotatedElementName, Map<resourceValue, ResourceData>>`. basically just add another chain to the lookup so the annotated element can have multiple resource values. Probably not worth doing unless we were worried about the processing performance of parsing the R class tree so many times

PR #159, opened 2017-03-16
Add interceptor to EpoxyController to allow for adjusting models

elihart:
This interceptor will run after `buildModels` and before the models are diffed and set on the adapter. The use case is adjustments to the models that are best done when you have all the models in aggregate to work with, such as toggling divider logic or moving a model for an experiment.

@ngsilverman @gpeal

### DIFF
--- a/epoxy-adapter/src/test/java/com/airbnb/epoxy/EpoxyControllerTest.java
+++ b/epoxy-adapter/src/test/java/com/airbnb/epoxy/EpoxyControllerTest.java
@@ -1,0 +1,195 @@
+package com.airbnb.epoxy;
+
+import android.support.v7.widget.RecyclerView.AdapterDataObserver;
+
+import com.airbnb.epoxy.EpoxyController.Interceptor;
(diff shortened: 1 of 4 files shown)


epoxy-adapter/src/test/java/com/airbnb/epoxy/EpoxyControllerTest.java
```diff
@@ -0,0 +1,195 @@
...
+      @Override
+      public void intercept(List<EpoxyModel<?>> models) {
+        savedModels = models;
+      }
+    });
+
+    controller.requestModelBuild();
+
+    savedModels.add(new TestModel());
+  }
```
ngsilverman: Maybe add a test with more than one interceptor, to verify that they're all executed, and in the right order.

elihart: Great idea, thanks!

PR #161, opened 2017-03-20
Add model usage validation in generated code

elihart:
This adds validations to the generated models that enforce immutability after the model is added to the controller. This will help enforce the pattern that models must be recreated when changes are needed, instead of modifying existing models. I tried to make the error messages as helpful and descriptive as possible to teach users how to use the EpoxyController correctly.

@ngsilverman @geralt-encore


elihart:
I'm gonna merge this and keep working, but if you have any comments please leave them and I'll address them in another PR

PR #166, opened 2017-03-22
Change how model validation param is set

elihart:
This changes the options in the package annotation to instead be specified as annotation processor options in the build.gradle. This makes it easier to apply options, and gives us more flexibility in how we apply them.

It also optimizes the hash code validation since that was really slowing down dev builds.

lastly this bumps to rc2


elihart:
@felipecsl Thanks for the help!

PR #167, opened 2017-03-23
Fix hiding models in EpoxyAdapter

elihart:
I had broken hiding models in the original EpoxyAdapter when I introduced EpoxyModelWithView. This fixes it and adds some tests


elihart:
Also bumps to RC3

PR #179, opened 2017-04-04
Option to delay model build by arbitrary amount

elihart:
Using this to delay a model update may be helpful in cases where user input is causing many
rapid changes in the models, such as typing. In that case, the view is already updated on
screen and constantly rebuilding models is potentially slow and unnecessary. The downside to
delaying the model build too long is that models will not be in sync with the data or view, and
scrolling the view offscreen and back onscreen will cause the model to bind old data.

### DIFF
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyController.java
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyController.java
@@ -57,7 +57,7 @@ public abstract class EpoxyController {
    */
   public void requestModelBuild() {
     if (isBuildingModels()) {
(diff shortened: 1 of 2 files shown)


thenickreynolds:
Do we perceive the update lag today or is this defensive?

elihart:
Mostly defensive. I haven't worked on a screen that has needed this, but I know we have many places that might and I haven't profiled them. On the more complicated controllers I have profiled (Explore), building models can take 100ms if you have loaded a lot of data, which is not something we want to do with every character typed on a keyboard. For smaller screens it shouldn't matter so much, but I think it is a good option to have when needed, and I don't see any cons.

Issue #175, opened 2017-04-03
Saved view state does not work for views nested inside EpoxyModelGroup


elihart:
If there are multiple of the same model group in the recyclerview, since the nested views don't have a unique id across each model group.

Issue #163, opened 2017-03-22
Update sample on 2.0 branch to show new features


elihart:
To get ready for the 2.0 release we need to update the sample project to show how to use the new features, as well as take the opportunity for a few other improvements.

We should show:
1. The new EpoxyController. I've already updated the sample to use this instead of the adapter.
2. Building a view programmatically. I've created a Carousel view and CarouselModel that does this.
3. Showing how to use a nested carousel. I've already added the Carousel view/model and have a basic usage in the sample. This should be improved though.
4. Showing shared recycler view pools between all the carousels and main recyclerview.
5. Demonstrate saving state. Save the state of all the data and restore from the activity.
6. Demonstrate the new onModelClick listener. Possibly add a click listener to the color square that toggles its color.
7. Demonstrate the onBind method. Not sure what the implementation would do.
8. Demonstrate the EpoxyModelGroup to show how to group models.

To tie all this together I'm thinking that the current buttons (add/clear/shuffle/change) should add a horizontal carousel instead of a single color block (and maybe get rid of the `change` button), and then we can clear or shuffle the carousels.

The carousel group can be a EpoxyModelGroup with a horizontal linear layout. It can have buttons on the left side stacked vertically (add/clear/shuffle/change) and a Carousel taking up the rest of the horizontal space. The buttons would add/remove a color block into the carousel. 

We currently store a list of ColorData to maintain state. This would be changed to List<List<ColorData>> to represent the nested carousels. The main activity would save and restore this state.

Besides that it would be nice to polish the UI with nicer colors, layout, and design. I'm not great at coming up with nice designs so I'm open to anybody willing to improve it if they have good ideas.

elihart:
@geralt-encore are you still interested in working on this? I'd really appreciate it!

geralt-encore:
Of course! Not sure how much I will able to do, but I'll definitely will try to help you as much as I can with this.

elihart:
@geralt-encore thanks! Let me know if you have questions and what you think you might be able to do. Whatever you can't get to I'll probably work on in a week or so

elihart:
@geralt-encore Have you had any time to work on this? I'm finishing up documentation and other changes and will start on this soon if you haven't

geralt-encore:
Sorry, it was a tough week 😔 
I might have some time on the weekend but can't really promise.

elihart:
[...]

PR #181, opened 2017-04-05
Fix model group saved state and update sample

elihart:
This fixes https://github.com/airbnb/epoxy/issues/175 and adds a few features to the sample for https://github.com/airbnb/epoxy/issues/163.

The sample now has icon buttons for the carousels instead of text, and I change the color block click animation to be a lottie heart animation instead of just a color change. This was we can show how to use the new payload feature.

@geralt-encore @ngsilverman 

image


epoxy-sample/src/main/java/com/airbnb/epoxy/sample/models/ColorModel.java
```diff
@@ -17,14 +24,83 @@
...
+      if (lottieView.isAnimating()) {
+        // Reverse it just for fun
+        lottieView.resumeReverseAnimation();
+      } else {
+        cancelAnimation(lottieView);
+      }
+      return;
+    }
+
+    lottieView.addAnimatorListener(new AnimatorListener() {
```
geralt-encore: `SimpleAnimatorListener` can be used instead of `AnimatorListener` to avoid redundant override of `onAnimationRepeat`.

elihart: Good point, thanks!

PR #188, opened 2017-04-11
Databinding model

elihart:
This is mostly working, I just need to make it so if there is a change payload we only update variables if they changed.

The generated code looks like 
```java
@Override
  protected void setDataBindingVariables(ViewDataBinding binding) {
    if (!binding.setVariable(BR.textRes, textRes)) {
      throw new IllegalStateException("The attribute textRes was defined in your data binding model (com.airbnb.epoxy.sample.models.ButtonModel) but a data variable of that name was not found in the layout.");
    }
    if (!binding.setVariable(BR.clickListener, clickListener)) {
      throw new IllegalStateException("The attribute clickListener was defined in your data binding model (com.airbnb.epoxy.sample.models.ButtonModel) but a data variable of that name was not found in the layout.");
    }
  }

  @Override
  protected void setDataBindingVariables(ViewDataBinding binding, EpoxyModel previousModel) {
    if (!(previousModel instanceof ButtonModel_)) {
      setDataBindingVariables(binding);
      return;
    }
    ButtonModel_ otherModel = (ButtonModel_) previousModel;
    // TODO: compare attributes and update variable if necessary
  }
```


elihart:
This is done, if anyone is interested in taking a look. Will merge later this weekend

elihart:
fyi @ngsilverman

Issue #185, opened 2017-04-08
Better support for item change animations


elihart:
From an item animator we can delegate to each model to allow them to handle their change animations

elihart:
@tschuchortdev let me know if you're still interested in putting up a PR for this

tschuchortdev:
@elihart Yes I'm still working on it. I'd say the feature is about 80% finished. Initially I thought it would be already finished completely by now, but there were some complications. The A-key on my keyboard broke so I couldn't code for a few days. 

Currently I'm trying to implement the interrupted animation logic, which is relatively complex and thus porting it directly from the existing Kotlin code is really ugly due to Java's verbosity (it was borderline unreadable in Kotlin already).   

Is there decent documentation somewhere for 2.0 already? I really have no idea how this new controller stuff works, which makes it hard to test for me.

elihart:
Oh boy, that is quite a complication. Glad your keyboard is fixed now. Too bad we can't fix Java so easily :(

I'm currently working on the documentation at https://github.com/airbnb/epoxy/wiki/Epoxy-Controller. I'm hoping to have that done today. Besides that the new sample shows example usage https://github.com/airbnb/epoxy/tree/2.0/epoxy-sample/src/main/java/com/airbnb/epoxy/sample

If you have any specific questions on it I'm happy to answer them here too. Thanks for continuing the work!

sergeyfitis:
Any progress? Seems it would be useful for me.

tschuchortdev:
@sergeyfitis: almost done. you can take a look at the PR, maybe you have some ideas how to improve it.

elihart:
I don't think it makes sense for epoxy to try to tackle this anymore

PR #189, opened 2017-04-13
[WIP] Custom Change Animations

tschuchortdev:
WIP feature to delegate change animations from the `ItemAnimator` to the `EpoxyModel` to make models truly composable. (See #185) 

Hopefully I can push a working implementation on the weekend. This still has a ton of compiler errors from playing around with various ideas, but the general implementation is sound. It's mostly about finding the right interface now, like how to integrate change animations with regular binding.  

@elihart

### DIFF
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/BaseEpoxyAdapter.java
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/BaseEpoxyAdapter.java
@@ -145,8 +145,8 @@ abstract class BaseEpoxyAdapter extends RecyclerView.Adapter<EpoxyViewHolder> {
 
     // set our item animator decorator so we can intercept calls to the item animator
     // and the user doesn't have to worry about setting the decorator himself
-    if (!(recycler.getItemAnimator() instanceof ChangeAnimationDelegator)) {
-      recycler.setItemAnimator(new ChangeAnimationDelegator(recycler.getItemAnimator()));
+    if (!(recycler.getItemAnimator() instanceof EpoxyAnimationDelegator)) {
+      recycler.setItemAnimator(new EpoxyAnimationDelegator(recycler.getItemAnimator()));
     }
   }
 
@@ -159,8 +159,8 @@ abstract class BaseEpoxyAdapter extends RecyclerView.Adapter<EpoxyViewHolder> {
     // the user decides to attach a non-epoxy adapter afterwards
     ItemAnimator itemAnimator = recycler.getItemAnimator();
 
-    if (itemAnimator instanceof ChangeAnimationDelegator) {
-      recycler.setItemAnimator(((ChangeAnimationDelegator) itemAnimator).getDecoratedAnimator());
+    if (itemAnimator instanceof EpoxyAnimationDelegator) {
+      recycler.setItemAnimator(((EpoxyAnimationDelegator) itemAnimator).getDecoratedAnimator());
     }
   }
 
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/BaseEpoxyAdapter.java
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/BaseEpoxyAdapter.java
@@ -5,6 +5,7 @@ import android.support.annotation.CallSuper;
 import android.support.annotation.Nullable;
 import android.support.v7.widget.GridLayoutManager.SpanSizeLookup;
 import android.support.v7.widget.RecyclerView;
+import android.support.v7.widget.RecyclerView.ItemAnimator;
 import android.view.View;
 import android.view.ViewGroup;
 
@@ -13,17 +14,13 @@ import java.util.List;
 
 abstract class BaseEpoxyAdapter extends RecyclerView.Adapter<EpoxyViewHolder> {
   private static final String SAVED_STATE_ARG_VIEW_HOLDERS = "saved_state_view_holders";
-
-  private int spanCount = 1;
-
   private final ViewTypeManager viewTypeManager = new ViewTypeManager();
   /**
    * Keeps track of view holders that are currently bound so we can save their state in {@link
    * #onSaveInstanceState(Bundle)}.
    */
   private final BoundViewHolders boundViewHolders = new BoundViewHolders();
-  private ViewHolderState viewHolderState = new ViewHolderState();
-
+  private int spanCount = 1;
   private final SpanSizeLookup spanSizeLookup = new SpanSizeLookup() {
 
     @Override
@@ -43,6 +40,7 @@ abstract class BaseEpoxyAdapter extends RecyclerView.Adapter<EpoxyViewHolder> {
       }
     }
   };
+  private ViewHolderState viewHolderState = new ViewHolderState();
 
   public BaseEpoxyAdapter() {
     // Defaults to stable ids since view models generate unique ids. Set this to false in the
@@ -138,6 +136,32 @@ abstract class BaseEpoxyAdapter extends RecyclerView.Adapter<EpoxyViewHolder> {
    */
   protected void onModelBound(EpoxyViewHolder holder, EpoxyModel<?> model, int position) {
 
+  }
+
+  @CallSuper
+  @Override
+  public void onAttachedToRecyclerView(RecyclerView recycler) {
+    super.onAttachedToRecyclerView(recycler);
+
+    // set our item animator decorator so we can intercept calls to the item animator
+    // and the user doesn't have to worry about setting the decorator himself
+    if (!(recycler.getItemAnimator() instanceof ChangeAnimationDelegator)) {
+      recycler.setItemAnimator(new ChangeAnimationDelegator(recycler.getItemAnimator()));
+    }
+  }
+
+  @CallSuper
+  @Override
+  public void onDetachedFromRecyclerView(RecyclerView recycler) {
+    super.onDetachedFromRecyclerView(recycler);
+
+    // remove our item animator decorator when the adapter is detached so there are no problems if
+    // the user decides to attach a non-epoxy adapter afterwards
+    ItemAnimator itemAnimator = recycler.getItemAnimator();
+
+    if (itemAnimator instanceof ChangeAnimationDelegator) {
+      recycler.setItemAnimator(((ChangeAnimationDelegator) itemAnimator).getDecoratedAnimator());
+    }
   }
 
   /**
@@ -266,6 +290,10 @@ abstract class BaseEpoxyAdapter extends RecyclerView.Adapter<EpoxyViewHolder> {
     return spanSizeLookup;
   }
 
+  public int getSpanCount() {
+    return spanCount;
+  }
+
   /**
    * If you are using a grid layout manager you must call this to set the span count of the grid.
    * This span count will be passed on to the models so models can choose what span count to be.
@@ -275,10 +303,6 @@ abstract class BaseEpoxyAdapter extends RecyclerView.Adapter<EpoxyViewHolder> {
    */
   public void setSpanCount(int spanCount) {
     this.spanCount = spanCount;
-  }
-
-  public int getSpanCount() {
-    return spanCount;
   }
 
   public boolean isMultiSpan() {
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/BiHashMap.java
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/BiHashMap.java
@@ -1,0 +1,135 @@
+package com.airbnb.epoxy;
+
+import android.support.annotation.NonNull;
+
+import java.util.AbstractCollection;
+import java.util.AbstractMap;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Set;
+import java.util.function.BiFunction;
+
+/**
+ * simple bimap implementation using two maps
+ *
+ * @param <K> key type
+ * @param <V> value type
+ */
+class BiHashMap<K, V> extends AbstractMap<K, V> {
+  private Map<K, V> mMap;
+  private Map<V, K> mInverseMap;
+
+  public BiHashMap() {
+    mMap = new HashMap<>();
+    mInverseMap = new HashMap<>();
+  }
+
+  public BiHashMap(int initialCapacity) {
+    mMap = new HashMap<>(initialCapacity);
+    mInverseMap = new HashMap<>(initialCapacity);
+  }
+
+  private BiHashMap(Map<K, V> map, Map<V, K> inverseMap) {
+    mMap = map;
+    mInverseMap = inverseMap;
+  }
+
+  public BiHashMap<V, K> inverse() {
+    return new BiHashMap<>(mInverseMap, mMap);
+  }
+
+  public void clear() {
+    mMap.clear();
+    mInverseMap.clear();
+  }
+
+  public boolean containsKey(Object key) {
+    return mMap.containsKey(key);
+  }
+
+  public boolean containsValue(Object value) {
+    return mMap.containsValue(value);
+  }
+
+  public int size() {
+    return mMap.size();
+  }
+
+  public boolean isEmpty() {
+    return mMap.isEmpty();
+  }
+
+  public synchronized V put(K key, V value) {
+    mInverseMap.put(value, key);
+    return mMap.put(key, value);
+  }
+
+  public V get(Object key) {
+    return mMap.get(key);
+  }
+
+  public synchronized V remove(Object key) {
+    V value = mMap.remove(key);
+    mInverseMap.remove(value);
+    return value;
+  }
+
+  public synchronized void putAll(@NonNull Map<? extends K, ? extends V> m) {
+    super.putAll(m);
+  }
+
+  @Override
+  public synchronized V merge(K key, V value,
+      BiFunction<? super V, ? super V, ? extends V> remappingFunction) {
+    return super.merge(key, value, remappingFunction);
+  }
+
+  @NonNull
+  @Override
+  public Set<K> keySet() {
+    return mMap.keySet();
+  }
+
+  @NonNull
+  @Override
+  public Collection<V> values() {
+    return new AbstractCollection<V>() {
+      @Override
+      public int size() {
+        return mMap.size();
+      }
+
+      @NonNull
+      @Override
+      public Iterator<V> iterator() {
+        return new IteratorDecorator<V>(mMap.values().iterator()) {
+          @Override
+          public void remove() {
+            super.remove();
+            mInverseMap.remove(current);
+          }
+        };
+      }
+
+      @Override
+      public boolean add(V v) {
+        throw new UnsupportedOperationException(
+            "adding to the value collection of a map is not supported");
+      }
+
+      @Override
+      public boolean addAll(@NonNull Collection<? extends V> c) {
+        throw new UnsupportedOperationException(
+            "adding to the value collection of a map is not supported");
+      }
+    };
+  }
+
+  @NonNull
+  @Override
+  public Set<Entry<K, V>> entrySet() {
+    return mMap.entrySet();
+  }
+}
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/ChangeAnimationDelegator.java
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/ChangeAnimationDelegator.java
@@ -1,0 +1,168 @@
+package com.airbnb.epoxy;
+
+import android.animation.AnimatorListenerAdapter;
+import android.animation.EpoxyItemAnimation.AnimatorListener;
+import android.support.annotation.NonNull;
+import android.support.v7.widget.RecyclerView.ItemAnimator;
+import android.support.v7.widget.RecyclerView.ViewHolder;
+
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * ItemAnimator that wraps any other ItemAnimator and delegates change animations to
+ * {@link EpoxyModel}
+ */
+public final class ChangeAnimationDelegator extends ItemAnimatorDecorator
+    implements AnimatorListener {
+  /* use a BiMap so we can use the same listener for every animator and quickly find the holder for
+     an animation. We do this to be able to remove only OUR listener from an animator and not those
+     that the user might have set */
+  @NonNull private final BiHashMap<ViewHolder, EpoxyItemAnimation> mChangeAnimations =
+      new BiHashMap<>();
+
+  public ChangeAnimationDelegator(@NonNull ItemAnimator decoratedAnimator) {
+    super(decoratedAnimator);
+  }
+
+  public long getChangeDuration() {
+    long maxDuration = super.getChangeDuration();
+
+    for (EpoxyItemAnimation anim : mChangeAnimations.values()) {
+      maxDuration = Math.max(maxDuration, anim.getDuration());
+    }
+
+    return maxDuration;
+  }
+
+  public void onAnimationEnd(EpoxyItemAnimation anim) {
+    dispatchAnimationFinished(mChangeAnimations.inverse().get(anim));
+    mChangeAnimations.inverse().remove(anim);
+
+    if (mChangeAnimations.isEmpty()) {
+      dispatchChangeAnimationsFinished();
+    }
+  }
+
+  @Override
+  public void onAnimationCancel(EpoxyItemAnimation anim) {
+    mChangeAnimations.inverse().remove(anim);
+  }
+
+  public void onAnimationStart(EpoxyItemAnimation anim) {
+    dispatchAnimationStarted(mChangeAnimations.inverse().get(anim));
+  }
+
+  @SuppressWarnings("unchecked")
+  public boolean animateChange(
+      @NonNull final ViewHolder oldHolder, @NonNull final ViewHolder newHolder,
+      @NonNull ItemHolderInfo preLayoutInfo, @NonNull ItemHolderInfo postLayoutInfo) {
+
+    if (oldHolder == newHolder && oldHolder instanceof EpoxyViewHolder
+        && canHolderAnimateChanges((EpoxyViewHolder) oldHolder)) {
+
+      final EpoxyViewHolder holder = ((EpoxyViewHolder) oldHolder);
+      final EpoxyModel newModel = holder.getModel();
+      final Object viewToBind = holder.objectToBind();
+      final EpoxyModel oldModel = DiffPayload.getModelFromPayload();
+
+      final EpoxyItemAnimation oldAnim = mChangeAnimations.get(oldHolder);
+
+      if (oldAnim == null || !oldAnim.isRunning()) {
+        EpoxyItemAnimation newAnim = newModel.bindChangesAnimated(viewToBind, oldModel);
+        newAnim.addListener(this);
+        mChangeAnimations.put(holder, newAnim);
+      } else {
+        EpoxyItemAnimation continuedAnim =
+            newModel.bindChangesAnimated(viewToBind, oldModel, oldAnim);
+
+        if (continuedAnim != null) {
+          //cancel old anim if user decided to create a new animator instead of reusing the old one
+          if (continuedAnim != oldAnim) {
+            oldAnim.cancel();
+            continuedAnim.start();
+          }
+
+          continuedAnim.addListener(this);
+          mChangeAnimations.put(holder, continuedAnim);
+        } else {
+          final EpoxyItemAnimation newAnim = newModel.bindChangesAnimated(viewToBind, oldModel);
+          newAnim.addListener(this);
+
+          oldAnim.removeListener(this);
+
+          oldAnim.addListener(new AnimatorListenerAdapter() {
+            @Override
+            public void onAnimationEnd(EpoxyItemAnimation animation) {
+              mChangeAnimations.put(holder, newAnim);
+              newAnim.start();
+            }
+          });
+        }
+      }
+
+      return true;
+    } else {
+      return super.animateChange(oldHolder, newHolder, preLayoutInfo, postLayoutInfo);
+    }
+  }
+
+  public void runPendingAnimations() {
+    super.runPendingAnimations();
+
+    // start each animator in the map
+    for (EpoxyItemAnimation anim : mChangeAnimations.values()) {
+      if (!anim.isRunning()) {
+        // play change animations after removals, in accordance with DefaultItemAnimator
+        if (disappearAnimationsScheduled()) {
+          anim.setStartDelay(getRemoveDuration());
+        }
+
+        anim.start();
+        // don't dispatch animation start here because it's already done in the listener
+      }
+    }
+  }
+
+  public void endAnimation(@NonNull ViewHolder holder) {
+    super.endAnimation(holder);
+
+    EpoxyItemAnimation anim = mChangeAnimations.get(holder);
+    if (anim != null) {
+      anim.end();
+      // don't dispatch animation end here because it's already done in the listener
+    }
+  }
+
+  public void endAnimations() {
+    super.endAnimations();
+
+    for (EpoxyItemAnimation anim : mChangeAnimations.values()) {
+      anim.end();
+      // don't dispatch animation end here because it's already done in the listener
+    }
+  }
+
+  public boolean isRunning() {
+    boolean changesRunning = false;
+
+    for (EpoxyItemAnimation anim : mChangeAnimations.values()) {
+      changesRunning = changesRunning || anim.isRunning();
+    }
+
+    return changesRunning || super.isRunning();
+  }
+
+  public boolean canReuseUpdatedViewHolder(@NonNull ViewHolder holder,
+      @NonNull List<Object> payloads) {
+    // pass empty list to super because it contains only the previous model and DefaultItemAnimator
+    // would decide not to cross fade
+    return (holder instanceof EpoxyViewHolder && !payloads.isEmpty())
+        || super.canReuseUpdatedViewHolder(holder, Collections.emptyList());
+  }
+
+  @SuppressWarnings("unchecked")
+  private boolean canHolderAnimateChanges(@NonNull EpoxyViewHolder holder) {
+    return ((EpoxyModel) holder.getModel()).canReuseViewHolder();
+  }
+}
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/ChangeAnimationDelegator.java
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/ChangeAnimationDelegator.java
@@ -1,30 +1,31 @@
 package com.airbnb.epoxy;
 
 import android.animation.AnimatorListenerAdapter;
-import android.animation.EpoxyItemAnimation.AnimatorListener;
 import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
 import android.support.v7.widget.RecyclerView.ItemAnimator;
+import android.support.v7.widget.RecyclerView.State;
 import android.support.v7.widget.RecyclerView.ViewHolder;
 
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 /**
  * ItemAnimator that wraps any other ItemAnimator and delegates change animations to
  * {@link EpoxyModel}
  */
 public final class ChangeAnimationDelegator extends ItemAnimatorDecorator
-    implements AnimatorListener {
-  /* use a BiMap so we can use the same listener for every animator and quickly find the holder for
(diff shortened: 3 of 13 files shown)


elihart:
Awesome! This is a great start to get the discussion going, no rush for the full implementation. I'll try to go through this soon. Thanks for putting it up :)

epoxy-adapter/src/main/java/com/airbnb/epoxy/BaseEpoxyAdapter.java
```diff
@@ -140,6 +138,32 @@ protected void onModelBound(EpoxyViewHolder holder, EpoxyModel<?> model, int pos
...
   }
 
+  @CallSuper
+  @Override
+  public void onAttachedToRecyclerView(RecyclerView recycler) {
+    super.onAttachedToRecyclerView(recycler);
+
+    // set our item animator decorator so we can intercept calls to the item animator
+    // and the user doesn't have to worry about setting the decorator himself
+    if (!(recycler.getItemAnimator() instanceof ChangeAnimationDelegator)) {
```
elihart: Should we only set this if the animator is the default item animator? Otherwise if they want to use a custom animator _without_ our wrapper we would just replace their animator. If they want to use a custom animator with our wrapper they can wrap it themselves, which I don't think is unreasonable.

elihart: Related thought - if they add their own animator after the adapter is attached it would not be wrappe, but adding it before would wrap it. This seems like confusing behavior, so it might be clearer if we never automatically wrap their custom animator.

epoxy-adapter/src/main/java/com/airbnb/epoxy/BiHashMap.java
```diff
@@ -0,0 +1,135 @@
...
+import java.util.function.BiFunction;
+
+/**
+ * simple bimap implementation using two maps
+ *
+ * @param <K> key type
+ * @param <V> value type
+ */
+class BiHashMap<K, V> extends AbstractMap<K, V> {
+  private Map<K, V> mMap;
```
elihart: nit: we don't use `m` prefixes in this repo

epoxy-adapter/src/main/java/com/airbnb/epoxy/BiHashMap.java
```diff
@@ -0,0 +1,135 @@
...
+
+  public int size() {
+    return mMap.size();
+  }
+
+  public boolean isEmpty() {
+    return mMap.isEmpty();
+  }
+
+  public synchronized V put(K key, V value) {
```
elihart: does this need to be thread safe?

epoxy-adapter/src/main/java/com/airbnb/epoxy/ItemAnimatorDecorator.java
```diff
@@ -0,0 +1,256 @@
...
+        //remove holder from the lists of currently animating holders
+        mAppearAnims.remove(holder);
+        mChangeAnims.remove(holder);
+        mDisappearAnims.remove(holder);
+        mPersistAnims.remove(holder);
+      }
+    });
+  }
+
+  private static void setAnimationFinishedListener(ItemAnimator itemAnimator,
```
elihart: This is only needed if we are wrapping somebody's custom item animator that doesn't extend SimpleItemAnimator, correct? Could we use a different animator if the wrapped animator does extend SimpleItemAnimator so we can avoid reflection in that common case?

That way if you're just using the built in change animation delegation to Epoxy models you won't have the reflection slow down

epoxy-sample/src/main/java/com/airbnb/epoxy/sample/models/ColorModel.java
```diff
@@ -86,6 +94,28 @@ private void cancelAnimation(LottieAnimationView lottieView) {
   }
 
   @Override
+  public boolean canAnimateChanges(ColorHolder view) {
+    return false;
+  }
+
+  @Override
+  public Animator animateChanges(ColorHolder view) {
+    int currentColor = ((ColorDrawable) view.getBackground()).getColor();
```
elihart: side note: it would be cool to use the old animator to gracefully update which color is being animated too, since if you toggle the "change" button quickly you can cycle through many colors.

Would also be cool to used the new Spring animation in the support library

epoxy-adapter/src/main/java/com/airbnb/epoxy/ChangeAnimationDelegator.java
```diff
@@ -0,0 +1,168 @@
...
+  public void onAnimationStart(EpoxyItemAnimation anim) {
+    dispatchAnimationStarted(mChangeAnimations.inverse().get(anim));
+  }
+
+  @SuppressWarnings("unchecked")
+  public boolean animateChange(
+      @NonNull final ViewHolder oldHolder, @NonNull final ViewHolder newHolder,
+      @NonNull ItemHolderInfo preLayoutInfo, @NonNull ItemHolderInfo postLayoutInfo) {
+
+    if (oldHolder == newHolder && oldHolder instanceof EpoxyViewHolder
```
elihart: Don't think we need the `instanceof EpoxyViewHolder` checks, we can just cast it. This isn't made to be used without Epoxy

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModel.java
```diff
@@ -137,7 +183,54 @@ public void bind(T view, EpoxyModel<?> previouslyBoundModel) {
...
+  }
+
+  /**
+   * Called when the model has changed but still represents the same item. Return true if you want
+   * to animate these changes manually in the viewholder instead of simple cross fading or whatever
+   * animation the {@link android.support.v7.widget.RecyclerView.ItemAnimator) does by default.
+   *
+   * @return true if you wish to animate changes in the model
+   */
+  public boolean canReuseViewHolder(@NonNull EpoxyModel<?> previouslyBoundModel) {
```
elihart: are you still planning to rename `canReuseViewHolder` to `canAnimateChanges`?

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModel.java
```diff
@@ -137,7 +183,54 @@ public void bind(T view, EpoxyModel<?> previouslyBoundModel) {
...
+   * will be run in sequence.
+   *
+   * @param interruptedAnimation the interrupted running change animation you returned in the
+   *                             previous call to {@link #animateChanges(), or null if no animation
+   *                             was running already. Guaranteed to be of the same type, so casting
+   *                             is safe.
+   * @return non-null animator to be started by the ItemAnimator when it sees fit
+   */
+  @Nullable
+  public EpoxyItemAnimation bindChangesAnimated(@NonNull T view,
```
elihart: it looks like the item animator doesn't call this if `interruptedAnimation` is null. Can we make `interruptedAnimation` not nullable?

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModel.java
```diff
@@ -137,7 +183,54 @@ public void bind(T view, EpoxyModel<?> previouslyBoundModel) {
...
+      @NonNull EpoxyModel<?> previouslyBoundModel, @Nullable Animator interruptedAnimation) {
```
elihart: the `previouslyBoundModel` is only given if the model is used with the `EpoxyController`, so any usage in the old `EpoxyAdapter` wouldn't support these model animations. I'm ok with that, since I'd like to phase that adapter out, but just FYI. 

If we want to also support the adapter we would have to pass the normal payloads list instead of the previous model. If we don't want to support the adapter we should mention that in the javadoc of these methods.

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModel.java
```diff
@@ -123,8 +170,7 @@ public void bind(T view, List<Object> payloads) {
    *                             process, and follows the same general conditions for all
    *                             recyclerview change payloads.
    */
-  public void bind(T view, EpoxyModel<?> previouslyBoundModel) {
-    bind(view);
+  public void bindChanges(@NonNull T view, @NonNull EpoxyModel<?> previouslyBoundModel) {
```
elihart: just to be clear, you're proposing a breaking change of renaming `bind` to `bindChanges`? if we do do that it would probably be good to do that one change it its own PR for clarity and since it will require updating some tests

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyViewHolder.java
```diff
@@ -34,7 +34,7 @@ public void bind(@SuppressWarnings("rawtypes") EpoxyModel model,
 
     if (previouslyBoundModel != null) {
       // noinspection unchecked
-      model.bind(objectToBind(), previouslyBoundModel);
+      model.bindChanges(objectToBind(), previouslyBoundModel);
```
elihart: I think there is a problem with calling `bindChanges` here - it will always be called, whether or not the model is reusing the view holder. If they are _not_ reusing the viewholder then they need to do a full bind, not just bind changes.

We could check if the model is reusing a view holder and do `bindChanges` or just `bind` depending on that.

epoxy-adapter/src/main/java/com/airbnb/epoxy/ChangeAnimationDelegator.java
```diff
@@ -0,0 +1,168 @@
...
+import android.support.v7.widget.RecyclerView.ViewHolder;
+
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * ItemAnimator that wraps any other ItemAnimator and delegates change animations to
+ * {@link EpoxyModel}
+ */
+public final class ChangeAnimationDelegator extends ItemAnimatorDecorator
```
tschuchortdev: I'm not a fan of that convention tbh. When I'm reading code and encounter a library I don't know, I just think "what the hell is an Epoxy?". It doesn't really make the code any more understandable and on the off chance that I do need to know where a class comes from, I can just middle click to go to definition.

We can add the Epoxy prefix to follow the convention (which is most important after all) but maybe we should come up with a different name then, since it's pretty long already.

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModel.java
elihart: I'm not yet comfortable using `canAnimateChanges` as a proxy for whether to reuse the viewholder.

A common case I have is where a view has a nested image, and that image changes. In this case everything else is the same and we just want to load the new image. We really don't want to create a new viewholder, because the view is expensive to create and we don't need the cross fade animation. Instead we use Glide to manage our image loading and it automatically uses our placeholder, and crossfades the new image as it is loaded. however, it should not need to implement `bindChangesAnimated` and return an animation.

That is a case we need to support. To do that, we need

1. a `bind` method that passes the previous model
2. A way to set `canReuseViewHolder` to true
3. No animation returned from `bindChangesAnimated`

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModel.java
```diff
@@ -137,7 +183,54 @@ public void bind(T view, EpoxyModel<?> previouslyBoundModel) {
...
+   * finished, but do not start the animation yourself. Basically like {@link #bind(T)} but with
+   * animations.
+   * Duration will not be changed by the ItemAnimator or any calls to
+   * {@link android.support.v7.widget.RecyclerView.ItemAnimator#setChangeDuration(long)}, but start
+   * delay may be overwritten.
+   * Note that any listeners you set may be removed or the animation canceled at any time.
+   *
+   * @return non-null animator to be started by the ItemAnimator when it sees fit
+   */
+  @Nullable
```
elihart: javadoc says the return is non-null, but the method is annotated nullable

epoxy-adapter/src/main/java/com/airbnb/epoxy/ChangeAnimationDelegator.java
```diff
@@ -0,0 +1,168 @@
...
+      new BiHashMap<>();
+
+  public ChangeAnimationDelegator(@NonNull ItemAnimator decoratedAnimator) {
+    super(decoratedAnimator);
+  }
+
+  public long getChangeDuration() {
+    long maxDuration = super.getChangeDuration();
+
+    for (EpoxyItemAnimation anim : mChangeAnimations.values()) {
```
tschuchortdev: `DefaultItemAnimator` calls it about 2-3 times per changed item. Impossible to know how often a user might call it. 

Granted, that fold could possibly be slow, but is it really worth optimizing? It's not called at a performance critical time (like scrolling) and we'd still have to recompute it at least twice per item change (when the change animation is added to `mChangeAnimations` and removed). It's likely not worth the added complexity unless the user is calling `getChangeDuration` in a loop in which case blame is to be but on them for not caching it. 
Looks like premature optimization to me.

epoxy-adapter/src/main/java/com/airbnb/epoxy/ChangeAnimationDelegator.java
```diff
@@ -0,0 +1,168 @@
...
+      final EpoxyModel oldModel = DiffPayload.getModelFromPayload();
+
+      final EpoxyItemAnimation oldAnim = mChangeAnimations.get(oldHolder);
+
+      if (oldAnim == null || !oldAnim.isRunning()) {
+        EpoxyItemAnimation newAnim = newModel.bindChangesAnimated(viewToBind, oldModel);
+        newAnim.addListener(this);
+        mChangeAnimations.put(holder, newAnim);
+      } else {
+        EpoxyItemAnimation continuedAnim =
```
tschuchortdev: How? We can check if `animation.isStarted()` but the user could still do anything in `bindChangesAnimated`.

I'll add the check anyway, since it's not performance critical and might prevent some accidental misuse.

epoxy-adapter/src/main/java/com/airbnb/epoxy/ChangeAnimationDelegator.java
```diff
@@ -0,0 +1,168 @@
...
+            newModel.bindChangesAnimated(viewToBind, oldModel, oldAnim);
+
+        if (continuedAnim != null) {
+          //cancel old anim if user decided to create a new animator instead of reusing the old one
+          if (continuedAnim != oldAnim) {
+            oldAnim.cancel();
+            continuedAnim.start();
+          }
+
+          continuedAnim.addListener(this);
```
tschuchortdev: Good spot. onStart would call `dispatchAnimationStarted`, but we already did that for the interrupted animation... Might cause a bug. I'll have to consult the example by the RV creators.

epoxy-adapter/src/main/java/com/airbnb/epoxy/ItemAnimatorDecorator.java
tschuchortdev: I don't think so. `SimpleItemAnimator` doesn't seem to have a public `set*Listener` method either, or am I mistaken?

tschuchortdev: Maybe if we have a working prototype we can convince Yigit Boyar that it is indeed possible (and sensible) to delegate change animations and accept my PR to make the listener public.

epoxy-sample/src/main/java/com/airbnb/epoxy/sample/models/ColorModel.java
tschuchortdev: This is actually a remnant of the example in my Kotlin RV library that I copy&pasted. I'll definitely handle interrupted animations in the sample once I have figured out the interface.

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModel.java
tschuchortdev: We can also leave the old name for backwards compatibility, but it wouldn't be as clear. Maybe we can get away with it since 2.0 was released only 5 days ago, but the few people who had time to upgrade would certainly not be amused.

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModel.java
tschuchortdev: Indeed, I just forgot to refactor it yesterday. But it's not necessarily the final name

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModel.java
tschuchortdev: My idea was that you can overwrite only the method with `interruptedAnimation` parameter if you want to handle that, so you don't have duplicate logic in both methods (like with the payloads methods in `Adapter` and `ItemAnimator`).

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyViewHolder.java
tschuchortdev: Fixed it this morning already. Like I said, it's unfinished, but it's still good that you point it out, otherwise I might have missed it.

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModel.java
tschuchortdev: java doc is not up to date (I will come to that when the rest is finished)

epoxy-adapter/src/main/java/com/airbnb/epoxy/ChangeAnimationDelegator.java
```diff
@@ -0,0 +1,168 @@
...
+
+    for (EpoxyItemAnimation anim : mChangeAnimations.values()) {
+      maxDuration = Math.max(maxDuration, anim.getDuration());
+    }
+
+    return maxDuration;
+  }
+
+  public void onAnimationEnd(EpoxyItemAnimation anim) {
+    dispatchAnimationFinished(mChangeAnimations.inverse().get(anim));
```
tschuchortdev: done

epoxy-adapter/src/main/java/com/airbnb/epoxy/BaseEpoxyAdapter.java
elihart: If their custom animator is for insertions/removals and they don't want to do anything for item changes, then would we still run that reflection code in the delegate animator? Just wondering if they might not want to use our delegate for performance reasons if they don't need the item change animation support. Regardless, it seems like there should be _some_ way to override our wrapper if they have some unforseen case. I don't want to force it on users if they don't want it.

I would imagine that most people only want the item change animations so they wouldn't have their own animator that needs wrapping (at least that's the case for our team), so the automatic animator handling would still be awesome for those cases.

epoxy-adapter/src/main/java/com/airbnb/epoxy/ChangeAnimationDelegator.java
elihart: Presumably if somebody is using `ChangeAnimationDelegator` they are using it with an EpoxyController so they know what Epoxy is, but I see your point.

The main reason I like the namespacing is that if you have a large app you might have another class called `ChangeAnimationDelegator` beside's Epoxy's. Then it is confusing which one is which. I see this a lot in our codebase. Small apps wouldn't really have that problem, but I tend to build for our use case and the problems I see on our team.

epoxy-adapter/src/main/java/com/airbnb/epoxy/ChangeAnimationDelegator.java
elihart: yep, good points. just brainstorming

epoxy-adapter/src/main/java/com/airbnb/epoxy/ChangeAnimationDelegator.java
elihart: yeah, mainly just by checking `animation.isStarted()`. Mostly I'm trying to think "what checks can we do to make sure the user is doing what they should", and since we are starting the animation ourselves, this seems like an easy one to validate. Any other ideas on things we can check to make sure the user sets up the animation properly?

epoxy-adapter/src/main/java/com/airbnb/epoxy/ChangeAnimationDelegator.java
elihart: definitely seems tricky. It just stood out as odd, and if we do need to add the listener afterward for some reason we should add a comment about why

epoxy-adapter/src/main/java/com/airbnb/epoxy/ItemAnimatorDecorator.java
elihart: Ah, you're right, I think I'm just mixing things up. In our previous convos you said

> What do you think about restricting users to SimpleItemAnimators. That could certainly make our lifes a lot easier

So I was thinking that if the wrapped animator was a `SimpleItemAnimator` we could simplify this somehow or not need the reflection. You understand the animators much better than me though :P

epoxy-sample/src/main/java/com/airbnb/epoxy/sample/models/ColorModel.java
elihart: awesome

epoxy-adapter/src/main/java/com/airbnb/epoxy/BaseEpoxyAdapter.java
tschuchortdev: > If their custom animator is for insertions/removals and they don't want to do anything for item changes, then would we still run that reflection code in the delegate animator?

yes

Since there would be problems when setting an animator after the adapter, we can't go with the `onAttachedToRecyclerView` approach anyway, so the user will have to set the Decorator themself and thus has complete control over when it's used. 

We can also subclass `DefaultItemAnimator` to avoid reflection when the user wants to use only `DefaultItemAnimator` and not a custom `ItemAnimator`. There should definitely be a way to use any `ItemAnimator` with our decorator though.

We should profile many adds/removes/changes and see if reflection could become a performance problem (what a good opportunity to try the new profiler in AS). I don't think the reflective listener is called when scrolling (except at the end of endless RVs)

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModel.java
elihart: It would be great to only have to implement one method, but it seems we would have to make some changes so `EpoxyItemAnimation bindChangesAnimated(view, previouslyBoundModel, interruptedAnimation)` is actually called with a null animation?

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyViewHolder.java
elihart: yep, sorry if I'm getting too nitpicky with comments at this point since I know it is unfinished. Just wasn't sure what you were planning on changing, and what was a bug you didn't notice, so decided to err on letting you know :P

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModel.java
tschuchortdev: Yeah, we just have to change all `newModel.bindChangesAnimated(viewToBind, previousModel);` calls to `newModel.bindChangesAnimated(viewToBind, previousModel, null);` in `ChangeAnimationDelegator.animateChange`. I meant to do that but forgot amidst changing the interface so often.

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyViewHolder.java
tschuchortdev: No, it's good to have a rigorous code review

epoxy-sample/src/main/java/com/airbnb/epoxy/sample/models/ColorModel.java
tschuchortdev: Speaking of which, I think we should change the Lottie animation in the example as it might be confusing regarding the purpose of `bindChange`. Clicking the view doesn't actually change the state of the item so it should be handled by the view in a simple onClick-listener independent of all RV code. You wouldn't animate button hovering or something like that as a "state change" either. Instead we could use something like the heart animation in the Lottie samples that animates between two discrete states (undliked/liked):

epoxy-adapter/src/main/java/com/airbnb/epoxy/BaseEpoxyAdapter.java
elihart: That all sounds great :+1:

epoxy-sample/src/main/java/com/airbnb/epoxy/sample/models/ColorModel.java
elihart: You're right that using the state change to trigger the Lottie animation is unnecessary, I just did it as an example of how `bind` can be used with the previous model to see the difference, and the lottie animation is kind of fun.

We can change it to something else, but since it is an example project we should do something to show how to use that method.

epoxy-adapter/src/main/java/com/airbnb/epoxy/ChangeAnimationDelegator.java
```diff
@@ -153,16 +203,58 @@ public boolean isRunning() {
...
 
+  @Override
   public boolean canReuseUpdatedViewHolder(@NonNull ViewHolder holder,
       @NonNull List<Object> payloads) {
     // pass empty list to super because it contains only the previous model and DefaultItemAnimator
     // would decide not to cross fade
     return (holder instanceof EpoxyViewHolder && !payloads.isEmpty())
-        || super.canReuseUpdatedViewHolder(holder, Collections.emptyList());
+        || super.canReuseUpdatedViewHolder(holder,
+        Collections.emptyList()); //TODO: only remove our payloads
```
elihart: > //TODO: only remove our payloads

At least for the EpoxyController there can be no other payloads besides ours. Not sure if you want to support EpoxyAdapte

elihart: Based on the `could not get previousModel from change payloads` error below it seems like you want to only support EpoxyController

elihart:
Nice work, are you ready for me to review in depth or do you still have more changes planned?

epoxy-adapter/src/main/java/com/airbnb/epoxy/ChangeAnimationDelegator.java
tschuchortdev: I'll only support controller for now. Maybe we can add support for adapter later on

tschuchortdev:
I'll keep this updated, then squash everything and make a new PR wenn I have a merge candidate ready. Mostly the interface for animations is missing right now... I'm not quite sure yet how to do it elegantly. The use has to cast it back so I can't use any anonymous classes

elihart:
Sounds like a good plan. Let me know if you want me to look at any WIP things before you squash everything into a final PR.

epoxy-sample/src/main/java/com/airbnb/epoxy/sample/models/ColorModel.java
```diff
@@ -54,6 +61,31 @@ public void bindChanges(@NonNull ColorHolder holder,
...
+  @Override
+  public EpoxyItemAnimation bindChangesAnimated(@NonNull ColorHolder view,
+                                                @NonNull EpoxyModel<?> previouslyBoundModel,
+                                                @Nullable EpoxyItemAnimation interruptedAnimation) {
+
+    int currentColor = ((ColorModel) previouslyBoundModel).color;
+    float currentRotation = 0f;
+
+    if(interruptedAnimation != null) {
+      AnimatorSet previousAnim = (AnimatorSet) ((EpoxyItemAnimation.AnimatorWrapper) interruptedAnimation).wrappedAnimator;
```
tschuchortdev: @elihart any ideas how to improve this? Having to cast so much is very annoying

elihart: Yeah, the casting is a shame. The only thing I can think of is using an interface to specify the model and animation type, and that wouldn't allow people to have two different animation types if they wanted.

Can we have the type just be `Animator` instead of `EpoxyItemAnimation`? If we need EpoxyItemAnimation we could wrap it internally after they pass back an `Animator`. That way they can avoid the `EpoxyItemAnimation.fromAnimator(combinedAnim)` and `((EpoxyItemAnimation.AnimatorWrapper) interruptedAnimation).wrappedAnimator` boilerplate at least.

Is it necessary to get the current animation values off of the previous animation? I would have thought that the view would already be updated with that value, and by starting an animation to a new value it would automatically transition from the previous animation.

Using something like https://github.com/florent37/ViewAnimator for a fluent animation description here would be really nice, and we could encourage usage like that.

With something like that this whole method could be replaced with just
```
return holder.cardView.animate()
        .backgroundColor(color)
        .rotationXBy(360f);
```

The problem with that library, and with the built in android `view.animate()` is that they handle the animations themselves instead of returning an animator object. It would be nice to have a fluent animator language that returns the animator, which would make implementing animations here a breeze

I guess this begs the question, if you can get the current animation values off the view, why do you need the previous animator? I'm not sure...

epoxy-sample/src/main/java/com/airbnb/epoxy/sample/models/ColorModel.java
```diff
@@ -86,6 +126,27 @@ private void cancelAnimation(LottieAnimationView lottieView) {
   }
 
   @Override
+  public boolean canAnimateChanges(ColorHolder view) {
+    return false;
+  }
+
+  @Override
+  public Animator animateChanges(ColorHolder view) {
```
tschuchortdev: `animateChanges` and `canAnimateChanges` are from the previous example, I just haven't removed them yet.

epoxy-sample/src/main/java/com/airbnb/epoxy/sample/models/ColorModel.java
tschuchortdev: Unfortunately, the `EpoxyItemAnimation` interface is needed so you can use animations that are not `Animator`s, for example `ViewPropertyAnimatorCompat`, Lottie, etc...

I guess you don't really need the previous animation and could get the values directly from the holder, the same is probably true for `previousModel` as well.

I've been thinking, do we actually need all that stuff? What would happen if we completely ignored all RV conventions and did all the binding in `ItemAnimator.animateChange` or analogously all the animations in `Adapter.bindViewHolder`? The user could hold on to the previous animation in the model and stop all animations in `unbind`.

elihart: I see. Well, Epoxy is API 16+ so people won't need `ViewPropertyAnimatorCompat`. It would be nice to be able to just use `Animator` to simplify things, but I'm not sure how much it would limit people. Since `EpoxyItemAnimation` only supports Animator by default, I'm not sure how many people would actually use it for other things like Lottie since it would be a decent about of extra work to create the wrapper. I imagine in that case they would just manage the animation directly via lottie like I was doing in the sample before.

I think the `previousModel` is useful for knowing which animations to run - eg a "like" animation vs a "vote" animation - if the view supports different animations. Also, what if something changes that they don't want to animate. Maybe they want to animate a "like" change but don't want to animate an image change.

You're right that users could manage animations themselves. I think that removing the overhead of storing the animation and remembering to stop it correctly would be great things to handle automatically though. Otherwise I imagine people will make mistakes doing it themselves. Does the item animator also use the change animation duration to coordinate timing with move and removal animations?

The simplest thing I can think of it have `bindChanges` run during `Adapter.bindViewHolder` and return a Nullable `Animator`. We can store that in the epoxy viewholder and use it in the item animator in `animateChange`. That is quite a change from convention, and a little confusing in it's own right.

I'm thinking that it's a fundamental problem that binding a view and animating changes happen at different times, but are changing the same data. Combining them into one method would make sense in some ways.

tschuchortdev: I mean you could technically get the previous state from the viewHolder too, so `previousModel` is just as redundant as `interruptedAnimation`, apart from being a little more convenient.

 `DefaultItemAnimator`, and probably most third-party animators who are inspired by it, use `getChangeDuration` to play change animations after removals and before additions. Looks somewhat better but it's not necessary. 
Animatons are also not run immediately when `animate*` is called but in a batch in `runPendingAnimations` when RV tells it to. Not sure if this is really needed. It seems to work mostly fine when you run them in `bindViewHolder` and stop in `unbind`.

elihart: It does seem true that you could get previous state off the viewholder, the convenience of the previous model seems nice though.

The other coordination work done by the item animator and RV seems useful, but not necessary as you point out. I don't understand enough about recyclerview animations to know what the other downsides to managing animations manually would be - I'd be interested to hear yigit's take on it.

Still, if we can handle all these little things in an easy way so that people can set up animations fluently in their models then that would be a nice addition to the library.

tschuchortdev: it still seems kinda redundant to me. First you'll have to check the previous model to see if you need to change something, but then you *still* need to check the viewholder in case the previous animation hasn't finished yet. So why not check just the viewholder in the first case? Or you'll have to save your animations somewhere, in the holder probably, which is even more annoying

elihart: You're right, it does seem like it could be redundant in a lot of cases. I wonder if we would be limiting people by not providing the previous model though, or encouraging bad patterns by referencing the viewholder instead of the model.

elihart:
Hey @tschuchortdev , have you had any new ideas or progress on this?

tschuchortdev:
@elihart pretty much the only thing missing is the `EpoxyDefaultItemAnimator` without reflection (apart from the general design flaw I described above). But I probably won't be able to work on it this week due to my tight schedule.

elihart:
Thanks for the update, I know how that goes. good luck with things

tschuchortdev:
@elihart: Any ideas how to implement the `EpoxyDefaultItemAnimator`? It shares all it's logic with the `EpoxyAnimationDelegator` and only differs in the super class, so naturally my first idea was to pull out all the logic into a delegate. Unfortunately, it's not that easy, since the delegate would need to have access to all the super methods of the delegator. Of course something like this would be possible: 

```
interface ItemAnimatorInterface {
    long superGetChangeDuration();
    ...
}

class EpoxyAnimationImpl {
    ItemAnimatorInterface mDelegator;

    EpoxyAnimationImpl(ItemAnimatorInterface delegator) {
        mDelegator = delegator;
    }

     long getChangeDuration() {
        //do stuff and eventually call
        mDelegator.superGetChangeDuration();
     }     
}

class EpoxyAnimationDelegator extends ItemAnimator implements ItemAnimatorInterface {
    EpoxyAnimationImpl mImpl = new EpoxyAnimationImpl(this);    

    @Override
    long superGetChangeDuration() {
        return super.getChangeDuration();
    }

    @Override
    long getChangeDuration() {
        return mImpl.getChangeDuration();
    }

   ...
}
```

however, this is kind of sucks. My second idea would be something like this:

```
class EpoxyAnimationImpl {
    ItemAnimator mSuperAnimator;

    EpoxyAnimationImpl(ItemAnimator superAnimator) {
        mSuperAnimator = superAnimator;
    }

      long getChangeDuration() {
        //do stuff and eventually call
        mSuperAnimator.getChangeDuration();
     }
}

class EpoxyAnimatorDelegator extends ItemAnimator {
     EpoxyAnimationImpl mImpl = new EpoxyAnimationImpl(new ReflectionItemAnimator()); 

    //delegate all calls to mImpl here
}

class EpoxyDefaultItemAnimator extends DefaultItemAnimator {
    EpoxyAnimationImpl mImpl = new EpoxyAnimationImpl(new DefaultItemAnimator()); 

    //delegate all calls to mImpl here
}
```

elihart:
Hm, that's tricky. I'm not sure what the best option is.

Another thing to consider is that with the new view annotations and databinding support we've made it unnecessary to manually create model classes. Instead we could allow defining a callback on a model for creating an animator like 
```
new MyModel_()
  .myData(...)
 .onChange(previousModel -> {
      return new MyAnimation(...);
 });
```

Or alternatively people could still subclass the generated models.

elihart:
Doesn't seem like this is going anywhere

Issue #186, opened 2017-04-11
Allow adding AutoModels implicitly


elihart:
We can allow a system in the `buildModels` controller method where the last touched AutoModel is automatically added to the controller. This allows us to remove the `addTo` method when building a model. We can also remove the `this` parameter in `addIf`

For example the sample from the wiki could be changed to
```
protected void buildModels(List<Photo> photos, Boolean loadingMore) {
      headerModel
          .title("My Photos");

      for (Photo photo : photos) {
        add(new PhotoModel(photo));
      }

      loaderModel
          .addIf(loadingMore);
    }
```

This would work because the generated code instantiates each automodel. The generated model can have a constructor that takes the controller as a parameter, and it can store that controller for later. Anytime a setter is called on the model it would tell the controller to put that model in "staging" mode. the next time any other model is changed, the controller would add the last model in  staging.

This would force a model to be changed all at once. You wouldn't be able to change multiple models in a row, and then add them all at once. IMO this is a better change since it is better practice to group all model changes in one area for readability, and this would force that pattern.

I wouldn't want to enable this by default, since it could be a confusing pattern to understand if you aren't used to it, but we could make it an option in configuration - https://github.com/airbnb/epoxy/wiki/Configuration

PR #194, opened 2017-04-22
Option to implicitly add AutoModels

elihart:
This adds an option to automatically add automodels to a controller https://github.com/airbnb/epoxy/issues/186

If an automodel is modified in `buildModels` it is staged as the next model to add. The next time a different model is changed (or added), the previously staged model is added.

This is off by default.


elihart:
@ngsilverman fyi. I'm planning to add this option to Epoxy, but not necessarily use it in our app unless we agree to it as a team

PR #202, opened 2017-04-27
Initial support for Litho

elihart:
This adds basic support for Litho in a new optional dependency - `epoxy-litho`. 

There is a base litho model class that binds a litho component. The main feature though is that a subclass of the litho model will be completely generated based on every litho spec annotated with @LayoutSpec. The generated model will include fields/setters for each @Prop in the spec.

This means you can just create the spec, and that immediately use the generated model in your EpoxyController. It should make litho integration very easy. the litho models can be used alongside normal epoxy models.

I don't have a great grasp on the litho component lifecycle (bind/rebind/unbind/mount etc) so I'm not sure the best way to integrate the components into the recyclerview viewholder lifecycle. This seems to work fine for now though. I plan to leave this as experimental and continue to iterate on it.

A large part of this change is refactoring the processor code to allow for creating a generated model without EpoxyAttributes and an intermediate model to extend from. This opens the doors for generating models for databinding and views.

@ngsilverman @niccorder @geralt-encore


epoxy-processor/src/main/java/com/airbnb/epoxy/AttributeInfo.java
```diff
@@ -1,251 +1,51 @@
...
-import static com.airbnb.epoxy.ProcessorUtils.capitalizeFirstLetter;
 import static com.airbnb.epoxy.ProcessorUtils.isViewClickListenerType;
-import static com.airbnb.epoxy.ProcessorUtils.startsWithIs;
-import static javax.lang.model.element.Modifier.FINAL;
-import static javax.lang.model.element.Modifier.PRIVATE;
-import static javax.lang.model.element.Modifier.PROTECTED;
-import static javax.lang.model.element.Modifier.PUBLIC;
-import static javax.lang.model.element.Modifier.STATIC;
 
 class AttributeInfo {
```
geralt-encore: shouldn't it be abstract if there is no need to construct it directly anymore?

epoxy-processor/src/main/java/com/airbnb/epoxy/ClassNames.java
```diff
@@ -0,0 +1,17 @@
...
+interface ClassNames {
+  String PKG_LITHO = "com.facebook.litho";
+  String PKG_LITHO_ANNOTATIONS = "com.facebook.litho.annotations";
+
+  ClassName LITHO_COMPONENT = get(PKG_LITHO, "Component");
+  ClassName LITHO_COMPONENT_CONTEXT = get(PKG_LITHO, "ComponentContext");
+  ClassName LITHO_ANNOTATION_LAYOUT_SPEC = get(PKG_LITHO_ANNOTATIONS, "LayoutSpec");
+  ClassName LITHO_ANNOTATION_PROP = get(PKG_LITHO_ANNOTATIONS, "Prop");
+
+  void makeLinterShutUpAboutNoMethodInInterface();
```
geralt-encore: :D
Maybe making it a class instead of an interface and adding `static` manually worth not having linter complaints?

epoxy-processor/src/main/java/com/airbnb/epoxy/GeneratedModelInfo.java
```diff
@@ -24,103 +23,34 @@
...
 import javax.lang.model.type.TypeKind;
 import javax.lang.model.type.TypeMirror;
-import javax.lang.model.util.Elements;
 import javax.lang.model.util.Types;
 
-import static com.airbnb.epoxy.ProcessorUtils.getEpoxyObjectType;
-
-class ClassToGenerateInfo {
-  private static final String GENERATED_CLASS_NAME_SUFFIX = "_";
+class GeneratedModelInfo {
```
geralt-encore: same comment about `abstract`

epoxy-processor/src/main/java/com/airbnb/epoxy/AttributeInfo.java
niccorder: I'm indifferent about this as it is `package private`, and maybe @elihart some sort of potential use case for not making it `abstract`. Also, it seems a bit odd to me to make a simple bean/dto abstract, but then again I do see where you are coming from geralt 😄 .

epoxy-litho/src/main/java/com/airbnb/epoxy/EpoxyLithoModel.java
```diff
@@ -0,0 +1,88 @@
...
+  public void bind(LithoView view) {
+    Component<T> component = getComponent(view.getComponentContext());
+    if (view.getComponentTree() == null) {
+      view.setComponentTree(ComponentTree.create(view.getComponentContext(), component)
+          .asyncStateUpdates(false) // does this need to be off?
+          .layoutThreadLooper(Looper.getMainLooper())
+          .incrementalMount(
+              false) // this needs to be off for recyclerview to work. Should we manually
+          // incrementally mount views in a scroll listener?
+          .layoutDiffing(false) // should we use this?
```
niccorder: from [the documentation](http://fblitho.com/javadoc/com/facebook/litho/ComponentTree.Builder.html#layoutDiffing-boolean-) it seems as though this just does a diff of the view's component tree.

I'd take a look at [Litho's RecyclerBinder for inspiration](https://github.com/facebook/litho/blob/master/litho-widget/src/main/java/com/facebook/litho/widget/RecyclerBinder.java).

epoxy-litho/src/main/java/com/airbnb/epoxy/EpoxyLithoModel.java
```diff
@@ -0,0 +1,88 @@
...
+    // TODO: (eli_hart 4/23/17) Do this at some point to avoid leaking the context?
+    component = null;
+  }
+
+  @Override
+  public void bind(LithoView view) {
+    Component<T> component = getComponent(view.getComponentContext());
+    if (view.getComponentTree() == null) {
+      view.setComponentTree(ComponentTree.create(view.getComponentContext(), component)
+          .asyncStateUpdates(false) // does this need to be off?
```
niccorder: I'm not sure if this needs to be off since [Litho claims to allow you to access a view from any thread](https://github.com/facebook/litho/blob/master/docs/_docs/asynchronous-layout.md#sync-and-async-operations) in a safe mannor... but as I was doing some more research into the topic of recycling I came across [this line of code](https://github.com/facebook/litho/blob/master/litho-widget/src/main/java/com/facebook/litho/widget/RecyclerBinder.java#L172) which seems to show that they themselves haven't even implemented the async recycling correctly yet.

epoxy-processor/src/main/java/com/airbnb/epoxy/AttributeInfo.java
elihart: there should really be some sort of builder, but I got lazy... I can make it abstract at least

epoxy-processor/src/main/java/com/airbnb/epoxy/ClassNames.java
elihart: yeah :( I just hate the boilerplate :P

epoxy-litho/src/main/java/com/airbnb/epoxy/EpoxyLithoModel.java
elihart: yeah, I cloned the repo and dug into the recyclerbinder source code quite a bit. I was wondering if view diffing allows for better animations? I am also curious about the time vs memory tradeoff that the docs mention. I could also build an option into the EpoxyModel to let user's decide

epoxy-litho/src/main/java/com/airbnb/epoxy/EpoxyLithoModel.java
elihart: yep, I noticed that async adapter changes are not yet implemented, but that seems separate from state changes. this seems to only be called by ComponentContext#updateStateAsync which doesn't seem to have usages, so I'm not quite sure how it works, but it seems safer to not do anything async.

elihart:
@niccorder @geralt-encore Thanks for the reviews guys, really appreciate you taking a look! I've updated some of the code style per your suggestions.

I'd like to talk to some facebook engineers soon to get a better idea of how to do this integration well. Then I'll continue building out litho support slowly when I have time.

Issue #198, opened 2017-04-25
Support inheriting @AutoModel models from super classes


elihart:
Right now if you extend an EpoxyController and have @AutoModel annotations in both the subclass and parent, the automodels in the parent will not work in the subclass. We need to add support in the annotation processor to include automodels in the parent.

I didn't design the automodels for inheritance, but it shouldn't be too hard to add. We will also have to consider private/package private AutoModels and not generating helpers for abstract controllers.

If anyone wants to work on this I can provide pointers, but otherwise won't prioritize it.

tomoima525:
Hi @elihart 

Adding below in the proguard file fixed the issue. 

```
-keep class * extends com.airbnb.epoxy.EpoxyController { *; }
-keep class * extends com.airbnb.epoxy.ControllerHelper { *; }
```
Adding this setting at  https://github.com/airbnb/epoxy/blob/master/epoxy-adapter/proguard-rules.pro should solve the issue. Can I create the PR?

elihart:
@tomoima525 A PR would be great, thank you!

I don't think the `-keep class * extends com.airbnb.epoxy.ControllerHelper { *; }` line is needed since the EpoxyController classes are not referenced reflectively, only the ControllerHelpers are. Let me know if it doesn't work without that though.

I think we also need to add obfuscation protection. Judging by what Butterknife and Icepick does we would need something like 
```
-keepclasseswithmembernames class * { @com.airbnb.epoxy.AutoModel <fields>; }
```
I'm not too familiar with proguard though so if you understand it better feel free to improve on that.

tomoima525:
`-keep class * extends com.airbnb.epoxy.ControllerHelper { *; }` was needed because auto-generated EpoxyHelper class extends ControllerHelper. I took out `com.airbnb.epoxy.ControllerHelper { *; }` and didn't work.


```
/**
 * Generated file. Do not modify! */
public class FooAdapter_EpoxyHelper extends ControllerHelper<FooAdapter> {...}
```

on the other hand `keepclasseswithmembernames ` is probably not required since `AutoModel ` is just an annotation.

Btw, I was trying to reproduce this issue with `epoxy-sample` project, but epoxy-sample shows a white screen and does not load MainActivity View. (Tried with Android Emulator(OS M, N) and the real device N9(7.1.2))

I just updated Android Studio to 2.3.1, so that might be the cause, but can you take a look if the view shows up?

[...]

Issue #197, opened 2017-04-25
ProGuard issue with @AutoModel in my BaseController


jbmlaird:
Are you guys using any ProGuard rules with your autogenerated classes?

I tried `@AutoModel`ing a `HeaderModel` in a `BaseController` class but whenever I have a `Controller` that uses this base class, it will throw a NPE. Perhaps it's my implementation

```
public abstract class BaseController extends EpoxyController
{
    public String title = "";
    public String subtitle = "";
    public String imageUrl = "";

    // Can't use this as ProGuard causes issues with AutoModels
    // public @AutoModel HeaderModel_ header;

    public void setTitle(String title)
    {
        this.title = title;
    }
}
```

My stacktrace:

```
java.lang.RuntimeException: Unable to start activity ComponentInfo{bj.rxjavaexperimentation/bj.rxjavaexperimentation.artist.ArtistActivity}: java.lang.NullPointerException: Attempt to invoke virtual method 'bj.rxjavaexperimentation.epoxy.common.c bj.rxjavaexperimentation.epoxy.common.c.u(android.content.Context)' on a null object reference
        at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2924)
        at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2985)
        at android.app.ActivityThread.-wrap14(ActivityThread.java:-1)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1635)
        at android.os.Handler.dispatchMessage(Handler.java:102)
        at android.os.Looper.loop(Looper.java:154)
        at android.app.ActivityThread.main(ActivityThread.java:6692)
        at java.lang.reflect.Method.invoke(Method.java:-2)
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1468)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1358)
Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'bj.rxjavaexperimentation.epoxy.common.c bj.rxjavaexperimentation.epoxy.common.c.u(android.content.Context)' on a null object reference

        at bj.rxjavaexperimentation.artist.ArtistController.void buildModels()(ArtistController.java:39) <---- here

        at com.airbnb.epoxy.EpoxyController.void dispatchModelBuild()(EpoxyController.java:126)
[...]

PR #205, opened 2017-04-29
Fix @AutoModel Model NPE crash with Proguard

tomoima525:
Issue:
Epoxy Model class annotated with `@AutoModel` crashes with proguard setting on (ref #197, #198)

Cause:
Epoxy model was accessed from classes that extend ControllerHelper, EpoxyController which were obfuscated by proguard

Fix:
Apply proguard rules to autogenerated classes

### DIFF
--- a/epoxy-adapter/proguard-rules.pro
+++ b/epoxy-adapter/proguard-rules.pro
@@ -15,3 +15,8 @@
 #-keepclassmembers class fqcn.of.javascript.interface.for.webview {
 #   public *;
 #}
+
+-keep class * extends com.airbnb.epoxy.EpoxyController { *; }
+-keep class * extends com.airbnb.epoxy.ControllerHelper { *; }
+-keepclasseswithmembernames class * { @com.airbnb.epoxy.* <methods>; }
+-keepclasseswithmembernames class * { @com.airbnb.epoxy.* <fields>; }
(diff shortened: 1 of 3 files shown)


tomoima525:
@elihart 

So I checked with the sample and keeping `EpoxyController,ControllerHelper` worked.
Although keeping keepclasseswithmembernames AutoModel did not effect app build, I included anyway just in case

tomoima525:
By the way the sample app can not be built with the  latest master. #204 might be affecting

epoxy-adapter/proguard-rules.pro
```diff
@@ -15,3 +15,8 @@
 #-keepclassmembers class fqcn.of.javascript.interface.for.webview {
 #   public *;
 #}
+
+-keep class * extends com.airbnb.epoxy.EpoxyController { *; }
+-keep class * extends com.airbnb.epoxy.ControllerHelper { *; }
+-keepclasseswithmembernames class * { @com.airbnb.epoxy.AutoModel <methods>; }
+-keepclasseswithmembernames class * { @com.airbnb.epoxy.AutoModel <fields>; }
```
elihart: I don't think `keepclasseswithmembernames class * { @com.airbnb.epoxy.AutoModel <methods>;` is needed. The automodel annotation is only used on fields, not on methods.

elihart: Just add a new line at the end of the file here so that lint will pass

tomoima525:
Thanks for merging! And my apology I did not check the lint 😬

Issue #187, opened 2017-04-11
Completely generate DataBinding models


elihart:
I am almost done with https://github.com/airbnb/epoxy/issues/183 and it got me thinking that all somebody should really have to do is define a list of xml layouts that they want to use with databinding, and we could generate the whole EpoxyModel for them.

For example, with the upcoming support they would have to write a model like
```
@EpoxyModelClass(layout = R.layout_my_databinding_layout)
public abstract class MyModel extends DataBindingEpoxyModel {
   @EpoxyAttribute int variable1;
   @EpoxyAttribute boolean variable2;
    ... more attributes
}
```

but instead we  could have them define layouts in an annotation like ```@EpoxyDataBindingLayouts({R.layout_my_databinding_layout1, R.layout_my_databinding_layout1})```
and the annotation processor could inspect the R tree at process time to get the data variables and build the model completely. This is safer too because it guarantees that the variable names of the model stay in sync with the data names in the layout.

I don't think this is too hard - the R parsing would be the hardest part. It isn't super important because it isn't a lot of work to create a model like above, but it would be a nice boilerplate saver and I think it would be pretty cool.

geralt-encore:
Sounds pretty cool! Any ideas about R parsing? Something similar to this part which was taken from Butterknife?
Just wanna to double-check. So your idea is to parse data variables from layout's XML -> generate handwritten-like model (like in example) -> feed it to annotation processor and generate final model?

elihart:
@geralt-encore yeah, it would be cool! I just put up my WIP code for the databinding at https://github.com/airbnb/epoxy/pull/188 which this would build upon.

It seems like you understand it. Although I don't think we would actually generate the handwritten model, since then we would need to generate the `_` model (the normal generated model) on top of that. That could work, but we would need to make sure that the annotation processor handles rounds correctly. Alternatively we could just only make the `_` class to simplify. I don't see a need for the normal model. We would just need to add the fields to the generated model.

I'm thinking the generated model name would be based on the xml name. Camel cased, and possibly appended with something (`Model`? or at least `_`?)

And yes, I am thinking of parsing the R tree like butterknife does. I removed that R tree parsing in Epoxy since I figured out a simpler approach, but we could add it back in.

Basically for a layout like 
```
<layout xmlns:android="http://schemas.android.com/apk/res/android">

    <data>
        <variable
            name="textRes"
            type="int" />

        <variable
            name="clickListener"
            type="android.view.View.OnClickListener" />
[...]

PR #207, opened 2017-05-02
Completely generate DataBinding models

geralt-encore:
https://github.com/airbnb/epoxy/issues/187
This is a really rough implementation so, please, review it thoroughly. I feel like it can be improved a lot but I just wanted to make a PR as soon as possible to be able to discuss it.
Any ideas how to properly test it?


epoxy-processor/src/main/java/com/airbnb/epoxy/DataBindingAttributeInfo.java
```diff
@@ -1,4 +1,22 @@
...
+
 class DataBindingAttributeInfo extends AttributeInfo {
+
+  public DataBindingAttributeInfo(DataBindingModelInfo modelInfo, Element dataBindingElement) {
+    name = ProcessorUtils.decapitalizeFirstLetter(dataBindingElement.getSimpleName().toString());
+    typeName = TypeName.get(dataBindingElement.asType());
+    typeMirror = dataBindingElement.asType();
+    modelName = modelInfo.getGeneratedName().simpleName();
+    modelPackageName = modelInfo.getGeneratedName().packageName();
+    useInHash = true;
```
elihart: for `useInHash` we probably need to exclude certain things, like ClickListeners. I have the same problem and a TODO with the litho models and models generated from Views. we can leave this as is and solve them all at once later on, but let me know if you have ideas.

geralt-encore:
@elihart I can add a more complicated model for code generation test if you think that current one is not enough.

epoxy-processortest/src/test/java/com/airbnb/epoxy/DataBindingModelTest.java
```diff
@@ -96,4 +99,23 @@ public void testSimpleModelNoValidation() {
         .and()
         .generatesSources(generatedModel);
   }
+
+  @Test
+  public void testFullyGeneratedModel() {
+    JavaFileObject packageInfo = JavaFileObjects
```
elihart: possibly define the package-info file as a field in this class (like R and BR), since `package-info` is otherwise very vague and might be mixed up with other tests.

PR #216, opened 2017-05-17
#198 Support inheriting @AutoModel models from super classes

geralt-encore:
https://github.com/airbnb/epoxy/issues/198


geralt-encore:
No problem! I like to work on Epoxy 😊

Issue #219, opened 2017-05-21
Which is the most efficient way to write a model in Kotlin


kokeroulis:
Hello,

I am trying to port my java models into kotlin, so this is my model so far

```
@EpoxyModelClass(layout = R.layout.layout_foo)
abstract class BookModel : EpoxyModel<FooView>() {
    open @EpoxyAttribute var bookName: String = ""
    open @EpoxyAttribute var isRead: Boolean = false
    open @EpoxyAttribute var author: String? = null
    open @EpoxyAttribute(DoNotHash) var listener: View.OnClickListener? = null

    override fun bind(view: FooView?) {
    }

    override fun bind(view: FooView?, previouslyBoundModel: EpoxyModel<*>?) {
    }

    override fun unbind(view: FooView?) {
    }
}
```

So is this the correct way of doing it? Or i have overlooked into something?

Thank you

kokeroulis:
Sorry i forgot to mention that the above is code produces a generated model with infinite loop inside... This applies only to booleans that they have a name as "isSomething" for example

![screen shot 2017-05-22 at 09 31 39](https://cloud.githubusercontent.com/assets/470064/26295266/bcb7036c-3ed1-11e7-858d-ee84f0d60b85.png)

For example the above does work but as soon as you rename the "isRead" property to "read" everything works fine.

In the generated model there is the following code

 ```
  public boolean isRead() {
    return isRead();
  }
```
So can we consider this one a bug of the annotation processor or it is a "misuse" of mine?

geralt-encore:
It is definitely a bug. I'll look into it. Thanks for reporting =)

geralt-encore:
It is definitely a bug. I'll look into it. Thanks for reporting =)

elihart:
Thanks @geralt-encore !

I still haven't used much Kotlin, and not with Epoxy, but can you also set up the model like this?
 `BookModel(@EpoxyAttribute var bookName: String)`

I can update the Wiki for kotlin model recommendations.

@kokeroulis If you wait a few weeks for https://github.com/airbnb/epoxy/issues/171 you may not have to write the model at all btw, so it might not be worth porting all your models if it will be a lot of work for you.

No guarantees on when that feature ships or if it works perfectly for your use case (hopefully with time it'll get there though!), but I'm done with the alpha and we'll be testing it internally this week at least.

elihart:
Fixed with https://github.com/airbnb/epoxy/pull/220

[...]

PR #220, opened 2017-05-22
#219 Fix recursive getter calls

geralt-encore:
https://github.com/airbnb/epoxy/issues/219

### DIFF
--- a/epoxy-processor/src/main/java/com/airbnb/epoxy/AttributeInfo.java
+++ b/epoxy-processor/src/main/java/com/airbnb/epoxy/AttributeInfo.java
@@ -94,6 +94,11 @@ abstract class AttributeInfo {
     return isPrivate ? getterMethodName + "()" : name;
   }
 
+  // Special case to avoid generating recursive getter if field and its getter names are the same
+  String superGetterCode() {
+    return isPrivate ? String.format("super.%s()", getterMethodName) : name;
+  }
+
   String setterCode() {
     return (isGenerated ? "this." : "super.")
         + (isPrivate ? setterMethodName + "($L)" : name + " = $L");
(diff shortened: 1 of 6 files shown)


epoxy-sample/src/main/java/com/airbnb/epoxy/sample/models/ColorModel.java
```diff
@@ -28,6 +28,16 @@
   @EpoxyAttribute boolean playAnimation;
   @EpoxyAttribute(DoNotHash) View.OnClickListener clickListener;
 
+  @EpoxyAttribute private boolean test;
```
elihart: Did you mean to include this?

If we want a test we can put it in the integration test module, since otherwise this sample model might be changed later if we forget it is intended as a test

geralt-encore: Nope, it was just for some initial testing. I'll remove it

epoxy-processor/src/main/java/com/airbnb/epoxy/AttributeInfo.java
```diff
@@ -94,6 +94,11 @@ String getterCode() {
     return isPrivate ? getterMethodName + "()" : name;
   }
 
+  // Special case to avoid generating recursive getter if field and its getter names are the same
+  String superGetterCode() {
```
elihart: would it be possible to combine this with the existing `getterCode` method? I did something similar with `setterCode()` to handle the super case.

I would worry about mixing up `superGetterCode` and `getterCode` and confusion over when to use each. It seems better to have one method that handles it all.

elihart:
thanks for fixing this so quickly, just had some small comments.

epoxy-processor/src/main/java/com/airbnb/epoxy/AttributeInfo.java
geralt-encore: Yeah, I don't like this solution either, but I couldn't find anything better.
I tried to handle it in the more general way same as in `setterCode()`, but failed because getter is used in `equals` and it messes it up.

elihart: good point. I guess this is fine for now, if we have problems in the future we can refactor then

PR #222, opened 2017-05-22
Use view annotations to completely generate models

elihart:
View classes can be annotated so a model is generated for them. Setters in that view can be annotated to have an attribute for that setter included in the model. There is also some configuration for handling unbinding, defaults, method overloads, and some other things.

We will be testing this internally for a little while and iterating on this under a `2.2.0-SNAPSHOT` release.

Naming is still mostly undecided, suggestions are welcome!

@ngsilverman @geralt-encore


epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModel.java
```diff
@@ -469,6 +471,23 @@ public int getSpanSize(int totalSpanCount, int position, int itemCount) {
     return 1;
   }
 
+  public EpoxyModel<T> spanSizeCallback(@Nullable SpanSizeCallback spanSizeCallback) {
```
elihart: I added this so that span size can be determined in a callback instead of inside the model. Otherwise the models generated from views can't handle programmatically changing span size well.

elihart:
I still need to write a lot of tests for most of this stuff. I'll be working on that while we test it internally.

The annotation processor code is also starting to get messy, and it would probably be a good idea to refactor at some point... the data structures are getting weird since we have to support the new view features alongside the original model/attribute code.

geralt-encore:
Yep, annotation processor code definitely becomes harder and harder to follow 😔

elihart:
Yeah :( I'm thankful we have so many tests for the existing code so I don't break things as I work on this :P

elihart:
I could probably migrate the annotation processor code to kotlin to clean it up and have better null safety

geralt-encore:
Just wanted to ask you exactly about that 😄 
I did it some time ago for [StorIO](https://github.com/pushtorefresh/storio) and it was definitely a huge improvement to readability and code quality in general.
I would be happy to help you with that.

elihart:
that would be awesome!

I'd like to get 2.2.0 out first so the processor code is stable and there are more tests written, then we can start switching to kotlin.

elihart:
This is mostly done and pretty stable. We are using it and it seems great. Small name changes or features may still change while I continue polishing. Current version is `2.2.0.8-SNAPSHOT'` if you'd like to try it.

elihart:
Merging now so I can continue with some more small changes without having them lost in this giant.

PR #240, opened 2017-07-10
Initial conversion of annotation proccessor to Kotlin

geralt-encore:
I converted the easiest and smallest classes to Kotlin, for now, to make it more incremental and easier to review.


elihart:
Awesome! a bit swamped atm, but excited to review this later this week

epoxy-processor/src/main/java/com/airbnb/epoxy/ControllerModelField.kt
```diff
@@ -0,0 +1,36 @@
...
+
+import com.airbnb.epoxy.Utils.isFieldPackagePrivate
+
+class ControllerModelField(element: Element) {
+
+    var fieldName: String = element.simpleName.toString()
+    var typeName: TypeName = TypeName.get(element.asType())
+    var packagePrivate: Boolean = isFieldPackagePrivate(element)
+
+    override fun equals(other: Any?): Boolean {
```
elihart: could this be a data class so we don't need equals/hashCode/toString manually declared?

epoxy-processor/src/main/java/com/airbnb/epoxy/DataBindingAttributeInfo.kt
```diff
@@ -0,0 +1,26 @@
+package com.airbnb.epoxy
+
+import javax.lang.model.element.ExecutableElement
+import javax.lang.model.element.VariableElement
+
+import com.airbnb.epoxy.Utils.removeSetPrefix
+
+internal class DataBindingAttributeInfo(
```
elihart: nit: I'd like to follow the conventions at https://kotlinlang.org/docs/reference/coding-conventions.html and have the closing parenthesis be on a new line

epoxy-processor/src/main/java/com/airbnb/epoxy/ControllerModelField.kt
geralt-encore: The thing is that data class primary constructor must have only property parameters so if we want this class as a data class we need to change its constructor and init properties on the caller side. If you think that it makes sense - we can change it.

elihart: @geralt-encore Ah right. What if it is a data class with a private constructor and a companion with a factory method to build the class based on an Element?

I'm not sure if that is a common pattern, but it seems a little more elegant than manually implementing everything.

I'll go ahead and merge now just to get it in, but I'm curious what you think about that pattern ^

geralt-encore: Yep, you are right. I haven't thought about it but it does look like a good approach. I'll fix it in upcoming PR.

elihart: @geralt-encore cool, thanks!

btw, buck is on the verge of supporting Kotlin so we're very close to introducing it in our app. So, I've been studying up on it and we're trying to figure out which patterns to introduce to our team, and I'm curious to see how you do things :)

Issue #165, opened 2017-03-22
Rearrangeable support


elihart:
We are working on a small addition to Epoxy to better allow for using item touch interceptors for rearranging or swiping models.

This will come in a future 2.x release

Sroka:
Yes, please :)

This is the only lacking feature that is keeping me from moving to epoxy

elihart:
@Sroka Thanks for the feedback. what exactly would you like to see in this? swiping? drag and drop rearranging? something more? The more details you can give us the better we can design this for other people's needs.

What we're probably going to end up with is a subclass of item touch helper that helps deal with the epoxy models, and possibly interfaces to automatically call back to models that have been touched. We are leaning towards a simpler system like this so we can keep it very flexible

thenickreynolds:
Since there's additional interest I can put some time into this. Won't take
much and current time consumers are winding down.

elihart:
Thanks @thenickreynolds! let me know if I can help

Sroka:
@elihart Well, those APIs of RecyclerView are quite tricky to handle properly. Mostly it is coming from a lot of combinations that has to be handled. Like, what should happen when:

- Data is displayed in a vertical list. Drag and drop should be supported. Swiping as well. What should be  drag&drop behavior for expandable list items? We should be able to set some bounds limiting how far user can drag specific item. 
- Data is displayed in a horizontal list. Drag and drop should be supported. Swiping as well but this time there should be option to swipe vertically.
- Data is displayed in a grid. Drag and drop should be supported. Swiping probably not, unless we want to allow user to be able to swipe item only within its borders but this is bad UX probably.

While using other libraries I found that keeping the state of underlying data can be tricky. For example, User drags&drops item to some other place. We then, usually, make a request to our server that puts rearranged data there. Request can end with 200 what means items should be kept as user rearranged them or with some error and items should be put back on their places. The same goes for swiping. 

Additionally, when item is swiped out gmail app, for example, leaves in place of an item a button for user to undo archive action. Being able to handle something like this would be nice.

elihart:
@Sroka Thanks for the details! sounds like you're not looking for a specific use case, you just want good general support for touch.

You're right that there are many combinations and settings to use with the touch API's. You also mentioned a swipable view that shows a different view behind it as you swipe, as well as trickiness with updating data.

[...]

PR #250, opened 2017-07-31
Add method to move controller model

elihart:
Adds a method on the EpoxyController to efficiently move a model's position without rebuilding all models. This is intended to enable rearranging via item touch helper for https://github.com/airbnb/epoxy/issues/165

I'm going to follow up with an addition to the sample app and the wiki.

@thenickreynolds


epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModelGroup.java
```diff
@@ -237,7 +237,7 @@ protected final Holder createNewHolder() {
     private ViewGroup rootView;
 
     /**
-     * Get the root view group that holds all of the model views. You can override {@link
+     * Get the root Oview group that holds all of the model views. You can override {@link
```
thenickreynolds: typo

elihart: thanks

PR #255, opened 2017-08-05
#254 Inherit annotations in epoxy model generated classes

geralt-encore:
https://github.com/airbnb/epoxy/issues/254


geralt-encore:
Cool! I was trying to convert all `AttributeModel` model to Kotlin, but it turned out to be pretty hard to do in a Kotlin-way so I gave up for now =(

elihart:
Yeah, those classes are unfortunately a mess and overly complicated. they need a refactor, thanks for trying though :)

geralt-encore:
`ViewAttributeInfo` was the most complex one for me. Still have close to zero ideas what's going on in it.

Issue #270, opened 2017-08-29
Create Kotlin DSL for building models in controllers


elihart:
We can generate an extension function for each model, so building models would look like
```
buildModels() {
  headerView {
         id("header")
         title("my title")
  }
}
```

geralt-encore:
I'll try to look into it if you don't mind. Looks very interesting!

elihart:
I've actually already started :) I got kotlin poet hooked up and the basics working, it's exciting!

geralt-encore:
Got it. Have fun! Feel free to ping me if you need any help =)

elihart:
thanks :) you can check it out at https://github.com/airbnb/epoxy/compare/eli-kotlin

It's actually pretty straightforward, figuring out kotlin poet was the tricky part, and I'm updating to android studio 3.0 at the same time :P

elihart:
@geralt-encore it would be good to get your feedback on a few ideas on this actually.

1. don't use AutoModel with this and instead declare id as a string, as in the example. I think grouping all the code for a single model like that is easier to read, but I wish there was a better way to force somebody to add the id, if you forget it it is a runtime error, not sure how to build that into the DSL

2. I want to support a condition that when false would skip that model. I'm thinking
```
headerView(condition = myBoolean) {
         id("header")
         title("my title")
  }
```
I like that this is concise, and it is a nice optimization since the model is never created if the condition fails. not sure if there is a better syntax for this though

3. The generated model interface is quite messy right now. There is a ton of old features included on it and things that don't apply when building the model (show/hide/bind/equals/toString etc) plus all of the getters. This makes autocomplete confusing and makes it hard to understand what the actual options are for setting properties on the model. I'm thinking we can generate an interface that contains all of the actual setters and only expose that interface as the receiver of the lambda. Then autocomplete would only suggest  `id` and `title`

4. Should the function return the generated model? I'm thinking not, since at that point its already added to the controller and i don't want to encourage funky behavior by making it easy to keep a reference to it.

5. I'd like the function name to remove "model" wording from it. If the model is `HeaderViewModel` the generated function would be `headerView`. In the context of building models I think appending `model` is redundant and verbose

6. I'm thinking that the functions should be in the same package as the models. All models in the same package will have their own file

elihart:
An example of a generated extension function is 
```
inline fun EpoxyController.kotlinView(condition: Boolean = true,
    modelInitializer: KotlinViewModel_.() -> Unit) {
  if (condition) {
[...]

PR #278, opened 2017-09-05
Generate kotlin extension functions for building models

elihart:
https://github.com/airbnb/epoxy/issues/270

I also bumped support to Android Studio 3.0, which required tweaking quite a bit of things to make everything work.


epoxy-processor/src/main/java/com/airbnb/epoxy/KotlinExtensionWriter.kt
```diff
@@ -0,0 +1,158 @@
...
+        models.map { buildExtensionForModel(it) }
+                .forEach { fileBuilder.addFun(it) }
+
+        return fileBuilder.build()
+    }
+
+    private fun buildExtensionForModel(model: GeneratedModelInfo): FunSpec {
+        val constructorIsNotPublic = model.constructors
+                .filter { it.params.isEmpty() }
+                .any { !it.modifiers.contains(Modifier.PUBLIC) }
```
geralt-encore: could be changed to more idiomatic Kotlin:
`Modifier.PUBLIC !in modifiers`

epoxy-processor/src/main/java/com/airbnb/epoxy/KotlinExtensionWriter.kt
```diff
@@ -0,0 +1,158 @@
...
+
+fun String.lowerCaseFirstLetter(): String {
+    if (isEmpty()) {
+        return this
+    }
+
+    return Character.toLowerCase(get(0)) + substring(1)
+}
+
+typealias JavaClassName = com.squareup.javapoet.ClassName
```
geralt-encore: wow, it is a perfect use-case for typealiases  👍

epoxy-processor/src/main/java/com/airbnb/epoxy/KotlinExtensionWriter.kt
elihart: great point

epoxy-processor/src/main/java/com/airbnb/epoxy/KotlinExtensionWriter.kt
elihart: yeah :) I did it without them at first with FQN's and it was terrible 

Thanks for looking!

PR #280, opened 2017-09-12
Include canvas in ItemTouchHelper swipe callback

elihart:
This changes the `onSwipeProgressChanged` callback in EpoxyTouchHelper to take Canvas as a parameter. This will allows users to draw to the canvas to have more control in how the swipe is animated.

This is a breaking change to the touch helper API. I am ok with this because
1. It is fairly new and won't have much usage yet
2. This should be helpful to people
3. It is an easy change to make and probably won't have to be updated in many places

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2,7 +2,7 @@
 buildscript {
 
   ext.KOTLIN_VERSION = "1.1.4-3"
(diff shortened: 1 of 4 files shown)


skywalkerDavid:
LGTM, thank you for adding this. :+1:

Issue #293, opened 2017-09-23
Block clicks on views that are being removed


elihart:
We can check that position != NO_POSITION in our click listener wrapper.

I am also considering adding some debouncing support

niccorder:
I have some free time I'll hop on this. Let's save the debounce support for another issue, but I'm all for supporting debouncing!

elihart:
Great!

The easiest way to do this now would be to change `WrappedEpoxyModelClickListener#onClick` to check the adapter position before calling onClick of the nested listener.

That would be a very easy change, but it only helps people that used `OnModelClickListener`. I'm considering changing the annotation processor to wrap all click listeners with `WrappedEpoxyModelClickListener` so everything gets this protection. A downside is it creates extra click listener objects.

When we want to do the debouncing we would also change `WrappedEpoxyModelClickListener#onClick` and do something like https://github.com/JakeWharton/butterknife/blob/master/butterknife/src/main/java/butterknife/internal/DebouncingOnClickListener.java

Butterknife uses a post, but I'm wondering if just checking system clock is more efficient, and also if we may want to delay slightly longer, since requests to rebuild epoxy models are async and posted a frame later, so any changes do the adapter due to a click would take a frame to get notified (even longer if we add diffing)

niccorder:
Thanks for the detailed feedback! I'll touch on the two topics here.

### Blocking Clicks
The code change itself is minute, but the interesting aspect here is refactoring to allow for click tests to pass.

I'll throw up my code once I'm back on my laptop for your guidance, and to give clarity into what I mean.

### Debouncing Clicks
I was planning on doing it the ButterKnife way originally but you bring up a great point with using the system clock + the extra required delay.

I think we can solve this by creating a custom handler and using a combination of both ideas. I think it would look like:

```java
public class DebounceHandler extends Handler {
    private long debounceTime;
    private long lastPosted;

    @Override
    public void post(Runnable r) {
         // If lastPosted - currentTime < debounceTime -> remove message
         // Update last posted time.
         // Store runnable in map
         // PostDelay message by debounce time
   }
}
```

On my mobile phone and haven't thought it out filly yet but this might suffice

elihart:
Closed this since it is fixed, lets take the debounce discussion to https://github.com/airbnb/epoxy/issues/305

elihart:
A handler does synchronization and other in depth stuff under the hood to post, just checking the system clock might be good enough

PR #301, opened 2017-10-02
Block item click callback during item layout.

niccorder:
Adding a `RecyclerView.NO_POSITION != viewHolder.getAdapterPosition()`
check inside of the click listener wrapper ensures we do not call the
click listener when the item is going through a layout.

There was a need to make a change to theControllerLifecycleHelper 
to ensure that click still passed as the adapter was always supplying
`NO_POSITION` for unit tests. The simplest way to get the basic
functionality was to spy the `ViewHolder`, and stub this method.

ISSUE #293

### DIFF
--- a/epoxy-integrationtest/src/test/java/com/airbnb/epoxy/ControllerLifecycleHelper.java
+++ b/epoxy-integrationtest/src/test/java/com/airbnb/epoxy/ControllerLifecycleHelper.java
@@ -8,6 +8,9 @@ import org.robolectric.RuntimeEnvironment;
 import java.util.Collections;
 import java.util.List;
 
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.when;
+
 class ControllerLifecycleHelper {
   private EpoxyViewHolder viewHolder;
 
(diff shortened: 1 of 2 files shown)


epoxy-integrationtest/src/test/java/com/airbnb/epoxy/ControllerLifecycleHelper.java
```diff
@@ -36,11 +39,16 @@ public View bindModel(EpoxyModel<?> model) {
...
-        itemViewType
+    final EpoxyViewHolder viewHolder = spy(
+        adapter.onCreateViewHolder(
+            new FrameLayout(RuntimeEnvironment.application),
+            adapter.getItemViewType(position)
+        )
     );
+
+    // The simplest way to inject the position for testing.
+    when(viewHolder.getAdapterPosition()).thenReturn(position);
```
niccorder: Although I feel this isn't ideal, I found it to be the least obtrusive. Would love feedback if anyone has a more proper idea on how to fix.

elihart: This seems reasonable, thanks for fixing the tests too 👍

PR #302, opened 2017-10-02
Replace UnmodifiableList with java collection's.

niccorder:
This is a small clean-up to reduce a class which is not necessary as we
can use the `Collections.unmodifiableList(...)` instead of using the
custom `UnmodifiableList` epoxy had itself. This is not a refactor of
functionality, but redundancy.


elihart:
Not sure why I didn't use this originally... :P Thanks!

niccorder:
No problem! 

Noticed this class and thought it was probably accidental :D

Issue #285, opened 2017-09-15
Integrate with Paging Library component


elihart:
We should be able to create an EpoxyController subclass that makes it easy to work with this new component - https://developer.android.com/topic/libraries/architecture/paging.html

abeemukthees:
I'm also looking for this integration.

elihart:
I don't think it will be _too_ hard, but I also haven't looked at the source yet. I can maybe do it next week, but if someone else wants to jump on it that'd be great :)

geralt-encore:
I wouldn't mind trying to take look at it. Can't promise anything tho

elihart:
I actually was able to get started on this https://github.com/airbnb/epoxy/compare/eli-paging

Main work is in https://github.com/airbnb/epoxy/compare/eli-paging#diff-4c2e75cdcce43d0f970a208f344fb1c4

I haven't tested it yet, but it seems pretty straightforward. The one issue I can see is that we would have to iterate over all items in the list, including all the null (unloaded) ones. I'm looking for a way around that.

geralt-encore:
Yep, looks pretty straightforward. Let me know if you'll need any help =)

elihart:
What I have seems to work fine. There is a performance issue though as more items are loaded into the paged list. The paged list starts with just a few items, but as more items are loaded it seems like the list only ever grows larger - eg items from the beginning are not removed as you scroll down and  more items are loaded at the end. 

This is an issue because Epoxy is set up to build a model for each item, but it gets slow to build hundreds or thousands of items and diff them.

I am thinking that we can build a layer of paging on top of the paged list. Basically Epoxy can track where in the list the current bound items are, and only build models for the items around that. As you scroll epoxy would rebuild models as needed.

This seems doable, but will take some tweaking to optimize. The benefit is this will actually be useful for all list cases, not just the new Paged List arch component.

In my branch I've added a working sample. If anyone wants to take a look and discuss feel free to leave feedback here

elihart:
Good progress with https://github.com/airbnb/epoxy/pull/306

elihart:
Closing this since a basic version is done. We can track improvements in separate dedicated issues

PR #306, opened 2017-10-04
Paging EpoxyController

elihart:
For https://github.com/airbnb/epoxy/issues/285

This seems to be working well, it just needs tests.

I'll release it as beta in the upcoming 2.6.0 release.

A lot of this code is setting up a sample, the main code is in EpoxyPagingController -
 https://github.com/airbnb/epoxy/pull/306/files#diff-e18d5f431b3454795125f683cf64f9e7R35


epoxy-pagingsample/src/main/java/com/airbnb/epoxy/pagingsample/PagingSampleActivity.kt
```diff
@@ -0,0 +1,96 @@
...
+                }
+
+                return@bg PagedList.Builder<Int, User>().run {
+                    setDataSource(db.userDao().dataSource)
+                    setMainThreadExecutor(UiThreadExecutor)
+                    setBackgroundThreadExecutor(AsyncTask.THREAD_POOL_EXECUTOR)
+                    setConfig(PagedList.Config.Builder().run {
+                        setEnablePlaceholders(false)
+                        setPageSize(40)
+                        setInitialLoadSizeHint(80)
```
pcqpcq: Accord to the [doc](https://developer.android.com/reference/android/arch/paging/PagedList.Config.Builder.html#setInitialLoadSizeHint(int)), this will be ignored currently when using TiledDataSource.

`Defines how many items to load when first load occurs, if you are using a KeyedDataSource.
If you are using an TiledDataSource, this value is currently ignored`

Issue #314, opened 2017-10-13
Is it a good practice to get current list inside PagingEpoxyController


pcqpcq:
I found nothing like [getCurrentData()](https://github.com/airbnb/epoxy/blob/master/epoxy-adapter/src/main/java/com/airbnb/epoxy/TypedEpoxyController.java#L58)  in PagingEpoxyController, so I create one like this:

```java
public class MyController extends PagingEpoxyController<Foo> {

    // ...other stuff

    private PagedList<Foo> currentData;

    @Override
    public void setList(@Nullable PagedList<Foo> list) {
        super.setList(list);
        this.currentData = list;
    }

    protected PagedList<Foo> getCurrentData() {
        if (this.currentData != null) {
            return this.currentData;
        }
        return (PagedList<Foo>) Collections.<Foo>emptyList();
    }

}
```
with Placeholders disabled:
```java
new PagedList.Config.Builder()
        .setPageSize(10)
        .setPrefetchDistance(50)
        .setEnablePlaceholders(false)
        .build())
```

On the other hand, there is [getCurrentList()](https://developer.android.com/reference/android/arch/paging/PagedListAdapter.html#getCurrentList()) in PagedListAdapter
```java
    @Nullable
    public PagedList<T> getCurrentList() {
        return this.mHelper.getCurrentList();
    }
```

And, before I make a pr:
Why there is nothing like this in PagingEpoxyController? 
Is it a bad practice for PagingEpoxyController?

elihart:
I don't think this is bad practice, I just didn't consider it. Seems valid, thanks for the idea!

One concern, since both a List and PagedList can be set, which should be returned? There could be a separate method for each.

elihart:
There could be `getPagedList` which returns the current list or null, and then `getCurrentList` which returns the current data being displayed - either the current paged list snapshot or the java list that was set ( or empty list if nothing is set). What do you think?

pcqpcq:
Yeap, it's very nice, I've make a pr here: #317

pcqpcq:
closed as #317 being merged.

PR #317, opened 2017-10-14
add support for get currently being displayed data

pcqpcq:
see also: #314 .

with two methods:

```java
public List<T> getCurrentList();

public PagedList<T> getPagedList()
```

### DIFF
--- a/epoxy-paging/src/main/java/com/airbnb/epoxy/paging/PagingEpoxyController.java
+++ b/epoxy-paging/src/main/java/com/airbnb/epoxy/paging/PagingEpoxyController.java
@@ -176,7 +176,7 @@ public abstract class PagingEpoxyController<T> extends EpoxyController {
 
     updatePagedListSnapshot();
   }
-  
+
   /**
      * Returns the list currently being displayed by the EpoxyController.
      *
@@ -185,7 +185,7 @@ public abstract class PagingEpoxyController<T> extends EpoxyController {
   public List<T> getCurrentList() {
     return this.list == null ? Collections.<T>emptyList() : this.list;
   }
-  
+
   /**
      * Returns the pagedList currently being displayed by the EpoxyController.
      *
--- a/epoxy-paging/src/main/java/com/airbnb/epoxy/paging/PagingEpoxyController.java
+++ b/epoxy-paging/src/main/java/com/airbnb/epoxy/paging/PagingEpoxyController.java
@@ -178,7 +178,7 @@ public abstract class PagingEpoxyController<T> extends EpoxyController {
   }
 
   /**
-     * Returns the list currently being displayed by the EpoxyController when you set data 
+     * Returns the list currently being displayed by the EpoxyController when you set data
      * with an Android PagedList.
      * <p>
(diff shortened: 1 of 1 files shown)


epoxy-paging/src/main/java/com/airbnb/epoxy/paging/PagingEpoxyController.java
```diff
@@ -177,6 +177,24 @@ public void setList(@Nullable PagedList<T> list) {
...
+    return this.list == null ? Collections.<T>emptyList() : this.list;
+  }
+
+  /**
+     * Returns the pagedList currently being displayed by the EpoxyController.
+     *
+     * @return The pagedList currently being displayed.
+     */
+  public PagedList<T> getPagedList() {
+    return this.pagedList == null ? (PagedList<T>) Collections.<T>emptyList() : this.pagedList;
```
elihart: You can't just cast the list to a PagedList, that will crash. I think we need to return null in this case, it isn't possible to create our own empty paged list

epoxy-paging/src/main/java/com/airbnb/epoxy/paging/PagingEpoxyController.java
```diff
@@ -177,6 +177,24 @@ public void setList(@Nullable PagedList<T> list) {
     updatePagedListSnapshot();
   }
 
+  /**
+     * Returns the list currently being displayed by the EpoxyController.
```
elihart: can you note that this is either a snapshot of the paged list or the list set with `setList`?

epoxy-paging/src/main/java/com/airbnb/epoxy/paging/PagingEpoxyController.java
```diff
@@ -177,6 +177,24 @@ public void setList(@Nullable PagedList<T> list) {
...
+     *
+     * @return The list currently being displayed.
+     */
+  public List<T> getCurrentList() {
+    return this.list == null ? Collections.<T>emptyList() : this.list;
```
elihart: `list` is never null, you can return it directly

pcqpcq:
It's my negligence, I've update the code.

Sorry for reply until 2 days later.

epoxy-paging/src/main/java/com/airbnb/epoxy/paging/PagingEpoxyController.java
pcqpcq: yes, null check here is unnecessary.

epoxy-paging/src/main/java/com/airbnb/epoxy/paging/PagingEpoxyController.java
pcqpcq: null check here is unnecessary too.

epoxy-paging/src/main/java/com/airbnb/epoxy/paging/PagingEpoxyController.java
pcqpcq: Oh! you are right, this is inaccurate

elihart:
Great, thanks!

PR #321, opened 2017-10-16
Generate model interfaces for interfaces on view

elihart:
This is for custom views with the `@ModelView` annotation. If the view implements an interface then the generated model will also implement an interface representing the view's interface.

A separate interface is generated to represent the view's interface, this allows the model interface to include base model properties, like id, in addition to the interface methods. This also allows the generated interface to be flexible and not necessarily include every method in the view interface.

For example, say a view implements this
```
interface Clickable {
  void setClickable(boolean clickable);
  boolean isClickable();
}
```

If the `setClickable` implementation is annotated with `@ModelProp` then the generated model for the view will implement a `ClickablModel_` interface. This interface will include just the setter, not the getter.
```
interface ClickableModel_ {
 ClickableModel_ clickable(boolean clickable);
 ClickableModel_ id(long id);
 // other standard model methods
}
```

Then you can cast different models to `ClickableModel_` and access the shared props easily.

The generated interface does not necessarily mirror the original view interface. Instead, Epoxy looks at all generated models with that view type, gets the subset of props that they all share, and makes the generated interface implement methods for those props. This makes the generated interface flexible, and is easier to implement.

The generated interface is in the same package as the view interface.

The generated interface is the name of the original interface with `Model_` as a suffix.

Some caveats:
- This does not support views inheriting interfaces
- Does not support @ModelViews in other modules with the same interface
- Does not support abstract views partially implementing an interface

This was fairly difficult to implement and is complex, so I'm just supporting the basic case for now. I see a way forward to support the caveats, but might not get to it for a while unless there is demand.

I would also like to support doing something like this for models that inherit props from other models (ie a base model). That will have to be a slightly different approach, but similar ideas can be used.

### DIFF
--- a/epoxy-integrationtest/src/main/java/com/airbnb/epoxy/integrationtest/ViewWithInterface.kt
+++ b/epoxy-integrationtest/src/main/java/com/airbnb/epoxy/integrationtest/ViewWithInterface.kt
@@ -1,0 +1,64 @@
+package com.airbnb.epoxy.integrationtest
+
+import android.content.*
+import android.support.annotation.*
+import android.view.*
+import com.airbnb.epoxy.*
+import com.airbnb.epoxy.ModelView.Size
+
+@ModelView(autoLayout = Size.WRAP_WIDTH_MATCH_HEIGHT)
+class ViewWithInterface(context: Context) : View(context), InterfaceForView, InterfaceForView2 {
+
+    @ModelProp
+    override fun setText(title: CharSequence?) {
+
+    }
+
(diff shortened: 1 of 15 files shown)


elihart:
I'm going to merge for now so I can prepare and test a release, but if anybody has feedback please still leave it and I can make updates.

elihart:
updated the wiki https://github.com/airbnb/epoxy/wiki/Generating-Models-from-View-Annotations#view-interfaces

PR #326, opened 2017-10-19
Refactor for package structure.

niccorder:
This is a potential refactor to allow for a little bit of a cleaner package structure. This moves away from using explicit access modifiers in lieu of the android support annotation's `@RestrictTo(...)`.

This is not necessary, but it would allow for a tidier package structure.


epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModel.java
```diff
@@ -404,7 +404,7 @@ public void onInterceptorsFinished(EpoxyController controller) {
     }
   }
 
-  boolean isDebugValidationEnabled() {
+  public final boolean isDebugValidationEnabled() {
```
elihart: would prefer not to have these public

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyControllerAdapter.java
```diff
@@ -25,7 +27,7 @@ protected void onExceptionSwallowed(RuntimeException exception) {
   }
 
   @Override
-  List<EpoxyModel<?>> getCurrentModels() {
+  public List<EpoxyModel<?>> getCurrentModels() {
```
elihart: this can't be public

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyAdapter.java
```diff
@@ -29,7 +31,7 @@
   private DiffHelper diffHelper;
 
   @Override
-  List<EpoxyModel<?>> getCurrentModels() {
+  public List<EpoxyModel<?>> getCurrentModels() {
```
elihart: this can't be public

epoxy-adapter/src/main/java/com/airbnb/epoxy/BaseEpoxyAdapter.java
```diff
@@ -66,7 +72,11 @@ public int getItemCount() {
   }
 
   /** Return the models currently being used by the adapter to populate the recyclerview. */
-  abstract List<EpoxyModel<?>> getCurrentModels();
+  public abstract List<EpoxyModel<?>> getCurrentModels();
+
+  public void onModelAdded(EpoxyModel<?> model) {
```
elihart: this should be restricted

PR #343, opened 2017-11-06
Enable Enable @ModelProp @TextProp and @CallbackProp with all options on field attribute

juliafu1:
Enable Enable @ModelProp @TextProp and @CallbackProp with all options on field attribute

### DIFF
--- a/epoxy-processor/src/main/java/com/airbnb/epoxy/ModelViewInfo.kt
+++ b/epoxy-processor/src/main/java/com/airbnb/epoxy/ModelViewInfo.kt
@@ -60,7 +60,7 @@ internal class ModelViewInfo(
                 .filter { interfaceElement ->
                     // Only include the interface if the view has one of the interface methods annotated with a prop annotation
                     methodsOnView.any { viewMethod ->
-                        viewMethod.hasAnyAnnotation(ModelViewProcessor.modelPropAnnotations)
+                        viewMethod.hasAnyAnnotation(ModelViewProcessor.modelPropAnnotations.map { it.java })
                         && interfaceElement.executableElements().any { interfaceMethod ->
                             // To keep this simple we only compare name and ignore parameters, should be close enough
                              viewMethod.simpleName.toString() == interfaceMethod.simpleName.toString()
@@ -153,14 +153,14 @@ internal class ModelViewInfo(
         return ClassName.get(packageName, className)
     }
 
-    fun addProp(propMethod: ExecutableElement) {
-        addAttribute(ViewAttributeInfo(this, propMethod, typeUtils, elements, errorLogger,
+    fun addProp(prop: Element) {
+        addAttribute(ViewAttributeInfo(this, prop, typeUtils, elements, errorLogger,
                                        resourceProcessor))
     }
 
-    fun addPropIfNotExists(propMethod: ExecutableElement) {
+    fun addPropIfNotExists(prop: Element) {
         addAttributeIfNotExists(
-                ViewAttributeInfo(this, propMethod, typeUtils, elements, errorLogger,
+                ViewAttributeInfo(this, prop, typeUtils, elements, errorLogger,
                                   resourceProcessor))
     }
 
--- a/epoxy-processor/src/main/java/com/airbnb/epoxy/ModelViewProcessor.kt
+++ b/epoxy-processor/src/main/java/com/airbnb/epoxy/ModelViewProcessor.kt
@@ -105,19 +105,19 @@ internal class ModelViewProcessor(
     private fun processSetterAnnotations(roundEnv: RoundEnvironment) {
 
         for (propAnnotation in modelPropAnnotations) {
-            for (propMethod in roundEnv.getElementsAnnotatedWith(propAnnotation)) {
-                if (!validatePropElement(propMethod, propAnnotation)) {
+            for (prop in roundEnv.getElementsAnnotatedWith(propAnnotation.java)) {
+                if (!validatePropElement(prop, propAnnotation.java)) {
                     continue
                 }
 
-                val info = getModelInfoForMethodElement(propMethod)
+                val info = getModelInfoForPropElement(prop)
                 if (info == null) {
                     errorLogger.logError(
-                            "${propAnnotation.simpleName} annotation can only be used in classes annotated with ${ModelView::class.java.simpleName} (${propMethod.enclosingElement.simpleName}#${propMethod.simpleName})")
+                            "${propAnnotation.simpleName} annotation can only be used in classes annotated with ${ModelView::class.java.simpleName} (${prop.enclosingElement.simpleName}#${prop.simpleName})")
                     continue
                 }
 
-                info.addProp(propMethod as ExecutableElement)
+                info.addProp(prop)
             }
         }
     }
@@ -138,7 +138,7 @@ internal class ModelViewProcessor(
                 if (groupKey.isEmpty()) {
                     // Default to using the method name as the group name, so method overloads are grouped
                     // together by default
-                    groupKey = setterInfo.viewSetterMethodName
+                    groupKey = setterInfo.viewAttributeName
                 } else {
                     customGroups.add(groupKey)
                 }
@@ -157,7 +157,7 @@ internal class ModelViewProcessor(
                     if (it.size == 1) {
                         val attribute = it[0] as ViewAttributeInfo
                         errorLogger.logError(
-                                "Only one setter was included in the custom group '$customGroup' at ${viewInfo.viewElement.simpleName}#${attribute.viewSetterMethodName}. Groups should have at least 2 setters.")
+                                "Only one setter was included in the custom group '$customGroup' at ${viewInfo.viewElement.simpleName}#${attribute.viewAttributeName}. Groups should have at least 2 setters.")
                     }
                 }
             }
@@ -174,10 +174,34 @@ internal class ModelViewProcessor(
     }
 
     private fun validatePropElement(
-            methodElement: Element,
+            prop: Element,
             propAnnotation: Class<out Annotation>
-    ): Boolean =
-            validateExecutableElement(methodElement, propAnnotation, 1)
+    ): Boolean {
+        return when (prop) {
+            is ExecutableElement -> validateExecutableElement(prop, propAnnotation, 1)
+            is VariableElement -> validateVariableElement(prop, propAnnotation)
+            else -> {
+                errorLogger.logError(
+                        "%s annotations can only be on a method or a field(element: %s)",
+                        propAnnotation,
+                        prop.simpleName)
+                return false
+            }
+        }
+    }
+
+    private fun validateVariableElement(field: Element, annotationClass: Class<*>): Boolean {
+        var isValidField =  field.kind == ElementKind.FIELD
+                && !field.modifiers.contains(PRIVATE)
+                && !field.modifiers.contains(STATIC)
+
+        if (!isValidField) {
+            errorLogger.logError(
+                    "Field annotated with %s must not be static or private (field: %s)",
+                    annotationClass, field)
+        }
+        return isValidField
+    }
 
     private fun validateExecutableElement(
             element: Element,
@@ -215,7 +239,7 @@ internal class ModelViewProcessor(
                 continue
             }
 
-            val info = getModelInfoForMethodElement(recycleMethod)
+            val info = getModelInfoForPropElement(recycleMethod)
             if (info == null) {
                 errorLogger.logError("%s annotation can only be used in classes annotated with %s",
                                      OnViewRecycled::class.java, ModelView::class.java)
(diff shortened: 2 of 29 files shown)


epoxy-processor/src/main/java/com/airbnb/epoxy/ViewAttributeInfo.kt
```diff
@@ -9,31 +9,46 @@ import java.util.*
...
+import javax.lang.model.element.ElementKind
+import com.sun.tools.javac.util.BaseFileManager.getKind
+import javax.lang.model.element.ExecutableElement
+
+
 
 private val NOT_NULL_ANNOTATION_SPEC = AnnotationSpec.builder(NotNull::class.java).build()
 private val NON_NULL_ANNOTATION_SPEC = AnnotationSpec.builder(NonNull::class.java).build()
 
+internal enum class ViewAttributeType {
```
elihart: can you make this a sealed class instead?

epoxy-processor/src/main/java/com/airbnb/epoxy/ViewAttributeInfo.kt
```diff
@@ -9,31 +9,46 @@ import java.util.*
...
         errorLogger: ErrorLogger,
         resourceProcessor: ResourceProcessor
 ) : AttributeInfo() {
     val propName: String
-    val viewSetterMethodName: String
+    val viewSetterPropName: String
     val resetWithNull: Boolean
     val generateStringOverloads: Boolean
     var constantFieldNameForDefaultValue: String? = null
+    var viewAttributeTypeName: ViewAttributeType? = null
```
elihart: this should be a `val`

epoxy-processor/src/main/java/com/airbnb/epoxy/ViewAttributeInfo.kt
```diff
@@ -333,7 +379,7 @@ internal class ViewAttributeInfo(
             }
         }
 
-        builder.add("\n\n@see \$T#\$L(\$T)", viewElement.asType(), viewSetterMethodName, typeMirror)
+        builder.add("\n\n@see \$T#\$L(\$T)", viewElement.asType(), viewSetterPropName, typeMirror)
```
elihart: I don't think the generated javadoc will be right for fields

epoxy-processor/src/main/java/com/airbnb/epoxy/ViewAttributeInfo.kt
```diff
@@ -272,6 +314,10 @@ internal class ViewAttributeInfo(
             markedNullable: Boolean,
             typeName: TypeName
     ) {
+        //TODO juliafu support annotations for field
+        if (viewAttributeTypeName == ViewAttributeType.Field) {
```
elihart: I think this should just work already with no changes

epoxy-processor/src/main/java/com/airbnb/epoxy/ViewAttributeInfo.kt
```diff
@@ -226,27 +261,34 @@ internal class ViewAttributeInfo(
             errorLogger: ErrorLogger,
             options: Set<Option>,
             types: Types,
-            elements: Elements
+            elements: Elements,
+            setterElementOnView: Element
     ) {
+
+        if (!options.isEmpty() &&  viewAttributeTypeName == ViewAttributeType.Field) {
```
elihart: a field should be able to support options, I don't think it needs to be different there

epoxy-processor/src/main/java/com/airbnb/epoxy/ViewAttributeInfo.kt
```diff
@@ -128,6 +152,17 @@ internal class ViewAttributeInfo(
                 super.isRequired()
             }
 
+    private fun setViewAttributeType(
+            element: Element,
+            errorLogger: ErrorLogger
+    ) {
+        when {
```
elihart: it's cleaner to return a value than to set a field. the field should be a val to be safer

epoxy-processor/src/main/java/com/airbnb/epoxy/ViewAttributeInfo.kt
```diff
@@ -9,31 +9,46 @@ import java.util.*
...
+    var viewAttributeTypeName: ViewAttributeType? = null
 
     init {
         val propAnnotation = setterElementOnView.getAnnotation(ModelProp::class.java)
         val textAnnotation = setterElementOnView.getAnnotation(TextProp::class.java)
         val callbackAnnotation = setterElementOnView.getAnnotation(CallbackProp::class.java)
 
         val options = HashSet<Option>()
-        val param = setterElementOnView.parameters[0]
+        val param = if (setterElementOnView is ExecutableElement) setterElementOnView.parameters[0]
```
elihart: `setterElementOnView` should be renamed, it doesn't make sense for fields

epoxy-processor/src/main/java/com/airbnb/epoxy/ViewAttributeInfo.kt
```diff
@@ -43,6 +58,11 @@ internal class ViewAttributeInfo(
             options.addAll(Arrays.asList(*propAnnotation.options))
             options.addAll(Arrays.asList(*propAnnotation.value))
         } else if (textAnnotation != null) {
+            //TODO juliafu support textProp for field
+            if (viewAttributeTypeName == ViewAttributeType.Field) {
```
elihart: we shouldn't have to do anything special here to support it

epoxy-processor/src/main/java/com/airbnb/epoxy/ViewAttributeInfo.kt
```diff
@@ -52,6 +72,10 @@ internal class ViewAttributeInfo(
             }
             options.add(Option.GenerateStringOverloads)
         } else if (callbackAnnotation != null) {
+            if (viewAttributeTypeName == ViewAttributeType.Field) {
```
elihart: same here

epoxy-processor/src/main/java/com/airbnb/epoxy/ViewAttributeInfo.kt
```diff
@@ -75,8 +99,8 @@ internal class ViewAttributeInfo(
         modelName = modelInfo.generatedName.simpleName()
         modelPackageName = modelInfo.generatedClassName.packageName()
 
-        this.viewSetterMethodName = setterElementOnView.simpleName.toString()
-        propName = removeSetPrefix(viewSetterMethodName)
+        this.viewSetterPropName = setterElementOnView.simpleName.toString()
```
elihart: I don't know if the name `viewSetterPropName` makes sense for fields

juliafu1:
@elihart I updated the code and addressed all comments
The testing is here: https://docs.google.com/document/d/1sT-AUi-Iiv3qI6dTfzbW3szRteq9qwaY-m1L7vnSY58/edit#bookmark=id.t7rle9y2y6xm

There are two things not fully working:
1) default value:
@ModelProp (defaultValue = "DEFAULT_STRING") String value;

The build stuck at  Task :epoxy-sample:compileDebugJavaWithJavac. I do not know why. There is no error message

2) annotation
@ModelProp @Nullable string value generates the following:

@Nullable
public TestViewModel_ value(final OnModelClickListener<TestViewModel_, TestView> value) {
  .....
}

I need to figure out how to make the @Nullable annotation on parameter. 

PTAL

epoxy-processor/src/main/java/com/airbnb/epoxy/ModelViewInfo.kt
```diff
@@ -153,11 +153,12 @@ internal class ModelViewInfo(
...
     }
 
-    fun addProp(propMethod: ExecutableElement) {
-        addAttribute(ViewAttributeInfo(this, propMethod, typeUtils, elements, errorLogger,
+    fun addProp(prop: Element) {
+        addAttribute(ViewAttributeInfo(this, prop, typeUtils, elements, errorLogger,
                                        resourceProcessor))
     }
 
+    //TODO juliafu handle the super class case
```
elihart: can you add more details to this comment? not sure what needs to be done exactly

epoxy-processor/src/main/java/com/airbnb/epoxy/ModelViewProcessor.kt
```diff
@@ -356,12 +380,14 @@ internal class ModelViewProcessor(
...
     fun hasModelsToWrite() = styleableModelsToWrite.isNotEmpty()
 
-    private fun getModelInfoForMethodElement(element: Element): ModelViewInfo? =
+    private fun getModelInfoForPropElement(element: Element): ModelViewInfo? =
             element.enclosingElement?.let { modelClassMap[it] }
 
     companion object {
         val modelPropAnnotations = listOf(ModelProp::class, TextProp::class,
                                           CallbackProp::class).map { it.java }
+        val modelPropAnnotationsNames = listOf(ModelProp::class, TextProp::class,
```
elihart: it would be nice not to duplicate the list of prop classes

epoxy-processor/src/main/java/com/airbnb/epoxy/ModelViewWriter.kt
```diff
@@ -214,9 +217,14 @@ internal class ModelViewWriter(
             boundObjectParam: ParameterSpec,
             viewAttribute: ViewAttributeInfo
     ): CodeBlock {
-        return CodeBlock.of("\$L.\$L(\$L);\n", boundObjectParam.name,
-                            viewAttribute.viewSetterMethodName,
-                            getValueToSetOnView(viewAttribute, boundObjectParam))
+        val expression = if (viewAttribute.viewAttributeTypeName == ViewAttributeType.Field)
+            "\$L.\$L=\$L;\n" else "\$L.\$L(\$L);\n"
```
elihart: I think we can do `CodeBlock.builder().addStatement()...` to avoid having to do `;\n` manually.

elihart: this is mostly duplicated with the unbind code, the only difference is that the unbind code sets null. can you think of a way to DRY it out?

epoxy-processor/src/main/java/com/airbnb/epoxy/ViewAttributeInfo.kt
```diff
@@ -128,6 +137,21 @@ internal class ViewAttributeInfo(
                 super.isRequired()
             }
 
+    private fun getViewAttributeType(
+            element: Element,
+            errorLogger: ErrorLogger
+    ): ViewAttributeType? {
+        return when {
+            element is ExecutableElement -> ViewAttributeType.Method
+            element.kind === ElementKind.FIELD -> ViewAttributeType.Field
```
elihart: don't use triple equals here

elihart: can you do `when(element.kind) is FIELD ... is METHOD...`?

epoxy-processor/src/main/java/com/airbnb/epoxy/ViewAttributeInfo.kt
```diff
@@ -274,7 +298,10 @@ internal class ViewAttributeInfo(
     ) {
         for (annotationMirror in paramElement.annotationMirrors) {
             val annotationElement = types.asElement(annotationMirror.annotationType)
-            val builder = AnnotationSpec.builder(ClassName.get(annotationElement as TypeElement))
+            if (ModelViewProcessor.modelPropAnnotationsNames.contains(ClassName.get(annotationElement as TypeElement))) {
+                continue
+            }
+            val builder = AnnotationSpec.builder(ClassName.get(annotationElement))
```
elihart: save `ClassName.get(annotationElement)` to a val since it is done twice?

epoxy-processor/src/main/java/com/airbnb/epoxy/ViewAttributeInfo.kt
```diff
@@ -274,7 +298,10 @@ internal class ViewAttributeInfo(
     ) {
         for (annotationMirror in paramElement.annotationMirrors) {
             val annotationElement = types.asElement(annotationMirror.annotationType)
-            val builder = AnnotationSpec.builder(ClassName.get(annotationElement as TypeElement))
+            if (ModelViewProcessor.modelPropAnnotationsNames.contains(ClassName.get(annotationElement as TypeElement))) {
```
elihart: can do `if (annotationClass in ModelViewProcessor.modelPropAnnotationsNames) continue`

epoxy-processor/src/main/java/com/airbnb/epoxy/ViewAttributeInfo.kt
juliafu1: It couldn't use 'is'.  The compiler tells: "'is' over enum entry is not allowed, use comparison instead". I can use '==' instead of '===' but it actually doesn't make a difference for enum

juliafu1:
@elihart all comments addressed. I also tested the parent view case (if the parent view has an attribute, the generated child view model shall have that attribute as well). The test is added in https://docs.google.com/document/d/1sT-AUi-Iiv3qI6dTfzbW3szRteq9qwaY-m1L7vnSY58/edit#bookmark=id.t7rle9y2y6xm. Thank you.

juliafu1:
@elihart Tests added. PTAL. Thank you

elihart:
Great! tests look pretty exhaustive - I'll try to give this one final lookover soon and then merge 👍

juliafu1:
Thank you!

Issue #379, opened 2018-01-15
Issue / Question with OnModelClicked callbacks and @CallbackProp


tikimcfee:
Hey again!

I hit a minor snag in the combination above; using an `OnModelClicked` listener and using `@CallbackProp`. The models I'm using are of the auto-generated variety when using custom annotated views (still one of my favorite features). Long story short, I was trying to..

1. Reduce the number of click listeners by setting the instance of a listener to a single object
2. Switch, based on some of the callback model's properties, what action to perform.

It looks like when rebuilding/binding models, there's an equality check that says something along the lines of, "Oh, well this _old_ model's listener wasn't null, and this one isn't either. Nothing so set, I'll skip that property set call." Problem there though is the the callback into the view isn't set, and I don't have a chance to call `setClickListener` on the view again. It ends up that the _old_ listener remains registered with the view, and when it's clicked, it ends up calling me back with the wrong model pointer (blah!)

I got a little turned around trying to figure out the best way to handle this, and settled on the simple approach of going _back_ to per-model listeners with `IgnoreRequiresHashCode`.

I think the fix for this one is something like `if the listener you're setting is a model click listener, and the old model already had one, reset its model to this new instance.`

On the other-other hand, I could be completely and totally misusing the click listener Callback prop thing. Feel free to pull out the wet noodles for a flogging if necessary.

As always, cheers and thanks for your time!

elihart:
I think this is spot on, great catch! Much appreciate you reporting this with such great detail, must have been a frustrating bug to track down!

Your idea for a fix would work, but adding more exceptions to the generated code makes the annotation processor complex and difficult to maintain, so I think we can solve this with a slightly different approach like this - https://github.com/airbnb/epoxy/pull/381

I'll think about it a little longer and get some tests around this, but I'll try to release the fix soon.

Thanks again, and sorry for the issue!

P.S. A `WrappedEpoxyModelClickListener` instance is created for every model click listener when a new model is bound. This is a bit wasteful, this doesn't need to get created until the model is first bound. It would be a nice improvement to reduce object allocation when creating models by deferring this.

tikimcfee:
Man, I took a look at that commit - it does look a bit complicated ;) 

I appreciate your looking into it, and I'm glad you've got a fix in mind for it. No big rush on this one since the workaround really isn't bad at the moment (the use case I have at the moment only contains a few rows). Happy to test out as you need!

[...]

PR #381, opened 2018-01-16
Update wrapped click listener when model is rebound

elihart:
Addresses https://github.com/airbnb/epoxy/issues/379

Since the EpoxyViewHolder is aware of when a view is rebound it can make sure that previously bound click listeners have their model references updated.

This also adds a small change to remove click listener references to the bound view when the model is unbound.


elihart:
Closing in favor of https://github.com/airbnb/epoxy/pull/393

Issue #396, opened 2018-01-31
Cyclic recycler view


abyrnes:
I have a horizontal recycler view that uses `PagerSnapHelper` so it acts as a view pager. The pager needs to loop back around to the first item when it reaches the end. I am already using epoxy as it was an easy option since the recycler view in question is nested inside another recycler view. I was just wondering if there's an easy way to achieve this cyclic behavior with epoxy.

elihart:
See https://github.com/airbnb/epoxy/issues/375

elihart:
@abyrnes I put this together to come up with a better way to support cyclic scrolling https://github.com/airbnb/epoxy/pull/401

let me know what you think

abyrnes:
I took what you wrote and used it and it works great! The only issue I see, if you can call it an issue, is it allows cyclic scrolling even when there's a single item in the adapter, which I imagine is usually undesired. Do you think it would be reasonable to provide a hook to enable/disable this behavior?

abyrnes:
```java
@Override
public int getItemCount() {
  return getActualItemCount() == 1 ? 1 : Integer.MAX_VALUE;
}
```
Solves the issue
So it should be trivial to toss in a flag that lets the user determine the behavior

elihart:
@abyrnes Thanks for trying it and for the feedback. Glad it works!

You're right that scrolling with a single item is probably usually undesirable, but I want to make sure I leave it flexible for anyone that might want that.

I'll add something to provide control

abyrnes:
@elihart Sounds great! The only other thing I can think of that it doesn't support is being able to scroll to the left when at position 0

elihart:
@abyrnes that's a good point. I don't think there is a good way for the adapter to get around that, but what you can do is force the recyclerview to scroll to (Integer.max_value / actual count / 2) so it starts on the first item, in the middle of the list.

I also think that data set changes won't work well with this approach - the differ will only report that the original item range changed, and recyclerview won't update views past that.

abyrnes:
> I also think that data set changes won't work well with this approach - the differ will only report that the original item range changed, and recyclerview won't update views past that.

Yea that's pretty big. Fortunately for me my particular use case doesn't require the data set to change.

elihart:
Yeah, I don't feel too comfortable adding this with a quirk that large.

Instead of making this an official part of the library I can set up a wiki page, have a gist for this code, list the gotchas and work arounds. And people can do whatever they want with that

[...]

Issue #375, opened 2018-01-09
Carousel with circular scrolling


sauyee333:
Hi,
I have an image list that need to be able to scroll both ways. The original implementation is using viewpager. After I convert to using carousel, I couldn't find a way to scroll both left and right (loop effect). Please advise. Thanks.

elihart:
RecyclerView itself doesn't readily support circular infinite scrolling. However, you can approximate it by simply adding your items in a loop many times.

That would be less efficient - it could be possible for Epoxy to support circular scrolling with a configuration option to be more efficient, but I haven't looked at that option too closely yet.

elihart:
@sauyee333 I put this together to come up with a better way to support circular scrolling https://github.com/airbnb/epoxy/pull/401

let me know what you think

PR #401, opened 2018-02-14
Cyclic wrapping adapter

elihart:
Possible way to address https://github.com/airbnb/epoxy/issues/396 and https://github.com/airbnb/epoxy/issues/375

I haven't tested this yet, but it seems like something like this should work.

Anybody see any possible issues?

### DIFF
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/CyclicEpoxyAdapter.java
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/CyclicEpoxyAdapter.java
@@ -1,22 +1,144 @@
 package com.airbnb.epoxy;
 
 import android.support.v7.widget.RecyclerView;
+import android.support.v7.widget.RecyclerView.Adapter;
+import android.support.v7.widget.RecyclerView.AdapterDataObserver;
 import android.support.v7.widget.RecyclerView.ViewHolder;
 import android.view.ViewGroup;
 
+import java.util.List;
+
 public class CyclicEpoxyAdapter extends RecyclerView.Adapter {
+
+  private final Adapter adapter;
+
+  public CyclicEpoxyAdapter(RecyclerView.Adapter adapter) {
+    this.adapter = adapter;
+
+    super.setHasStableIds(adapter.hasStableIds());
+    super.registerAdapterDataObserver(observerDelegate);
+  }
+
+  public int getActualItemCount() {
+    return adapter.getItemCount();
+  }
+
+  /**
+   * Do Not Use!!!
+   * <p>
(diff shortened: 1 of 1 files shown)


mradzinski:
Checked the code and it looks fine. I don't see any possible issues and although I haven't tested the branch it looks like it should work as it is. I think I'll be able to test it this weekend, if I find anything weird I'll let you know @elihart

elihart:
Great!

mradzinski:
@elihart I tested this. So far so good, works as expected and nothing weird arose from my tests (didn't exactly conducted extensive ones, but tested regular use cases and they all seemed fine). Let me know if I can be of any other help.

elihart:
@mradzinski Thanks a bunch for testing!

I have decided not to merge this and instead provide it as a gist with instructions in the wiki: https://github.com/airbnb/epoxy/wiki/Cyclic-Adapters

I don't feel good about supporting this officially when item changes won't work properly. The wiki should provide enough information for people to customize this to best suit their use cases.

mradzinski:
> I don't feel good about supporting this officially when item changes won't work properly

And that @elihart is one of the tests I didn't conduct 😞, shame on me. But yes, I agree that the wiki should provide enough info. Also, I gave it a thought last weekend while I was testing this and sometimes keeping libraries doing one thing but doing that one thing right is the best approach one can take. I consider Epoxy to be one of those libraries that don't certainly need to have lots of preconfigured/premade use cases since it's extensible by nature and the things it does it does them right AF. 

Us as developers should give a thought or two about how to accomplish certain stuff. As an example, Epoxy doesn't have a `TypedEpoxyController` which allows for delayed builds, however that's easily accomplished by creating your own Typed Controller that allows for delayed builds just by extending from `EpoxyController` and checking the other Typed Controllers code.

Cheers and thanks as always for your hard work!

bkulaksiz:
Hello @elihart 

If it is possible with EpoxyRecyclerView, I would like to make an endless automatic scrolling carousel, but I do not know how to do with epoxy. Is there are any suggestion to make it? I could not find any way to implement it.

Best regards.

bkulaksiz:
Any suggestion?

jrocharodrigues:
Hi,
Any guidance on how to use the cycle adapter with a carousel?

I've tried to do it by extending the Carousel class, overriding the setAdapter model, so it wraps the initial adapter inside the Cycle Adapter in the gist, but it doesn't work and crashes.
``` 
override fun setAdapter(adapter: Adapter<*>?) {
        super.setAdapter(CyclicAdapter(adapter))
    }
```

I've tried to fix the crashes but i was not able to make it work...





Thanks in advance

PR #406, opened 2018-03-11
make isBuildingModels public

anasanasanas:
make isBuildingModels public instead of package access to create custom TypedEpoxyController.

when creating custom Typed5EpoxyController that should accept five objects I ran into a compiler error
"isBuildingModels" Cannot be accessed from outside package

### DIFF
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyController.java
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyController.java
@@ -425,7 +425,7 @@ public abstract class EpoxyController {
     stagedModel = null;
   }
 
-  boolean isBuildingModels() {
+  public boolean isBuildingModels() {
     return modelsBeingBuilt != null;
   }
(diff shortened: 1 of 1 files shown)


elihart:
I would prefer to make this `protected` instead of `public`.

In general though I would advise against `Typed5EpoxyController`. With that many params it seems much cleaner and safer to have a kotlin data class to represent your data state

elihart:
@anasanasanas thanks for updating, merged.

The problem with that many params is the typed controller does not give helpful param names in `setData`, so all you see is `setData(data1, data2, data3)`.

if you make a wrapper class in kotlin
```
class ControllerState(val thingName: MyThing ...)

typedEpoxyController.setData(ControllerState(thingName = ...)
```

you have a lot more visibility into what the data actually represents, you can add kdoc for the parameters, you can store the data altogether and use `copy` to modify it, etc

anasanasanas:
I get it now.
many thanks @elihart

Issue #412, opened 2018-04-01
lifecycleOwner  property with databinding auto generated models


SBD580:
Hi,

Thanks for this great library!

Having an issue after upgrading to Android Studio 3.1 which introduce the Android Architecture Components. If my layout includes another layout the `setLifecycleOwner` method is being overridden in the generated binding class and then being treated as an attribute. This is causing a build time exception while the generated (*Model_) class is trying to use `BR.lifecycleOwner` which doesn't really exists.

elihart:
Thanks for reporting. Can you provide an example layout with the issue, as well as the generated model code that is wrong?

SBD580:
Sure, here you go:

activity_main.xml
```
<?xml version="1.0" encoding="utf-8"?>
<layout xmlns:android="http://schemas.android.com/apk/res/android">
    <data />
    <android.support.constraint.ConstraintLayout
        android:layout_width="match_parent"
        android:layout_height="match_parent">
        <include layout="@layout/included_layout" />
    </android.support.constraint.ConstraintLayout>
</layout>
```

included_layout.xml
```
<?xml version="1.0" encoding="utf-8"?>
<layout xmlns:android="http://schemas.android.com/apk/res/android">

    <data />

    <android.support.constraint.ConstraintLayout
        android:layout_width="match_parent"
        android:layout_height="match_parent" />
</layout>
```

build.gradle
```
apply plugin: 'com.android.application'

android {
    compileSdkVersion 27
    defaultConfig {
        applicationId "com.sbd580.eproxyissue412"
        minSdkVersion 21
        targetSdkVersion 27
        versionCode 1
        versionName "1.0"
        testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
    }
    dataBinding {
        enabled true
    }
    buildTypes {
        release {
            minifyEnabled false
            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
        }
    }
}

ext.set("supportVersion", '27.1.0')
//
configurations.all {
    resolutionStrategy {
        eachDependency { details ->
            // Force all of the primary support libraries to use the same version.
            if (details.requested.group == 'com.android.support'
                    && details.requested.name != 'multidex'
                    && details.requested.name != 'multidex-instrumentation') {
                details.useVersion supportVersion
            }
        }
    }
}

dependencies {
    implementation fileTree(dir: 'libs', include: ['*.jar'])
[...]

PR #413, opened 2018-04-02
put setLifecycleOwner in blacklist

beta1139:
Hi,

Thanks for great library.

I fixed about #412.


elihart:
Have you tested this? I don't know that this will work

PR #415, opened 2018-04-06
blacklist lifecycleOwner from generated models

### LINKED ISSUE #412: lifecycleOwner  property with databinding auto generated models

SBD580:
Fixes #412


elihart:
Thanks for putting this up. Looks like it should work, but it would be cleaner to abstract the change only in `DataBindingModelInfo`. I've merged this fix instead https://github.com/airbnb/epoxy/pull/418

PR #417, opened 2018-04-07
remove lifecycleOwner setter from auto generated data binding models

### LINKED ISSUE #412: lifecycleOwner  property with databinding auto generated models

lizhuogui:
fix #412


elihart:
Thanks for putting this up. This is a good approach, but we had already started on a similar change that did some other clean up at the same time, and I've merged this fix instead https://github.com/airbnb/epoxy/pull/418

thanks though!

PR #418, opened 2018-04-08
Blacklist setLifecycleOwner method in databinding field parsing, fixes #412

### LINKED ISSUE #412: lifecycleOwner  property with databinding auto generated models

ngsilverman:
ViewDataBinding subclasses can now override the setLifecycleOwner method causing it to be erroneously detected as a model field. This fix introduces a field blacklist to ignore this method and possibly others in the future.

To reproduce the issue and create a test for it I had to update the Android plugin version which forced me to also update Gradle, the target SDK, and the paging library. I removed the buildToolsVersion declarations to remove a warning (the build tools version is now provided by the Android plugin from what I understand).

### DIFF
--- a/epoxy-pagingsample/src/main/java/com/airbnb/epoxy/pagingsample/PagingSampleActivity.kt
+++ b/epoxy-pagingsample/src/main/java/com/airbnb/epoxy/pagingsample/PagingSampleActivity.kt
@@ -51,8 +51,8 @@ class PagingSampleActivity : AppCompatActivity() {
                             setPrefetchDistance(50)
                             build()
                         }).run {
-                    setMainThreadExecutor(UiThreadExecutor)
-                    setBackgroundThreadExecutor(AsyncTask.THREAD_POOL_EXECUTOR)
+                    setNotifyExecutor(UiThreadExecutor)
+                    setFetchExecutor(AsyncTask.THREAD_POOL_EXECUTOR)
                     build()
                 }
             }
--- a/epoxy-pagingsample/src/main/java/com/airbnb/epoxy/pagingsample/PagingSampleActivity.kt
+++ b/epoxy-pagingsample/src/main/java/com/airbnb/epoxy/pagingsample/PagingSampleActivity.kt
@@ -8,9 +8,9 @@ import android.os.Bundle
 import android.os.Handler
 import android.os.Looper
 import android.support.v7.app.AppCompatActivity
+import android.support.v7.widget.AppCompatTextView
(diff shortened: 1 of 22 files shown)


epoxy-pagingsample/src/main/java/com/airbnb/epoxy/pagingsample/PagingSampleActivity.kt
```diff
@@ -51,8 +51,8 @@ class PagingSampleActivity : AppCompatActivity() {
                             setPrefetchDistance(50)
                             build()
                         }).run {
-                    setMainThreadExecutor(UiThreadExecutor)
-                    setBackgroundThreadExecutor(AsyncTask.THREAD_POOL_EXECUTOR)
+                    setNotifyExecutor(UiThreadExecutor)
+                    setFetchExecutor(AsyncTask.THREAD_POOL_EXECUTOR)
```
ngsilverman: These methods were renamed, see https://developer.android.com/topic/libraries/architecture/release-notes.html#paging.

epoxy-processor/src/main/java/com/airbnb/epoxy/DataBindingModelInfo.kt
```diff
@@ -10,13 +10,13 @@ internal class DataBindingModelInfo(
...
-         val moduleName: String,
+        val moduleName: String,
         private val layoutPrefix: String = ""
 ) : GeneratedModelInfo() {
     private val dataBindingClassName: ClassName
 
-    val dataBindingClassElement: Element?
-        get() = getElementByName(dataBindingClassName, elementUtils, typeUtils)
+    private val dataBindingClassElement: TypeElement?
+        get() = getElementByName(dataBindingClassName, elementUtils, typeUtils) as TypeElement?
```
elihart: should this be `as?` instead?

ngsilverman: @elihart I'm thinking: if `getElementByName` doesn't return a `TypeElement` or `null` then something unexpected is happening since we're passing it a `ClassName`. In fact maybe it would make more sense to change the return type of `getElementByName(ClassName name, Elements elements, Types types)`. Thoughts?

ngsilverman: I changed the return type to `TypeElement` and everything still builds and all tests are passing so seems reasonable.

Issue #286, opened 2017-09-15
Diff on background thread for EpoxyController


elihart:
Diffing is done on the main thread right now. I don't think there are any limitations to moving it to the background thread for Epoxy Controller, which would be a nice optimization. We just have to make sure that we handle multiple diff calls at the same time correctly.

niccorder:
I'll go ahead and hop on this issue as well. 

Should we just have a `private final Executor diffExecutor = Executors.newSingleThreadExecutor()` live inside the `EpoxyController` and schedule diffs on this thread? Also, I haven't had the opportunity to handle testing in multi-threaded environments. I will most likely need your feedback for testing this.

elihart:
Awesome that you can work on this!

I think all EpoxyControllers could share the same thread so we don't have to keep creating new threads.

I'm not sure if it would be better to use `AsyncTask.THREAD_POOL_EXECUTOR` or create our own thread like you suggested.

I also don't have much experience with building multi-threaded systems so I can't give a lot of direction, but happy to discuss and code review. I should be able to help more with the Epoxy DiffHelper code changes.

niccorder:
Thanks for the info @elihart. Looking forward to the feedback on the PR.

I wanted to ask why all controllers should share the same thread? I thought it was a 1:1 relationship between controller to adapter. Thanks for your insight! I want to get as much info before beginning on this.

elihart:
@niccorder you're right that its a 1:1 relationship between controller and adapter, and generally only one controller should be diffing at one time. However, if most screens are using Epoxy it seems wasteful to create a new thread for each screen. Also, diffing usually only takes a few milliseconds and doesn't happen too often, so having a dedicated thread seems like overkill.

mradzinski:
@elihart I beg to differ regarding your statement that `diffing ... doesn't happen too often`. 

Actually, I just stumbled an app using Expoy where diffing happened at least 40 times per minute. It's an app backed up by a real-time DB (let's call it Firebase for simplicity), and since the only way of letting Epoxy know there is a change in the models (this app had listeners for likes, votes and things like that) without throwing the existent models away is to pass an ever-growing list of objects to the `setData` method of the controllers then full diffing was happening every time someone liked a post for example.

elihart:
@mradzinski thanks for sharing - in cases like that it is best practice to use `requestDelayedModelBuild` to debounce events. a delay of a second or so would prevent model rebuilding and diffing from happening too much much (to the point of causing bad performance).

[...]

PR #425, opened 2018-04-23
Async diffing and model building

elihart:
First pass at https://github.com/airbnb/epoxy/issues/286

cc @ngsilverman @gpeal


epoxy-adapter/src/main/java/com/airbnb/epoxy/AsyncListDifferWithPayload.java
```diff
@@ -0,0 +1,193 @@
...
+import java.util.concurrent.Executor;
+import java.util.concurrent.Executors;
+
+/**
+ * An adaptation of Google's {@link android.support.v7.recyclerview.extensions.AsyncListDiffer}
+ * that adds support for payloads in changes.
+ *
+ * Also adds support for canceling an in progress diff.
+ */
+class AsyncListDifferWithPayload<T> {
```
ngsilverman: Add a `@UiThread` annotation to the whole class?

epoxy-adapter/src/main/java/com/airbnb/epoxy/SimpleEpoxyController.java
```diff
@@ -10,6 +10,10 @@
   private List<? extends EpoxyModel<?>> currentModels;
   private boolean insideSetModels;
 
+  public SimpleEpoxyController() {
+    setDebugLoggingEnabled(true);
```
ngsilverman: Just curious, why make this change?

elihart: this was just for debugging purposes, I should have removed it. 

thanks for the catch!

epoxy-adapter/src/main/java/com/airbnb/epoxy/AsyncListDifferWithPayload.java
elihart: I actually want to figure out if we should support setting a new list from any thread. That could be nice when the models are built on a background thread. It would complicate this to synchronize things though. cc @gpeal

gpeal: I think the most important thing is that it happens off of the main thread rather than on a specific executor

epoxy-adapter/src/main/java/com/airbnb/epoxy/AsyncListDifferWithPayload.java
```diff
@@ -0,0 +1,193 @@
...
+      // fast simple first insert
+      updateCallback.onInserted(0, newList.size());
+      list = newList;
+      readOnlyList = Collections.unmodifiableList(newList);
+      return;
+    }
+
+    final DiffCallback<T> wrappedCallback = new DiffCallback<>(list, newList, diffCallback);
+
+    DIFF_EXECUTOR.execute(new Runnable() {
```
gpeal: This would be easier with coroutines :)

elihart: yeah :( not ready to add that dependency yet though

elihart:
There are two things left I need to address:

Update documentation to reflect that duplicate unique ids won't crash anymore. I could build debug handling to check this case however.

Right now some people rely on posting a runnable after requesting a model build to make sure something runs after item changes are notified (such as scrolling to the top of the list after an item is inserted). This will no longer work, so will be a breaking change. I'd like to make sure I have a good alternative to present, and possibly enable this conditionally to give people time to update

appsailor:
@elihart , any idea when this is getting released? Looking forward to trying it out on my upcoming android app which uses Firebase real time db. thx Aldrin

elihart:
@ngsilverman @gpeal @BenSchwab This is updated to make the differ completely thread safe (it can now handle read/write from any thread), and also adds support for async model building.

This is my approach:

```
EpoxyController has a second constructor that takes two `Handler` instances, one for running model building and one for handling diffing. By default these use the main thread, but they can be changed to allow async work.

The default handlers can be changed by setting the static fields `EpoxyController#defaultDiffingHandler` and `EpoxyController#defaultModelBuildingHandler` which lets you force async behavior for all controllers in an app.

For example, you can set the global default to async like this
    HandlerThread handlerThread = new HandlerThread("epoxy");
    handlerThread.start();
    Handler handler = new Handler(handlerThread.getLooper());
    EpoxyController.defaultDiffingHandler = handler;
   EpoxyController.defaultModelBuildingHandler = handler;

A Handler is used because epoxy needs to be able to easily work with the main thread as the default. Also, Epoxy doesn't support parallel model building, and a Handler enforces this with it's single threaded looping behavior.

Note that if you use async model building you must guarantee that all of your data access is thread safe - Epoxy does not help you with that. This means that your `EpoxyController#buildModels` implementation is thread safe in all of it's data access.

If the same Handler is provided for model building and diffing then all of that work will proceed synchronously.
```

I feel pretty good about these changes, but it definitely adds a fair amount of complexity that is hard to test. I played around with it in the sample app and it seems good, but we'll want to test this carefully in airbnb too.

I took great care to make sure we could release this without changing user's model building behavior at all. Default behavior is still posted, but then runs synchronously until adapter changes are notified. This isn't too trivial, because by default the differ posts it's work to whatever thread it is given. I set up a wrapper class to execute Runnables asynchronously if its looper is the same. This is another reason I need to use Handlers instead of an Executor.

Please review the concurrency logic carefully and lmk if you see room for improvement.

elihart:
@appsailor sorry, I just saw your comment. The work is now done, and I'll be releasing this week.

epoxy-adapter/src/main/java/com/airbnb/epoxy/AsyncEpoxyDiffer.java
```diff
@@ -0,0 +1,286 @@
...
+        }
+      }
+    });
+  }
+
+  /**
+   * Marks the generation as done, and updates the list if the generation is the most recent.
+   * Calls to this MUST be synchronized on "this" so list object and generation always stay in sync.
+   * This method isn't synchronized directly so callers have flexibility to includes other
+   * actions in their synchronized block
```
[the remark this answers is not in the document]
elihart: That makes a lot of sense, I should have looked that up :P

epoxy-adapter/src/main/java/com/airbnb/epoxy/AsyncEpoxyDiffer.java
```diff
@@ -0,0 +1,286 @@
...
+    synchronized (this) {
+      // Incrementing generation means any currently-running diffs are discarded when they finish
+      // We synchronize to guarantee list object and generation number are in sync
+      runGeneration = generationTracker.incrementAndGetNextScheduled();
+      previousList = list;
+    }
+
+    if (newList == previousList) {
+      // nothing to do
+      onRunCompleted(runGeneration, newList, new DiffResult(previousList, newList, null));
```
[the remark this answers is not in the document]
elihart: good idea

epoxy-adapter/src/main/java/com/airbnb/epoxy/AsyncEpoxyDiffer.java
```diff
@@ -0,0 +1,286 @@
...
+  /**
+   * Marks the generation as done, and updates the list if the generation is the most recent.
+   * Calls to this MUST be synchronized on "this" so list object and generation always stay in sync.
+   * This method isn't synchronized directly so callers have flexibility to includes other
+   * actions in their synchronized block
+   *
+   * @return True if the given generation is the most recent, in which case the given list was
+   * set. False if the generation is old and the list was ignored.
+   */
+  private boolean tryLatchList(@Nullable List<? extends EpoxyModel<?>> newList, int runGeneration) {
```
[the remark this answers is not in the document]
elihart: nope, anythread. annotated

PR #433, opened 2018-05-16
Reduce memory

baoti:
When I trace the number of EpoxyModel instances, I found that 3x instances hold in memory, and could't be released by GC.
This PR just want to NOTE them.

### DIFF
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/DiffHelper.java
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/DiffHelper.java
@@ -189,7 +189,14 @@ class DiffHelper {
     collectMoves(updateOpHelper);
     collectChanges(updateOpHelper);
 
+    resetOldState();
+
     return updateOpHelper;
+  }
+
+  private void resetOldState() {
+    oldStateList.clear();
+    oldStateMap.clear();
   }
 
   private void prepareStateForDiff() {
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModelGroup.java
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModelGroup.java
@@ -3,6 +3,7 @@ package com.airbnb.epoxy;
 import android.support.annotation.CallSuper;
 import android.support.annotation.LayoutRes;
 import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
 import android.view.View;
 import android.view.ViewGroup;
 import android.view.ViewGroup.LayoutParams;
@@ -129,6 +130,7 @@ public class EpoxyModelGroup extends EpoxyModelWithHolder<Holder>
 
   @Override
   public void handlePreBind(final EpoxyViewHolder holder, Holder groupHolder, final int position) {
+    groupHolder.setModelGroup(this);
     iterateModels(groupHolder, new IterateModelsCallback() {
       @Override
       public void onModel(EpoxyModel model, Object boundObject, View view, int modelIndex) {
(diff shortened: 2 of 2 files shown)


epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModelGroup.java
```diff
@@ -301,10 +307,14 @@ protected boolean useViewStubLayoutParams(EpoxyModel<?> model, int modelPosition
 
   @Override
   protected final Holder createNewHolder() {
-    return new Holder();
+    Holder holder = new Holder();
+    holder.setModelGroup(this);
```
elihart: `bindView` is only called once, when the Holder is first created. that means we only need the model group then, and not on all of the bind calls, so all this can be simplified. maybe the model group can be passed to the holder in the constructor

epoxy-adapter/src/main/java/com/airbnb/epoxy/DiffHelper.java
```diff
@@ -189,9 +189,16 @@ private UpdateOpHelper buildDiff(UpdateOpHelper updateOpHelper) {
     collectMoves(updateOpHelper);
     collectChanges(updateOpHelper);
 
+    resetOldState();
+
     return updateOpHelper;
   }
 
+  private void resetOldState() {
+    oldStateList.clear();
```
elihart: notice that `prepareStateForDiff` calls this too. it seems like we don't need to do it there if we do it here instead.

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModelGroup.java
baoti: I think should pass the model group to Holder.bindView, then the Holder instance will not hold the model group instance. In current version, the Holder instance hold the model group which create it forever, no chance to release it.

epoxy-adapter/src/main/java/com/airbnb/epoxy/DiffHelper.java
baoti: Yes, but it was delayed to next `buildModels` call. After the `buildModels` is called, oldStateList is cleared, but it hold new oldStateList instead. It always hold 2x instances (oldStateList and currentStateList).

baoti: Just like, in first time, it hold state-v1 (as oldStateList) and state-v2 (as currentStateList), after next buildModels call, it hold state-v2 (as oldStateList) and state-v3 (as currentStateList).

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModelGroup.java
baoti: > `bindView` is only called once, when the Holder is first created. that means we only need the model group then, and not on all of the bind calls, so all this can be simplified.

Yes.

elihart:
With https://github.com/airbnb/epoxy/pull/425 we won't be using this diffing code soon, so I am not too worried about that. Generally diffing happens many times over the life of the adapter, and it is released when the adapter is GC'ed so it isn't too bad, but thank you for catching it

I would like to improve the model group situation, but we need a cleaner approach

baoti:
@elihart Thanks your clean reply. I make the changes of model group cleaner.

Epoxy is a really nice UI library.

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModelGroup.java
```diff
@@ -352,6 +361,10 @@ protected void bindView(View itemView) {
 
         views.add(view);
       }
+
+      // Clear the reference, because of `EpoxyController.buildModels`
+      // will build new modelGroup instance during every call.
+      modelGroup = null;
```
elihart: clear models as well?

baoti: Yes! Will soon.

baoti:
Updated.

elihart:
Thanks! will release this soon

PR #440, opened 2018-05-30
Androidx migration

jeffreydelooff:


elihart:
Thanks, this is great! looks like the travis config needs to be updated for Android P?

jeffreydelooff:
Yes, looks like the license has to be accepted.

jeffreydelooff:
@elihart How do you normally create a snapshot or build to the local maven repository to test a new version in a real android project?

build.gradle
```diff
@@ -1,19 +1,24 @@
...
   dependencies {
-    classpath "com.android.tools.build:gradle:3.2.0-beta02"
+    classpath "com.android.tools.build:gradle:3.3.0-alpha04"
     classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$KOTLIN_VERSION"
     classpath 'com.bmuschko:gradle-nexus-plugin:2.3.1'
   }
 }
+
+plugins {
+  id 'com.github.ben-manes.versions' version '0.17.0'
```
elihart: I see this helps manage dependencies: https://github.com/ben-manes/gradle-versions-plugin

sounds useful, does it work well in your experience?

build.gradle
```diff
@@ -1,19 +1,24 @@
...
+  ext.KOTLIN_VERSION = "1.2.51"
   ext.ANDROID_PLUGIN_VERSION = "3.2.0-beta02"
 
   repositories {
     google()
     jcenter()
   }
   dependencies {
-    classpath "com.android.tools.build:gradle:3.2.0-beta02"
+    classpath "com.android.tools.build:gradle:3.3.0-alpha04"
```
elihart: are you using 3.3 canary? I don't know that we should get on the alpha track with this pr, it may be better to stick with 3.2 and merge this once it is released

build.gradle
jeffreydelooff: Yes, it helps list new milestones for all dependencies. We use it in all our company projects.

build.gradle
jeffreydelooff: No, I'll revert it back.

jeffreydelooff:
@elihart how do you normally solve conflicts? you rebase or merge origin master into the fork?

alexbchr:
New version of AndroidX (1.0.0-rc01) available since August 6th. According to the release notes:

> This release candidate of AndroidX is considered feature-complete and its public API surface is stable. This release will be shipped as final stable version barring any critical issues that may arise. This release should be safe to use in production.

Should Epoxy be released using this version, while we wait for the final version for coming out?

epoxy-integrationtest/src/test/java/com/airbnb/epoxy/ModelClickListenerTest.java
```diff
@@ -100,13 +102,19 @@ private View mockModelForClicking(EpoxyModel model) {
...
 
-    when(mockedView.getParent()).thenReturn(recyclerMock);
-    when(recyclerMock.findContainingViewHolder(mockedView)).thenReturn(holderMock);
     when(holderMock.getAdapterPosition()).thenReturn(1);
-    when(holderMock.getModel()).thenReturn(model);
+    doReturn(recyclerMock).when(mockedView).getParent();
+    doReturn(holderMock).when(recyclerMock).findContainingViewHolder(mockedView);
+    doReturn(model).when(holderMock).getModel();
+
+//    when(mockedView.getParent()).thenReturn(recyclerMock);
```
elihart: same here. issues with mockito?

update_processor_test_resources.rb
```diff
@@ -29,7 +29,7 @@ def updateTestClass(test_class_result)
     end
 
     # The test copies the source file to the build folder. We need to modify the original file to update its expected source
-    expected_source_file_path = expected_file_match.captures[0].sub "build/intermediates/classes/test/debug", "src/test/resources"
+    expected_source_file_path = expected_file_match.captures[0].sub "build/intermediates/sourceFolderJavaResources/debug", "src/test/resources"
```
elihart: thanks!

jeffreydelooff:
Resolved merge conflicts and fixed the tests.

elihart:
I've pushed `3.0.0-rc1`. I'll leave this as is for now until androidx is is fully released, and to make sure that nobody reports any issues with the build before I merge to master

alashow:
@elihart As of September 21, Android X has a stable release (1.0.0): https://developer.android.com/jetpack/androidx/androidx-rn#1.0.0

jeffreydelooff:
@elihart Just rebased again and ran into issues with the _EpoxyVisibilityTrackerTest.kt_ tests. For some reason they are off by a height of 2. Can you give any insights on this?

elihart:
Not sure why that would be, I guess some robolectric change?

@eboudrant could you give any insight since you have the most familiarity with these tests?

eboudrant:
I will check later but I guess we should increase the tolerance a bit in
the pixel height assertion.

elihart:
@jeffreydelooff I merged https://github.com/airbnb/epoxy/pull/590 - can you see if that fixed it?

jeffreydelooff:
@elihart Yes, #590 fixed it. Also have been able to enable epoxy-sample again.

elihart:
Thanks! This seems almost ready. Any idea why the manifest is having problems?
```
epoxy/epoxy-sample/src/main/AndroidManifest.xml:22:18-86 Error:
	Attribute application@appComponentFactory value=(androidx.core.app.CoreComponentFactory) from [androidx.core:core:1.0.0] AndroidManifest.xml:22:18-86
	is also present at [com.android.support:support-compat:28.0.0] AndroidManifest.xml:22:18-91 value=(android.support.v4.app.CoreComponentFactory).
```

jeffreydelooff:
No problem. No idea, I didn't have this issue yesterday and popped up today. I'll have to investigate, looks like two compat libraries both include it in their manifests.

jeffreydelooff:
Did some more digging and encountered a few more issues when enabling the epoxy-sample module:

1. Since some dependencies (litho, paris) for epoxy-sample have not migrated to androidx yet I ran into either duplicate or missing classes.
2. If I enable jetifier to fix these dependencies it cannot transform the rt.jar from the java runtime dependency in epoxy-processortest.

elihart:
@jeffreydelooff ah, that sounds quite tricky. Is it possible to enable jetifier for only the `epoxy-sample` module? does that need to affect `epoxy-processortest`?

any thoughts on ways around this, unless we leave the sample broken for now until other dependencies upgrade to androidx?

jeffreydelooff:
@elihart I looked into ways around it but couldn't find any clues. Neither enabling per module or able to exclude certain jars. For now that would be our only option.

shaishavgandhi05:
I think you can enable per module if you add a `gradle.properties` to that module and only enable jetifier there. I faced something similar with my `processor` module.

shaishavgandhi05:
Here it is for reference where I've disabled it in the `library` module.
https://github.com/shaishavgandhi05/navigator/blob/master/library/gradle.properties#L6

jeffreydelooff:
@shaishavgandhi05 Thanks for the suggestion, I wish it was that simple. I tried it but the epoxy-processortest still relies on Paris which generates a `ModelViewWithParisStyleApplier.java` file which still references `com.android.support.*` so those also need to be migrated/jetified. This testStyleableView test is the only one that fails.

jeffreydelooff:
Maybe it is easier to just migrate Paris as well. Will see if I can find the time or if someone else feels up to it. Looks like there is no issue or pr yet.

elihart:
@ngsilverman can your team migrate Paris to androidx? We need to do it for airbnb anyway

elihart:
Does litho cause any problems once you use jetifier in the sample module? if so I could just remove it, I don't think the Litho integration gets much/any usage

ngsilverman:
@elihart yeah, funny time I was actually working on this yesterday. I'll make it a priority.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# WIP 3.0.0 (TBD)
+
+- **Breaking** Migrated to androidx packages
```
ngsilverman: Why is this considered a breaking change? 🤔

alashow: @ngsilverman I think users who didn't migrate to androidx can't use libraries migrated to androidx.

update_processor_test_resources.rb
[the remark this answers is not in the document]
jeffreydelooff: I ran into this when I first started this PR. Looks like something has changed in the meantime.

CHANGELOG.md
jeffreydelooff: Yes, I think it is all or nothing.

ngsilverman: I haven't found any information about that. What makes you think that's the case?

alashow: Jetifier transforms old package names to new package names. But there is no Jetifier that transforms from new package names to old ones.

alashow: Most probably. For example, setting epoxy adapter (which extends androidx packaged base `RecyclerView.Adapter`) to `RecyclerView` with old package name won't work because it expects an old package named `RecyclerView.Adapter`. Right?
Assuming user uses just `RecyclerView` not `EpoxyRecyclerView`.

ngsilverman: Ah, I see, makes sense for this use case. I'm also trying to figure out if it should be considered a breaking change for Paris (https://github.com/airbnb/paris) but it doesn't seem so since there's no input parameter type from the AndroidX library.

alashow: You probably will need to test that out, if you didn't, by importing migrated Paris to old non-migrated sample app, as there might be weird clashes between AndroidX and old packages.

jeffreydelooff:
Just updated Paris and excluded com.android.support for the Litho dependencies. Litho would indeed be the only problem left but with the exclusion it still works.

elihart:
This is great! thanks for the continued work, looks like I can merge it now. I'll release a RC2 with these changes, and if there are no problems with that after a while I'll push a final release.

Thanks @jeffreydelooff !

jeffreydelooff:
No problem. Has been a fun experience.

We are actively using epoxy as well. :)

PR #444, opened 2018-06-04
Update README.md

edwinnyawoli:
Updated the docs to show that the `getAdapter()` method is used to get the adapter that is set on a recycler view.


edwinnyawoli:
👍

PR #450, opened 2018-06-15
Generate model factory methods

ngsilverman:
- Creates a new module, `epoxy-modelfactory`
  - Houses the new `ModelProperties` interface
  - Depends on Paris to enable `ModelProperties` to set styles
  - Contains the relevant tests
    - I tried adding a dependency on `epoxy-processortest` to reuse `ProcessorTestUtils` but that caused mysterious Lint issues so I copied a single method from it over instead.
- If this new module is present, for all models that have an empty constructor, no property group, and at least one property of a supported type, a static `from(ModelProperties properties)` method is generated.
```java
  public static MyViewModel_ from(ModelProperties properties) {
    MyViewModel_ model = new MyViewModel_();
    model.id(properties.getString("id"));
    // This is repeated for each attribute of a supported type
    if (properties.has("title")) {
      model.title(properties.getString("title"));
    }
    ...
    // If it's styleable
    Style style = properties.getStyle();
    if (style != null) {
      model.style(style);
    }
    return model;
  }
```

TODO
- [x] Fix Lint errors
- [x] Is the new module's package name setup correct?

(I rebased and reformatted `GeneratedModelWriter` but nothing changed. Maybe it hadn't been properly formatted in the first place?)


epoxy-processor/src/main/java/com/airbnb/epoxy/GeneratedModelWriter.kt
```diff
@@ -998,7 +1089,7 @@ internal class GeneratedModelWriter(
             if (attr is ViewAttributeInfo && attr.generateStringOverloads) {
                 methods.addAll(StringOverloadWriter(modelInfo, attr, configManager).buildMethods())
             } else {
-                if (attr.isViewClickListener) {
+                if (attr.isViewClickListener || attr.isViewLongClickListener) {
```
ngsilverman: ^ this is a change (though not a functional one, I modified `isViewClickListener` so I could use it elsewhere).

epoxy-processor/src/main/java/com/airbnb/epoxy/GeneratedModelWriter.kt
```diff
@@ -1295,6 +1412,106 @@ internal class GeneratedModelWriter(
         return method
     }
 
+    /**
```
ngsilverman: And here's the main addition besides tests.

epoxy-processortest/build.gradle
```diff
@@ -11,6 +11,12 @@ android {
   defaultConfig {
     minSdkVersion rootProject.MIN_SDK_VERSION
     targetSdkVersion rootProject.TARGET_SDK_VERSION
+
+    javaCompileOptions {
+      annotationProcessorOptions {
+        includeCompileClasspath false
```
ngsilverman: This makes it so Gradle doesn't complain that there are undeclared annotation processors in the classpath.

epoxy-modelfactory/src/main/java/com/airbnb/epoxy/ModelProperties.java
```diff
@@ -0,0 +1,39 @@
...
+public interface ModelProperties {
+
+  boolean has(@NonNull String propertyName);
+
+  boolean getBoolean(@NonNull String propertyName);
+
+  double getDouble(@NonNull String propertyName);
+
+  @NonNull
+  Drawable getDrawable(@NonNull String propertyName);
```
elihart: could support `getDrawableRes` by looking for @DrawableRes annotation

epoxy-modelfactory/src/main/java/com/airbnb/epoxy/ModelProperties.java
```diff
@@ -0,0 +1,39 @@
...
+import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
+import android.view.View.OnClickListener;
+
+import com.airbnb.paris.styles.Style;
+
+import java.util.List;
+
+public interface ModelProperties {
+
```
elihart: dedicated `getId()` method?

epoxy-modelfactory/src/main/java/com/airbnb/epoxy/ModelProperties.java
```diff
@@ -0,0 +1,39 @@
...
+import android.graphics.drawable.Drawable;
+import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
+import android.view.View.OnClickListener;
+
+import com.airbnb.paris.styles.Style;
+
+import java.util.List;
+
+public interface ModelProperties {
```
elihart: we can use kotlin for this module

epoxy-modelfactory/src/test/java/com/airbnb/epoxy/ProcessorTestUtils.kt
```diff
@@ -0,0 +1,36 @@
...
+import com.airbnb.paris.processor.ParisProcessor
+import com.google.common.truth.Truth.assert_
+import com.google.testing.compile.JavaFileObjects
+import com.google.testing.compile.JavaSourcesSubjectFactory.javaSources
+import javax.annotation.processing.Processor
+import javax.tools.JavaFileObject
+
+internal object ProcessorTestUtils {
+    @JvmOverloads
+    @JvmStatic
```
elihart: don't need the jvm helpers

epoxy-processor/src/main/java/com/airbnb/epoxy/GeneratedModelWriter.kt
```diff
@@ -1295,6 +1412,106 @@ internal class GeneratedModelWriter(
...
+        // Models that don't have an empty constructor are not supported because there would be no
+        // clear way to create new instances
+        val hasEmptyConstructor = modelInfo.constructors.isEmpty()
+                || modelInfo.constructors.any { it.params.isEmpty() }
+        if (!hasEmptyConstructor) {
+            return
+        }
+
+        val supportedAttributeInfo = modelInfo.attributeInfo.filter { attributeInfo ->
+            listOf(
```
elihart: might be easier to read to have this list declared outside the loop

epoxy-processor/src/main/java/com/airbnb/epoxy/GeneratedModelWriter.kt
```diff
@@ -1295,6 +1412,106 @@ internal class GeneratedModelWriter(
...
+                val jsonGetterName = when {
+                    attributeInfo.isBoolean -> "getBoolean"
+                    attributeInfo.isCharSequence
+                            || attributeInfo.isStringAttributeData -> "getString"
+                    attributeInfo.isCharSequenceList -> "getStringList"
+                    attributeInfo.isDouble -> "getDouble"
+                    attributeInfo.isDrawable -> "getDrawable"
+                    attributeInfo.isInt -> "getInt"
+                    attributeInfo.isViewClickListener -> "getOnClickListener"
+                    else -> throw IllegalStateException()
```
elihart: Put a message in the exception

epoxy-processor/src/main/java/com/airbnb/epoxy/GeneratedModelWriter.kt
```diff
@@ -1295,6 +1412,106 @@ internal class GeneratedModelWriter(
...
+            }
+        }
+
+        // If none of the properties are of a supported type the method isn't generated
+        if (supportedAttributeInfo.isEmpty()) {
+            return
+        }
+
+        // Groups aren't supported because we wouldn't know which type to choose from
+        if (modelInfo.attributeGroups.any { it.attributes.size > 1 }) {
```
elihart: check this before supported attributes?

epoxy-processor/src/main/java/com/airbnb/epoxy/GeneratedModelWriter.kt
```diff
@@ -1295,6 +1412,106 @@ internal class GeneratedModelWriter(
...
+                AttributeInfo::isInt,
+                AttributeInfo::isViewClickListener,
+                AttributeInfo::isStringAttributeData
+            ).any {
+                it.invoke(attributeInfo)
+            }
+        }
+
+        // If none of the properties are of a supported type the method isn't generated
+        if (supportedAttributeInfo.isEmpty()) {
```
elihart: should it be an error if any attributes are not supported?

epoxy-processortest/build.gradle
elihart: leave a code comment about it?

ngsilverman: That seems a bit redundant, the config itself is pretty explicit: don't include annotation processors that are in the compile classpath.

ngsilverman:
@elihart addressed all the feedback. :)

epoxy-modelfactory/src/main/java/com/airbnb/epoxy/ModelProperties.kt
```diff
@@ -0,0 +1,32 @@
+package com.airbnb.epoxy
+
+import android.support.annotation.DrawableRes
+import android.view.View.OnClickListener
+import com.airbnb.paris.styles.Style
+
+interface ModelProperties {
+
+    fun getId(): String
```
elihart: why fun over val?

epoxy-processortest/build.gradle
elihart: well you still left a github comment explaining it, so it isn't _too_ explicit ;)

ngsilverman: Only because it doesn't seem directly related to this PR. 😛

epoxy-modelfactory/src/main/java/com/airbnb/epoxy/ModelProperties.kt
ngsilverman: With everything else being a function it seemed inconsistent/out of place. 🤷‍♂️

epoxy-processor/src/main/java/com/airbnb/epoxy/AttributeInfo.java
```diff
@@ -231,7 +233,57 @@ public int hashCode() {
...
+
+  boolean isStringList() {
+    return Utils.isType(getTypeMirror(), "java.util.List<java.lang.String>")
+        || Utils.isType(getTypeMirror(), "java.util.List<kotlin.String>");
+  }
+
+  boolean isEpoxyModelList() {
+    return Utils.isType(
+        getTypeMirror(),
+        "java.util.List<? extends com.airbnb.epoxy.EpoxyModel<?>>"
```
elihart: what about kotlin support?

epoxy-processor/src/main/java/com/airbnb/epoxy/AttributeInfo.java
```diff
@@ -231,7 +233,57 @@ public int hashCode() {
...
+  boolean isBoolean() {
+    return getTypeMirror().getKind() == TypeKind.BOOLEAN;
+  }
+
+  boolean isCharSequenceOrString() {
+    return Utils.isCharSequenceOrStringType(getTypeMirror());
+  }
+
+  boolean isStringList() {
+    return Utils.isType(getTypeMirror(), "java.util.List<java.lang.String>")
```
elihart: maybe a good way to DRY up the two list checks is to have a `isList(type)` method that checks for java/kotlin lists and also checks the generic type as a parameter

epoxy-processor/src/main/java/com/airbnb/epoxy/AttributeInfo.java
elihart: would be nice to use a ClassName constant or something, not sure if that is possible with the wildcard

epoxy-processor/src/main/java/com/airbnb/epoxy/Utils.java
```diff
@@ -114,6 +116,21 @@ static boolean isViewLongClickListenerType(TypeMirror type) {
...
+        || Utils.isType(type, "kotlin.String");
+  }
+
+  static boolean isCharSequenceOrStringType(TypeMirror type) {
+    return Utils.isType(type, "java.lang.CharSequence")
+        || Utils.isType(type, "kotlin.CharSequence")
+        || isStringType(type);
+  }
+
+  static boolean isStringAttributeDataType(TypeMirror type) {
```
elihart: better to have a generic `isType(TypeMirror , ClassName)` method?

epoxy-processor/src/main/java/com/airbnb/epoxy/GeneratedModelWriter.kt
```diff
@@ -1295,6 +1412,115 @@ internal class GeneratedModelWriter(
...
+     * values provided via the ModelProperties interface. Notably, this generated method makes it
+     * easy to create models from JSON and various other data formats.
+     */
+    private fun addFromPropertiesMethodIfNeeded(
+        classBuilder: TypeSpec.Builder,
+        modelInfo: GeneratedModelInfo
+    ) {
+        // The epoxy-modelfactory module must be present to enable this functionality
+        val hasModelFactoryDependency =
+            getTypeMirror(EPOXY_MODEL_PROPERTIES.reflectionName(), elements) != null
```
elihart: opportunity for an `elements.isTypeLoaded(ClassName)` extension

epoxy-processor/src/main/java/com/airbnb/epoxy/GeneratedModelWriter.kt
```diff
@@ -1295,6 +1412,115 @@ internal class GeneratedModelWriter(
...
+        if (!hasModelFactoryDependency) {
+            return
+        }
+
+        // Models that don't have an empty constructor are not supported because there would be no
+        // clear way to create new instances
+        val hasEmptyConstructor = modelInfo.constructors.isEmpty()
```
elihart: could this be a method on the ModelInfo class?

epoxy-processor/src/main/java/com/airbnb/epoxy/AttributeInfo.java
ngsilverman: @elihart after doing some testing I just realized that we don't need to check for Kotlin types at all (for lists, strings or charsequences alike). Guess they get converted to Java types for interop.

ngsilverman:
@elihart Added a Kotlin view test and followed your suggestions. Let me know if this is good to go. Do you want me to handle the release?

elihart:
@ngsilverman yeah, seems good to go. if you can do the release that'd be great

PR #458, opened 2018-06-26
Fix generated ModelViewModel_ OnModelClickListener Setter's Annotations

baoti:
Fix wrong setter method annotation for model builder.

Now, the `@Nullable` annotation on the method:
```
@Nullable
public TestCallbackPropViewModel_ listener(final OnModelClickListener<TestCallbackPropViewModel_, TestCallbackPropView> listener) {
  // ...
  return this;
}
```

The PR fix it to annotation on the parameter:
```
public TestCallbackPropViewModel_ listener(@Nullable final OnModelClickListener<TestCallbackPropViewModel_, TestCallbackPropView> listener) {
  // ...
  return this;
}
```


baoti:
@elihart I make this change move `@Nullable` to `OnModelClickListener` parameter.

baoti:
That doesn't matter. Thanks! :smiley:

PR #487, opened 2018-07-31
[modelfactory] Enable grouped fields with different names, add support for long and @RawRes

ngsilverman:
Apologies for bundling 3 changes into a single commit. By the time I realized what I had done it had become really hard to detangle the changes. They're relatively small so hopefully it's not too much of an issue. Unit tests end up being most of the diff.

This does 3 things:
- Adds support for longs
- Adds support for @RawRes ints (for Lottie animations)
- Adds support for fields that are part of the same group as long as their name is different

@elihart


epoxy-processor/src/main/java/com/airbnb/epoxy/AttributeInfo.kt
```diff
@@ -122,23 +121,29 @@ internal abstract class AttributeInfo {
...
-            return false
         }
+        return false
+    }
 
     val isInt: Boolean
         get() = typeMirror.kind == TypeKind.INT
 
+    val isLong: Boolean
+        get() = typeMirror.kind == TypeKind.LONG
```
elihart: Does this work for both `Long` and `long`?

epoxy-processor/src/main/java/com/airbnb/epoxy/AttributeInfo.kt
```diff
@@ -122,23 +121,29 @@ internal abstract class AttributeInfo {
...
-                            return true
-                        }
-                    }
+        get() = isInt && hasAnnotation("DrawableRes")
+
+    val isRawRes: Boolean
+        get() = isInt && hasAnnotation("RawRes")
+
+    private fun hasAnnotation(annotationSimpleName: String): Boolean {
+        for (annotation in setterAnnotations) {
```
elihart:
```
setterAnnotations
.map { it.type }
.filterIsInstance<ClassName>()
.any { it.simpleName() == annotationSimpleName)
```

epoxy-processor/src/main/java/com/airbnb/epoxy/GeneratedModelWriter.kt
```diff
@@ -1494,28 +1498,54 @@ internal class GeneratedModelWriter(
...
+                                || attributeInfo.isStringAttributeData -> "getString"
+                        attributeInfo.isDouble -> "getDouble"
+                        attributeInfo.isDrawableRes -> "getDrawableRes"
+                        attributeInfo.isEpoxyModelList -> "getEpoxyModelList"
+                        attributeInfo.isInt && !attributeInfo.isDrawableRes && !attributeInfo.isRawRes -> "getInt"
+                        attributeInfo.isLong -> "getLong"
+                        attributeInfo.isRawRes -> "getRawRes"
+                        attributeInfo.isStringList -> "getStringList"
+                        attributeInfo.isViewClickListener -> "getOnClickListener"
+                        else -> throw IllegalStateException("Missing ModelProperties method for a supported attribute type.")
```
elihart: should use the errorLogger instead of throwing, so generating the rest of the models is not blocked. Otherwise error output is terrible since models don't exist

epoxy-processor/src/main/java/com/airbnb/epoxy/AttributeInfo.kt
ngsilverman: No, good catch. I'll add boxed support for all primitive types.

Issue #500, opened 2018-08-20
Compile errors with lambda in constructor


shaishavgandhi05:
If you have a model like such:
```kotlin
@EpoxyModelClass(layout = R.layout.activity_kotlin_sample)
abstract class KotlinConstructorSample(val listener: (someBool: Boolean, anotherInt: Int) -> Unit): EpoxyModelWithHolder<KotlinConstructorHolder>() {

}

class KotlinConstructorHolder(): EpoxyHolder() {
    override fun bindView(itemView: View) {

    }

}
```

The `EpoxyModelKotlinExtensions.kt` generated is this:
```kotlin
inline fun EpoxyController.kotlinConstructorSample(listener: Function2<in Boolean, in Int, Unit>, modelInitializer: KotlinConstructorSampleBuilder.() -> Unit) {
    KotlinConstructorSample_(listener).apply  {
        modelInitializer()
    }
    .addTo(this)
}
``` 

The `Function2` used gives a warning as it uses `kotlin.jvm.functions.Function2` instead of `kotlin.Function2`. 

The other more important error is that the `listener` should be `noinline`-ed. According to my understanding, since we're using the lambda to pass it to `KotlinConstructorSample_`, it's used as an object, meaning it can't be inlined. So we have to add a `noinline` to lambdas passed to constructor of the generated models.

PR #501, opened 2018-08-20
Fallback to kotlin.Function + noinline

shaishavgandhi05:
Initial pass for fixing issue in #500 .

I would love to add some tests for this, but I couldn't find any tests for Kotlin extension that are generated. 

#500 has relevant details on the exact thing causing the error.


epoxy-processor/src/main/java/com/airbnb/epoxy/PoetExtensions.kt
```diff
@@ -184,12 +189,16 @@ fun JavaParameterSpec.toKPoet(): KotlinParameterSpec {
 
     val nullable = annotations.any { (it.type as? JavaClassName)?.simpleName() == "Nullable" }
 
-    return KotlinParameterSpec.builder(
+    val parameterBuilder = KotlinParameterSpec.builder(
         paramName,
         type.toKPoet(nullable),
         *modifiers.toKModifier().toTypedArray()
-    ).build()
+    )
```
elihart: use `apply { }` instead?

epoxy-processor/src/main/java/com/airbnb/epoxy/PoetExtensions.kt
```diff
@@ -78,6 +79,10 @@ private fun JavaClassName.getPackageNameInKotlin(): String {
     return packageName()
 }
 
+fun isLambda(type: JavaTypeName): Boolean {
```
elihart: seems that this should check the kotlin package too, since any other random class could technically contain this

epoxy-processor/src/main/java/com/airbnb/epoxy/PoetExtensions.kt
```diff
@@ -54,18 +55,18 @@ fun JavaClassName.toKPoet(): KotlinClassName {
...
+            isBoxedPrimitive -> kotlinPkg
+            isLambda(this) -> kotlinPkg
+            else -> when (simpleName()) {
                 "Collection",
                 "List",
                 "Map",
                 "Set",
                 "Iterable" -> kotlinCollectionsPkg
-                "String" -> kotlinPkg
+                "String", "Function1" -> kotlinPkg
```
elihart: is it necessary to check `Function1` since the `isLambda` check is already done?

shaishavgandhi05: Oops that's my bad. Leftover code from when i was experimenting.

shaishavgandhi05:
Thanks! Will add this sample to `kotlinsample` module. 

I've struggled with Kotlin code generation testing also since we don't have an equivalent to the Google compile testing library for Kotlin. 

I usually fallback to adding sample classes in the `test` directory of the sample and make sure it compiles. I think adding it in the sample will eventually muddy the module for people who're only looking for samples. By adding in the `test` directory, we can be more carefree with the implementation details of the models (and lazy 😄 ).

What do you think?

elihart:
@shaishavgandhi05 that makes sense, the test directory would be great. Thanks a lot!

Hopefully we will have a kotlin compile testing library someday :P

epoxy-processor/src/main/java/com/airbnb/epoxy/PoetExtensions.kt
shaishavgandhi05: Added a check to see if it also contains the keyword `kotlin`.

shaishavgandhi05:
Anything blocking here? Have addressed the PR comments.

elihart:
Nothing blocking, looks great, thanks!

PR #509, opened 2018-08-24
Async improvements

elihart:
- Make the differ post back to the main thread async when it is run in the background. Inspiration from https://medium.com/@sweers/rxandroids-new-async-api-4ab5b3ad3e93
- Provide `AsyncEpoxyControlller` to easily use async behavior out of the box.
- Also bumps target/compile api to 28

### DIFF
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/AsyncEpoxyController.java
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/AsyncEpoxyController.java
@@ -1,0 +1,41 @@
+package com.airbnb.epoxy;
+
+import android.os.Handler;
+
+import static com.airbnb.epoxy.EpoxyAsyncUtil.MAIN_THREAD_HANDLER;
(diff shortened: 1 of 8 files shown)


epoxy-adapter/src/main/java/com/airbnb/epoxy/AsyncEpoxyController.java
```diff
@@ -0,0 +1,41 @@
...
+    this(true);
+  }
+
+  /**
+   * @param enableAsync True to do model building and diffing asynchronously, false to do them
+   *                    both on the main thread.
+   */
+  public AsyncEpoxyController(boolean enableAsync) {
+    this(true, true);
+  }
```
baoti: `enableAsync` is unused.

elihart: @baoti whoops! thanks a lot for catching that

PR #519, opened 2018-09-02
Reflect the annotations declared in constructor params

shaishavgandhi05:
If I declare a model like:
```kotlin
abstract class AnnotationModel(@StringRes val resId: Int): EpoxyModelWithHolder<AnnotationHolder>() {}
```
the Kotlin extension generated doesn't add the annotation to the extension function. However, it is added to the `AnnotationModel_` class. 

This PR fixes that.

### DIFF
--- a/epoxy-processor/src/main/java/com/airbnb/epoxy/PoetExtensions.kt
+++ b/epoxy-processor/src/main/java/com/airbnb/epoxy/PoetExtensions.kt
@@ -1,5 +1,6 @@
 package com.airbnb.epoxy
 
+import com.squareup.javapoet.CodeBlock
 import com.squareup.javapoet.TypeName
 import com.squareup.kotlinpoet.ANY
 import com.squareup.kotlinpoet.BOOLEAN
@@ -103,6 +104,8 @@ private fun JavaClassName.getSimpleNamesInKotlin(): List<String> {
     return originalNames
 }
 
+// Does not support transferring complex annotations which
+// have parameters and values associated with them.
 fun JavaAnnotationSpec.toKPoet(): KotlinAnnotationSpec {
     val annotationClass = KotlinClassName.bestGuess(type.toString())
     return KotlinAnnotationSpec.builder(annotationClass).build()
--- a/epoxy-processor/src/main/java/com/airbnb/epoxy/PoetExtensions.kt
+++ b/epoxy-processor/src/main/java/com/airbnb/epoxy/PoetExtensions.kt
@@ -13,6 +13,7 @@ import com.squareup.kotlinpoet.LONG
 import com.squareup.kotlinpoet.SHORT
 import com.squareup.kotlinpoet.UNIT
 import javax.lang.model.element.Modifier
+import kotlin.reflect.jvm.internal.impl.types.KotlinType
 
 typealias JavaClassName = com.squareup.javapoet.ClassName
 typealias JavaTypeName = com.squareup.javapoet.TypeName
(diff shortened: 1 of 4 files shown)


epoxy-processor/src/main/java/com/airbnb/epoxy/PoetExtensions.kt
```diff
@@ -102,6 +103,11 @@ private fun JavaClassName.getSimpleNamesInKotlin(): List<String> {
     return originalNames
 }
 
+fun JavaAnnotationSpec.toKPoet(): KotlinAnnotationSpec {
+    val annotationClass = KotlinClassName.bestGuess(type.toString())
```
elihart: This isn't quite right because it won't capture parameter values in the annotation - so it will only work for basic annotations.

If you're interested it would be great to add that full functionality, otherwise can you add a note warning of this partial implementation?

elihart: @shaishavgandhi05 great! yes, I think adding tests to `epoxy-processor` makes sense

shaishavgandhi05: After some investigation, this seems like something really hard to do. Parameters are added as `CodeBlock`'s to the `AnnotationSpec`. There's no way to convert a `JavaCodeBlock` to `KotlinCodeBlock` since `CodeBlock`'s have no getters to know what values they're going to emit.

shaishavgandhi05:
Have updated the PR with a bunch of tests for PoetExtensions. Also, as noted in the comment before, some investigation revealed that it was hard to transfer parameters from `JavaAnnotationSpec` to `KotlinAnnotationSpec` since `JavaCodeBlock` has no accessors that will let me move it to `KotlinCodeBlock`.

epoxy-processor/src/test/java/com/airbnb/epoxy/PoetExtensionsTest.kt
```diff
@@ -0,0 +1,79 @@
...
+
+import android.support.annotation.NonNull
+import com.squareup.kotlinpoet.asTypeName
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertTrue
+import org.junit.Test
+import javax.lang.model.element.Modifier
+
+class PoetExtensionsTest {
```
elihart: Thanks a lot for the tests!

epoxy-processor/src/main/java/com/airbnb/epoxy/PoetExtensions.kt
```diff
@@ -102,6 +104,13 @@ private fun JavaClassName.getSimpleNamesInKotlin(): List<String> {
     return originalNames
 }
 
+// Does not support transferring complex annotations which
+// have parameters and values associated with them.
+fun JavaAnnotationSpec.toKPoet(): KotlinAnnotationSpec {
```
elihart: maybe this can be nullable, and we would return null if the `JavaAnnotationSpec` has a code block?

shaishavgandhi05:
Yup you're right. Don't want to break existing functionality. I've added a test for the null case as well as added it to the sample model so that we are sure it doesn't cause any compile errors.

elihart:
Excellent, thanks again! I'll try to push a new release asap

shaishavgandhi05:
No hurries. It's a nice to have, not a deal breaker 😄

PR #528, opened 2018-09-06
Add support for setting the Padding via resource or directly in dp

### LINKED ISSUE #514: Support for setting the item spacing of a carousel
Currently it is possible to set the Padding of a carousel via a resource or dp while the item spacing can only be set as part of the `Carousel.Padding` object and only directly in pixels. Would be great to be able to set the item spacing as dp or resource too.

As i understand it's only a matter of adding a `@ModelProp` annotation right?

pwillmann:
fixes #514

### DIFF
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/Carousel.java
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/Carousel.java
@@ -433,7 +433,7 @@ public class Carousel extends EpoxyRecyclerView {
      *     decoration.
      * @param paddingType Unit / Type of the given paddings/ itemspacing.
      */
-    public Padding(
+    private Padding(
         int top, int bottom, int left, int right, int itemSpacing, PaddingType paddingType) {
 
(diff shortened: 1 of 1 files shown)


epoxy-adapter/src/main/java/com/airbnb/epoxy/Carousel.java
```diff
@@ -318,37 +333,115 @@ public void setPadding(@Nullable Padding padding) {
...
      */
-    public Padding(@Px int topPx, @Px int bottomPx, @Px int leftPx, @Px int rightPx,
-        @Px int itemSpacingPx) {
-
-      this.topPx = topPx;
-      this.bottomPx = bottomPx;
-      this.leftPx = leftPx;
-      this.rightPx = rightPx;
-      this.itemSpacingPx = itemSpacingPx;
+    public Padding(
```
elihart: I don't think we need to expose this constructor since all types can be set directly already, and having this additional way to do it might be confusing

PR #533, opened 2018-09-08
Paged list controller

yigit:
This CL introduces a new class that caches PagedList models but still allows injecting headers, filtering etc (based on models instead of items).


epoxy-paging/src/main/java/com/airbnb/epoxy/paging/CachingPagingEpoxyController.kt
```diff
@@ -0,0 +1,114 @@
...
+   * This function adds all built models to the adapter. You can override this method to add extra
+   * items into the model list or remove some.
+   */
+  open fun addModels(models: List<EpoxyModel<*>>) {
+    super.add(models)
+  }
+
+  /**
+   * Builds the model for a given item. This must return a single model for each item. If you want
+   * to inject headers etc, you can override [addModels] function.
```
elihart: Worth mentioning that a null T means a placeholder should be used?

epoxy-paging/src/main/java/com/airbnb/epoxy/paging/CachingPagingEpoxyController.kt
```diff
@@ -0,0 +1,114 @@
...
+   */
+  abstract fun buildItemModel(currentPosition: Int, item: T?): EpoxyModel<*>
+
+  override fun onModelBound(
+    holder: EpoxyViewHolder,
+    boundModel: EpoxyModel<*>,
+    position: Int,
+    previouslyBoundModel: EpoxyModel<*>?
+  ) {
+    // TODO the position may not be a good value if there are too many injected items.
```
elihart: this seems like it would rarely be a problem, but would be nice to handle. I don't see a great way to handle it well, but here is one general idea

1. Have a way to check whether `boundModel` is of type `T`
2. If so, have a method `modelCache.loadAround(boundModel)`
3. Have the model cache look up the model's position

Speed would be a concern - maybe an approach like this could be async

epoxy-paging/src/main/java/com/airbnb/epoxy/paging/PagedListModelCache.kt
```diff
@@ -0,0 +1,157 @@
...
+import java.lang.IllegalStateException
+import java.util.concurrent.Executor
+
+/**
+ * A PagedList stream wrapper that caches models built for each item. It tracks changes in paged lists and caches
+ * models for each item when they are invalidated to avoid rebuilding models for the whole list when PagedList is
+ * updated.
+ */
+internal class PagedListModelCache<T>(
+    private val modelBuilder: (Int, T?) -> EpoxyModel<*>,
```
elihart: nit, name the parameters for clarity?
```
(itemIndex: Int, item: T?) -> EpoxyModel<*>
```

epoxy-paging/src/main/java/com/airbnb/epoxy/paging/PagedListModelCache.kt
```diff
@@ -0,0 +1,157 @@
...
+    @Suppress("UNCHECKED_CAST")
+    return modelCache as List<EpoxyModel<*>>
+  }
+
+  fun loadAround(position: Int) {
+    triggerLoadAround(position)
+    lastPosition = position
+  }
+
+  fun triggerLoadAround(position: Int) {
```
elihart: should this be private?

elihart:
I can follow up with a change to the wiki once this is merged

epoxy-paging/src/main/java/com/airbnb/epoxy/paging/CachingPagingEpoxyController.kt
```diff
@@ -0,0 +1,114 @@
...
+   */
+  itemDiffCallback: DiffUtil.ItemCallback<T> = DEFAULT_ITEM_DIFF_CALLBACK as DiffUtil.ItemCallback<T>
+) : EpoxyController(modelBuildingHandler, diffingHandler) {
+  // this is where we keep the already built models
+  private val modelCache = PagedListModelCache(
+    modelBuilder = { pos, item ->
+      buildItemModel(pos, item)
+    },
+    rebuildCallback = {
+      requestDelayedModelBuild(0)
```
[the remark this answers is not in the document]
yigit: I was worried about sending too many especially for the first initialization but i guess it will only cause 1 unnecessary binding. replaced

epoxy-paging/src/main/java/com/airbnb/epoxy/paging/CachingPagingEpoxyController.kt
yigit: done.

epoxy-paging/src/main/java/com/airbnb/epoxy/paging/CachingPagingEpoxyController.kt
yigit: yea not sure what to do here, may get even worse if someone lays out items in reverse etc in `addModels`. I can consider commenting about this in `addModels` but i believe that is distraction as people are unlikely to do something like that. 

we cannot check `boundModel` is type `T` because we don't know what the bound model refers to  (we didn't create it). We can consider maintaining a map from `boundModel` to `position`. Seems like that is the only feasible solution.
Can we do it as a followup ?

epoxy-paging/src/main/java/com/airbnb/epoxy/paging/CachingPagingEpoxyController.kt
```diff
@@ -0,0 +1,114 @@
...
+ *
+ * Internally, it caches the model for each item in the [PagedList]. You should override
+ * [buildItemModel] method to build the model for the given item. Since [PagedList] might include
+ * `null` items if placeholders are enabled, this method needs to handle `null` values in the list.
+ *
+ * By default, the model for each item is added  to the model list. To change this behavior (to
+ * filter items or inject extra items), you can override [addModels] function and manually add built
+ * models.
+ */
+abstract class CachingPagingEpoxyController<T>(
```
yigit: I'm not happy with this name actually :/ the caching part bothers me.
Should we call it `PagedListEpoxyController`? I didn't go with it since there is `PagingEpoxyController`

elihart: @yigit sure, I like `PagedListEpoxyController`. I am in favor of getting rid of `PagingEpoxyController` eventually once this is adopted

epoxy-paging/src/main/java/com/airbnb/epoxy/paging/CachingPagingEpoxyController.kt
elihart: sure, I think a followup would be great. This is fine for now

elihart:
@yigit this looks ready to merge :) if you would like to rename the class to `PagedListEpoxyController` I'll wait for that

elihart:
Awesome, thanks again!

alashow:
Hi. How can we start using it & any docs?

elihart:
@alashow A version has not been released with this change yet. I can push one  sometime this week.  I'll update this thread when I do

Issue #535, opened 2018-09-10
Unifying the carousel padding parameter orders ? 


martinbonnin:
Typical parameter order in the android framework is `left, top, right, bottom`. This works for [View.setPadding](https://developer.android.com/reference/android/view/View.html#setPadding(int,%20int,%20int,%20int)), [Rect](https://developer.android.com/reference/android/graphics/Rect.html#Rect(int,%20int,%20int,%20int)), [Canvas](https://developer.android.com/reference/android/graphics/Canvas.html#drawRect(float,%20float,%20float,%20float,%20android.graphics.Paint)) and many others.

Would you be open to changing the order of [Carousel.Padding](https://github.com/airbnb/epoxy/blob/82d4161f9340d3fcb6e0ac4b6dd2f9fc502727e9/epoxy-adapter/src/main/java/com/airbnb/epoxy/Carousel.java#L396) which currently uses `top, bottom, left, right`?

That would make my brain a lot less confused. That being said, this will most likely break api-compatibilty at the same time so maybe it's too late ?

shaishavgandhi05:
I think Epoxy will do a 3.0 with AndroidX. That might be a good opportunity to get this change in.

martinbonnin:
Pull request done at https://github.com/airbnb/epoxy/pull/536/files if that can make it for 3.0.
There's a lot of copy/paste down there, hopefully I didn't forget to swap things...

elihart:
Finally merged! will focus on 3.0.0 release next to include this. 

thanks again :)

PR #536, opened 2018-09-10
make the padding parameter order the same as in the Android Framework

martinbonnin:
see discussion in https://github.com/airbnb/epoxy/issues/535

### DIFF
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/Carousel.java
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/Carousel.java
@@ -333,10 +333,10 @@ public class Carousel extends EpoxyRecyclerView {
    * @see #setPadding(Padding)
    */
   public static class Padding {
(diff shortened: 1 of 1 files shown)


elihart:
I'm not sure why I ordered it the way I did, but I see your point about changing it.

I'm hesitant though because it won't be a compiler breaking change, it will just have different runtime behavior, so it would be really easy for people to forget to update.

elihart:
I think we can include this with 3.0 - but I would also like to change the main Padding constructor for pixels to be `Padding.px(...)` to match the new dp and resource support. this will make it more likely that people update correctly too I think

martinbonnin:
`Padding.px` pushed. If we really wanted to make sure everyone updates their code, we could rename all these static methods:

* `Padding.fromPx()`, `Padding.fromDp()`
* or `Padding.createPx()`, `Padding.createDp()`

Issue #414, opened 2018-04-03
Allow a hash key to be specify with @CallbackProp


elihart:
@Callback prop uses `DoNotHash` so that listeners are not constantly rebound. However, if state is captured in the listener lambda it can become outdated and cause the wrong behavior if the model is updated.

This is only an issue if the callback relies on model state that can change (which is generally not recommended), but sometimes it is hard to avoid.

To help in this case we can have Epoxy support an optional object key as another parameter in the generated callback setter. The hash of this key can dictate when the callback is rebound to the view.

NoOneNo:
mark

Issue #502, opened 2018-08-21
Data binding not updating listener variables on bind


chrisbanes:
I wanted to open this issue since I don't really understand why the some of the data binding `setDataBindingVariables()` logic is implemented as:

``` java
if (((popupMenuListener == null) != (that.popupMenuListener == null))) {
  binding.setVariable(BR.popupMenuListener, popupMenuListener);
}
```

I'm guessing it is implicitly using `EpoxyAttribute(NoHash)` on my behalf.

Anyway my issue is that the listener is never updated on rebinds. Surely if a variable is set, it should be set on the binding. If the items are equals/==hashCode then sure don't update the binding, but the default so always be to set the binding.

At the moment it only updates the binding if the value goes from null to non-null and vice-versa, which is just confusing IMO.

elihart:
Right, it's automatically applying [DoNotHash](https://github.com/airbnb/epoxy/blob/3aa26934eec27d12449c504ad3b58bcade0c0597/epoxy-annotations/src/main/java/com/airbnb/epoxy/EpoxyAttribute.java#L46)

I understand the confusion,  but imo the alternative is worse. If you have a bunch of models, and they all have anonymous click listeners, then every time you rebuild models the differ would think everything changed.

I think this expected behavior will differ depending on your use cases though. For me, generally we have immutable data and our click listener is never going to change behavior. If you do have mutable data this can be confusing and make life a bit harder.

This is also why I created the [OnModelClickListener](https://github.com/airbnb/epoxy/wiki/Epoxy-Models#click-listeners), so even though the click listener isn't rebound it always has access to the most recently created model and can get any data it needs off of it (similar to https://github.com/airbnb/epoxy/issues/499)

Databinding is the one place you don't have control over this setting though, and I would like it to be a choice. I think we could pretty easily modify the databinding annotations for providing layouts to add a boolean setting to disable this DoNotHash behavior by default

chrisbanes:
Interesting. Your mention of _anonymous_ click listeners has made me think about another possible solution.

I could create a non-anonymous listener class which has a mutable property. I can then set an `onBind` listener on the model which update the listener instance.

elihart:
Sure, that would work I think. That's basically what the `OnModelClickListener` is already doing

chrisbanes:
Cool, this isn't an `OnClickListener` so I can't use that :)

elihart:
[...]

Issue #499, opened 2018-08-15
Get actual entity inside click listener callback.


ramanbranavitski:
Hi. Spent a lot of time to find a way to get clicked entity inside clickListener callback.
Have the following code

```
@Override
    protected void buildModels(final List<Item> items) {

        for (Item item : items) {
            new ListItemViewModel()
                    .id(item.getId())
                    .name(item.getName())
                    .clickListener()
                    .addTo(this);
        }
    }

   public interface OnItemClickListener {
        void onItemClick(Item clickedItem);
    }
```
 I've tried to use just item from cycle, but as I noticed this entity cached somewhere and if it changes it's state I still receive "old" entity with not actual fields. Is there any way to receive clicked item? Thanks.

elihart:
Have you looked at `OnModelClickListener` and https://github.com/airbnb/epoxy/wiki/Epoxy-Models#click-listeners

Normally epoxy doesn't rebind properties that don't change - that is an optimization that is documented in the wiki. The click listener is a special case though that will always have the most recently created model.

How are you creating your model? The rebinding changes based on the implementation

ramanbranavitski:
Yes, I have. But it's not very convenient to calculate proper  if you're working with 2 different collections of data(let it be `List<Type1>` and `List<Type2>`) and headers for each one. If you have 1 element int these 2 arrays and click on the last item you will receive `position=3` from `OnModelClickListener` but it's the firs element of `List<Type2>`. Right now I'm using this approach as I haven't found any other working solutions. My model is 

```
@ModelView(autoLayout = ModelView.Size.MATCH_WIDTH_WRAP_HEIGHT)
public class ReminderListItemView extends ConstraintLayout {

    @BindView(R.id.tv_reminder_message) TextView tvMessage;
    @BindView(R.id.tv_reminder_name) TextView tvReminderName;

    public ReminderListItemView(final Context context) {
        super(context);
        init();
    }

    private void init() {
        inflate(getContext(), R.layout.item_reminder, this);
        ButterKnife.bind(this);
    }

    @TextProp
    public void setMessage(@Nullable final CharSequence message) {
        tvMessage.setText(message);
    }

    @TextProp
    public void setReminderName(@Nullable final CharSequence name) {
        if (TextUtils.isEmpty(name)) {
            tvReminderName.setText(R.string.reminder);
        } else {
            tvReminderName.setText(name);
        }
    }

    @CallbackProp
[...]

PR #540, opened 2018-09-13
New system for keying listeners

elihart:
There has been a lot of issues and confusion, both within airbnb and externally, because of the `DoNotHash` option and how to deal with callbacks that capture state that may change.

This PR takes the approach suggested in this issue: https://github.com/airbnb/epoxy/issues/414

There is a `KeyedListener` class that wraps any kind of callback, and implements equals/hashcode with the value of a given key.

To make this easy to use I've written several helper functions in kotlin - java would make dealing with the generics a big pain, and kotlin also lets us do some nice things with reified type checking and extension functions. This is the first time I'm adding kotlin to the core module, but I think it's about time :P 

Note, I haven't written much code comments yet to avoid wasting time if the design changes, so consider it a WIP.

Usage for creating the listeners looks like this:
```kotlin
CarouselModel_().apply {

        // Click listener keyed on model
        onClickListener(keyedClickListener {

        })

        // Long click listener keyed on model
        keyedLongClickListener {
            return@keyedLongClickListener false
        }

        // Custom listener type, keyed on model
        keyedListener {
            4
        }

        // Custom listener type, keyed on user defined key
        keyedListener("custom key") {

        }

        // Custom interface callback, keyed on custom key
        keyedListener("some key", object : MyCustomInterface {
            override fun doSomething(view: View) {

                // Get details on the model and view
                val (model, parentView, clickedView, position) = clickContext(view) ?: return
            }
        })
    }
```

**High level points**
- Helpers like `keyedClickListener` create a click listener that is keyed on the model value, excluding listeners. So if the model changes at all the click listener is considered different and will be updated
- There are other helpers for using custom lambdas or interfaces as keyed callbacks.
- A custom key can be used if you don't want to key on the model
- Keyed listeners can access the `ClickContext` to get the most up to date model object, view, and adapter position. This can be accessed via the clicked view, or for default click listeners it is the lambda receiver for easy access
- I've updated the `@CallbackProp` annotation to only apply DoNotHash behavior if the type does not have equals/hashcode. This means you can use a KeyedListener with CallbackProp and it will be updated as expected

**Issues**
- This adds another way to handle click listeners, on top of the previous `OnModelClickListener` support. This is a bit confusing.
- If `@CallbackProp` is used you still need to explicitly define the listener type as KeyedClickListener for the hashing to work (same for databinding)
- The extension functions currently operate on the EpoxyModel as a receiver, so they won't work with the model builder extension functions. I would need to generate other methods for those to work

In general there is a lot of knowledge and overhead required to choose the right option and make sure your code has the right behavior. I think the only way to solve this is with a fairly drastic breaking change of how the generated models use listeners - that is maybe possible in the future, but this seems like a good step for now.

@ngsilverman @chrisbanes

### DIFF
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/KeyedListeners.kt
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/KeyedListeners.kt
@@ -1,0 +1,163 @@
+package com.airbnb.epoxy
+
+import android.support.v7.widget.RecyclerView
+import android.util.Log
+import android.view.View
+import android.widget.CompoundButton
+
+// todo only apply DONotHash in callback prop if equals/hashcode does not exist on type?
+// Big refactor to remove DoNotHash completely and enforce keyed listeners everywhere?
+// Only apply if a build flag is passed?
+
+// Generate a kotlin extension function on builder interfaces for easily passing keyed listeners?
+//inline fun CarouselModelBuilder.onClickListener(crossinline onClick: ClickContext<View, Carousel, CarouselModel_>.() -> Unit) {
+//    onClickListener((this as CarouselModel_).keyedClickListener(onClick))
+//}
+
+// Generate keyed subclass of any type annotated with callback prop?
+
+inline fun <reified V : View, reified P : Any, reified T : EpoxyModel<P>> T.clickContext(
+    clickedView: V
+): ClickContext<V, P, T>? = ClickContext.from(clickedView)
+
+inline fun <reified V : Any, reified T : EpoxyModel<V>, Listener> T.keyedListener(
+    key: Any = this,
+    listener: Listener
+) = KeyedListener(key, listener)
+
+inline fun <reified V : Any, reified T : EpoxyModel<V>> T.keyedLongClickListener(
+    crossinline listener: ClickContext<View, V, T>.() -> Boolean
+) = KeyedLongClickListener(this, View.OnLongClickListener { v ->
+    val context = ClickContext.from<View, V, T>(v) ?: return@OnLongClickListener false
+    return@OnLongClickListener context.listener()
+})
+
+class KeyedLongClickListener(
+    key: Any,
+    listener: View.OnLongClickListener
+) : KeyedListener<View.OnLongClickListener>(key, listener), View.OnLongClickListener {
+
+    override fun onLongClick(v: View) = listener.onLongClick(v)
+}
+
+inline fun <reified V : Any, reified T : EpoxyModel<V>> T.keyedCheckedChangeListener(
+    crossinline listener: ClickContext<CompoundButton, V, T>.(isChecked: Boolean) -> Unit
+) = KeyedCheckedChangedListener(this, CompoundButton.OnCheckedChangeListener { v, isChecked ->
+    ClickContext.from<CompoundButton, V, T>(v)?.apply {
+        listener(isChecked)
+    }
+})
+
+class KeyedCheckedChangedListener(
+    key: Any,
+    listener: CompoundButton.OnCheckedChangeListener
+) : KeyedListener<CompoundButton.OnCheckedChangeListener>(key, listener),
+    CompoundButton.OnCheckedChangeListener {
+
+    override fun onCheckedChanged(buttonView: CompoundButton, isChecked: Boolean) {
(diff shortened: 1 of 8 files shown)


epoxy-adapter/src/main/java/com/airbnb/epoxy/KeyedListeners.kt
```diff
@@ -0,0 +1,163 @@
...
+        model?.inKeyedListenerEquals = true
+
+        val result = key.hashCode()
+
+        model?.inKeyedListenerEquals = false
+        return result
+    }
+}
+
+data class ClickContext<V : View, P : Any, T : EpoxyModel<P>>(
```
ngsilverman: The name "ClickContext" confused me a bit at first, until I understood that it's not directly related to clicks. It's just context about a view, and could be used at any time, including when non-click events happen, right?

epoxy-adapter/src/main/java/com/airbnb/epoxy/KeyedListeners.kt
```diff
@@ -0,0 +1,163 @@
...
+inline fun <reified V : Any, reified T : EpoxyModel<V>> T.keyedClickListener(
+    crossinline listener: ClickContext<View, V, T>.() -> Unit
+) = KeyedClickListener(this, View.OnClickListener { v ->
+    ClickContext.from<View, V, T>(v)?.apply(listener)
+})
+
+class KeyedClickListener(
+    key: Any,
+    listener: View.OnClickListener
+) : KeyedListener<View.OnClickListener>(key, listener), View.OnClickListener {
```
ngsilverman: Delegation could help here, and for other keyed listener types:
```kotlin
class KeyedClickListener(
    key: Any,
    listener: OnClickListener
) : KeyedListener<OnClickListener>(key, listener), OnClickListener by listener
```

ngsilverman:
@elihart an additional thought: it seems like we could solve this nicely if we knew which values are captured by the listener closures. Since I don't think we can get them automatically, maybe we could make those the keys:
```kotlin
CarouselModel_().apply {
    onClickListener(keyedClickListener(listOf(stateA, stateB) {
        // Does something with stateA and stateB.
    }))
}
```

elihart:
@ngsilverman thanks for looking, I guess I should give more background

> When using a listener keyed on the model itself, wouldn't it only be rebound if another model property changes? In which case it would have been rebound anyway.

The problem is that epoxy only rebinds properties that change, so if another property on the model changes it gets rebound, but not the click listener - in the current system the only way to have the listener rebound is if the hashcode changes, which doesn't happen because of DoNotHash - or otherwise it would always be rebound for anonymous click listeners.

Issues have been brought up like 
https://github.com/airbnb/epoxy/issues/502
https://github.com/airbnb/epoxy/issues/499

and someone at airbnb just messaged me wondering why their code wasn't working and they were doing this
```
onClickListener { v ->
                MvRxActivity.startActivity<UserProfileFragment>(v.context, UserProfileArguments(state.userId))
            }
```

the userid was changing and the click listener had captured the old one.

Another example is if you have two different listener objects and you want to swap between the two, that's hard to do right now.

It just occurred to me that another approach could be to have epoxy's internal binding rebind @CallbackProps if any non callbackprop changed - same idea as I have now in the keyed listener, but it would be more built in. However, there are still subtle cases, like I posted above, where the model state doesn't change but we still need the click listener to change, which is really hard to automatically detect and fix.

I'm still searching for a perfect solution that catches all cases and isn't confusing - not sure if that is a white whale though

epoxy-adapter/src/main/java/com/airbnb/epoxy/KeyedListeners.kt
```diff
@@ -0,0 +1,163 @@
...
+
+    override fun onClick(v: View) = listener.onClick(v)
+}
+
+open class KeyedListener<Listener>(private val key: Any, val listener: Listener) {
+
+    // Only include the key, and not the listener, in equals/hashcode
+    override fun equals(other: Any?): Boolean {
+        val model = key as? EpoxyModel<*>
+        if (model?.inKeyedListenerEquals == true) return true
```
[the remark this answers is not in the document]
ngsilverman: @elihart could this cause issues if `equals` runs concurrently?

PR #560, opened 2018-10-02
Implements visibility events for EpoxyRecyclerView/EpoxyViewHolder.

eboudrant:
Visibility events in Epoxy Inspired from Litho : https://fblitho.com/docs/visibility-handling
- **Visible Event**: this event is triggered when at least one pixel of the Component is visible.
- **Invisible Event**: this event is triggered when the Component no longer has any pixels on the screen.
- **Focused Visible Event**: this event is triggered when either the Component occupies at least half of the viewport, or, if the Component is smaller than half the viewport, when it is fully visible.
- **Unfocused Visible Event**: this event is triggered when the Component is no longer focused, i.e. it is not fully visible and does not occupy at least half the viewport.
- **Full Impression Visible Event**: this event is triggered when the entire Component has passed through the viewport at some point.
- **VisibilityChangedEvent**: this event is triggered when the visibility of the Component on the screen changes.

As this require to add few listener on the `RecyclerView` the trqacking is opt-in :

```
val epoxyVisibilityTracker = EpoxyVisibilityTracker()
epoxyVisibilityTracker.attach(recyclerView)
```

Works with `EpoxyModelWithHolder`

```
@EpoxyModelClass(layout = R.layout.model_color)
public abstract class ColorModel extends EpoxyModelWithHolder<ColorHolder> {
  ...
  @Override
  public void visibilityStateChanged(@VisibilityState int visibilityState, @NonNull T view) { ... }

  @Override
  public void visibilityChanged(float visibleHeight, float visibleWidth, int percentVisibleHeight,
      int percentVisibleWidth, @NonNull T view) { ... }
}
```

And with annotations on the custom view :

```
@ModelView(autoLayout = ModelView.Size.MATCH_WIDTH_WRAP_HEIGHT)
class ItemCustomView @JvmOverloads constructor(context: Context : LinearLayout(context) {
    ...
    @OnVisibilityStateChanged
    fun onVisibilityStateChanged(
        @OnModelVisibilityStateChangedListener.VisibilityState visibilityState: Int
    ) { ... }
    }

    @OnVisibilityChanged
    fun onVisibilityChanged(
      percentVisibleHeight: Float,
      percentVisibleWidth: Float,
      visibleHeight: Int,
      visibleWidth: Int
    ) { ... }
}
```
Or from the model via : `OnModelVisibilityStateChangedListener` / `OnModelVisibilityChangedListener` :

```
anyItem {
  id("$id")
  onVisibilityStateChanged { model, view, visibilityState -> ... }
  ...
}
```

Unit testing for the visibility tracker can be found in `EpoxyVisibilityTrackerTest.kt`


elihart:
This is cool, thanks for doing all this work - it definitely fills a good use case. I'll try to make time to review more closely later this week

eboudrant:
> This is cool, thanks for doing all this work - it definitely fills a good use case. I'll try to make time to review more closely later this week

Ok, @elihart take your time. I think I need to push another change today about having a `OnModelVisibilityChangedListener`... similar to `OnModelBoundListener`.

epoxy-adapter/src/main/java/com/airbnb/epoxy/OnModelVisibilityStateChangedListener.java
```diff
@@ -0,0 +1,58 @@
...
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+/** Used to register an onVisibilityChanged callback with a generated model. */
+public interface OnModelVisibilityStateChangedListener<T extends EpoxyModel<?>, V> {
+
+  @Retention(RetentionPolicy.SOURCE)
+  @IntDef({VISIBLE, INVISIBLE, FOCUSED_VISIBLE, UNFOCUSED_VISIBLE, FULL_IMPRESSION_VISIBLE})
+  @interface VisibilityState {
```
elihart: nesting the VisibilityState interface inside OnModelVisibilityStateChangedListener makes it quite verbose to reference state's like `VISIBLE`. what do you think of moving the `VisibilityState` interface to be top level?

I want to be careful that the naming isn't too generic to have it mixed up with other names people might use in their apps - maybe `EpoxyVisibility` or `ModelVisibility` could be used

epoxy-adapter/src/main/java/com/airbnb/epoxy/OnModelVisibilityStateChangedListener.java
```diff
@@ -0,0 +1,58 @@
+package com.airbnb.epoxy;
+
+import android.support.annotation.IntDef;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+/** Used to register an onVisibilityChanged callback with a generated model. */
+public interface OnModelVisibilityStateChangedListener<T extends EpoxyModel<?>, V> {
```
elihart: `EpoxyModel<V>`

epoxy-annotations/src/main/java/com/airbnb/epoxy/OnVisibilityChanged.java
```diff
@@ -0,0 +1,18 @@
+package com.airbnb.epoxy;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * This can be used to annotate methods inside classes with a {@link ModelView} annotation. Methods
+ * with this annotation will be called when the viewport is
```
elihart: I'm not clear on what changes to the viewport mean in this context. can this documentation link to the epoxy model method that this maps to?

would be nice to update the link to point to epoxy specific documentation when we have it

epoxy-annotations/src/main/java/com/airbnb/epoxy/OnVisibilityStateChanged.java
```diff
@@ -0,0 +1,20 @@
+package com.airbnb.epoxy;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * This can be used to annotate methods inside classes with a {@link ModelView} annotation. Methods
```
elihart: can this documentation link to the epoxy model method that this maps to?

elihart: Seems like we should say what parameters a method with this annotation can have?

epoxy-processor/src/main/java/com/airbnb/epoxy/GeneratedModelWriter.kt
```diff
@@ -435,6 +484,74 @@ internal class GeneratedModelWriter(
...
+
+        val onVisibilityStateChanged = MethodSpec.methodBuilder("onVisibilityStateChanged")
+            .addJavadoc(
+                "Register a listener that will be called when this model visibility state "
+                        + "has changed.\n"
+                        + "<p>\n"
+                        + "The listener will contribute to this model's hashCode state per the {@link\n"
+                        + "com.airbnb.epoxy.EpoxyAttribute.Option#DoNotHash} rules.\n"
+                        + "<p>\n"
+                        + "You may clear the listener by setting a null value, or by calling "
```
elihart: let's remove `You may clear the listener by setting a null value, or by calling {@link #reset()} */ `mutable models are not a pattern we encourage, and is the wrong advice for most people since so few use EpoxyAdapter

epoxy-processor/src/main/java/com/airbnb/epoxy/GeneratedModelWriter.kt
```diff
@@ -435,6 +484,74 @@ internal class GeneratedModelWriter(
...
+
+        val onVisibilityChanged = MethodSpec.methodBuilder("onVisibilityChanged")
+            .addJavadoc(
+                "Register a listener that will be called when this model visibility has "
+                        + "changed.\n"
+                        + "<p>\n"
+                        + "The listener will contribute to this model's hashCode state per the {@link\n"
+                        + "com.airbnb.epoxy.EpoxyAttribute.Option#DoNotHash} rules.\n"
+                        + "<p>\n"
+                        + "You may clear the listener by setting a null value, or by calling "
```
elihart: same here

epoxy-processor/src/main/java/com/airbnb/epoxy/ModelViewProcessor.kt
```diff
@@ -354,7 +408,13 @@ internal class ModelViewProcessor(
...
 
     private fun validateResetElement(resetMethod: Element): Boolean =
-            validateExecutableElement(resetMethod, OnViewRecycled::class.java, 0)
+        validateExecutableElement(resetMethod, OnViewRecycled::class.java, 0)
+
+    private fun validateVisibilityStateChangedElement(visibilityMethod: Element): Boolean =
+        validateExecutableElement(visibilityMethod, OnVisibilityStateChanged::class.java, 1)
+
+    private fun validateVisibilityChangedElement(visibilityMethod: Element): Boolean =
+        validateExecutableElement(visibilityMethod, OnVisibilityChanged::class.java, 4)
```
elihart: `validateExecutableElement` was previously only used for parameters that could be anything, so only the count mattered, for this it is important that we give a clear warning if the types don't match the expected. Can you change `validateExecutableElement` to take an optional list of parameter types, and emit an error if the types don't match?

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModel.java
```diff
@@ -155,6 +156,13 @@ public void bind(@NonNull T view, @NonNull EpoxyModel<?> previouslyBoundModel) {
   public void unbind(@NonNull T view) {
   }
 
+  public void visibilityStateChanged(@VisibilityState int visibilityState, @NonNull T view) {
```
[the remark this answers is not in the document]
eboudrant: I was not sure so I used the same naming convention than the `bind(...)` / `unbind(...)` methods in the same class.

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityItem.java
```diff
@@ -0,0 +1,157 @@
...
+class EpoxyVisibilityItem {
+
+  private final Rect localVisibleRect = new Rect();
+
+  private int adapterPosition = RecyclerView.NO_POSITION;
+
+  @Px
+  private int sizeInScrollingDirection;
+
+  private int otherSize;
```
josephyanks: It might be nice for this to be a more descriptive name

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    if (vi.getAdapterPosition() != epoxyHolder.getAdapterPosition()) {
+      // EpoxyVisibilityItem being re-used for a different position
+      vi.reset(epoxyHolder.getAdapterPosition());
+    }
+
+    if (vi.update(itemView, recyclerView, vertical)) {
+      // View is measured, process events
+      vi.handleVisible(epoxyHolder);
```
josephyanks: It seems inefficient here to have to call every single one of these handle methods. It looks like it is possible to be both visible and invisible at the same time here (if detach event = true), which feels like a bug. Perhaps you should let detach events "win" and not fire their counterpart (e.g. only fire handleInvisible if the detach event is true).

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.java
```diff
@@ -0,0 +1,146 @@
...
+        @NonNull View recyclerView,
+        int left, int top, int right, int bottom,
+        int oldLeft, int oldTop, int oldRight, int oldBottom
+    ) {
+      processChildren();
+    }
+
+    @Override
+    public void onScrolled(@NonNull RecyclerView recyclerView, int dx, int dy) {
+      processChildren();
```
josephyanks: Is this efficient in large lists? Unsure if you maybe want to debounce calling this with every scroll

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModel.java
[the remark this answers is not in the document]
eboudrant: Changed to `onVisibility...`

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityItem.java
```diff
@@ -0,0 +1,156 @@
...
+
+  /** Store last value for deduping */
+  private int lastVisibleSizeNotified = -1;
+
+  /**
+   * Update the visibility item according the current layout.
+   *
+   * @param view        the current {@link com.airbnb.epoxy.EpoxyViewHolder}'s itemView
+   * @param parent      the {@link android.support.v7.widget.RecyclerView}
+   * @param orientation the {@link android.support.v7.widget.LinearLayoutManager}'s orientation
```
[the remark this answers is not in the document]
eboudrant: Done, we can actually use `canScrollVertically()` on `LayoutManager`. 

Just understand that the visibility tracker is not really working yet when it can scroll in both direction (ex with a model in a carousel nested in a vertical list). And I would prefer address it in a different pull request. We could mention it in the documentation in the meantime.

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.java
```diff
@@ -0,0 +1,164 @@
...
+import android.support.v7.widget.RecyclerView.OnScrollListener;
+import android.support.v7.widget.RecyclerView.ViewHolder;
+import android.util.SparseArray;
+import android.view.View;
+import android.view.View.OnLayoutChangeListener;
+
+/**
+ * Helper class to handle visibility event from an {@link com.airbnb.epoxy.EpoxyRecyclerView}.
+ */
+class EpoxyVisibilityTracker {
```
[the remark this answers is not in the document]
eboudrant: done

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.java
josephyanks: Sorry, to be clearer about my comment, it seems like a possible source of bugs that both visible and invisible could be true if detach = true. It seems like it should be 

```
if(!vi.handleVisible(eh)) { vi.handleInvisible(eh) } 
```

or inverted, since invisible is the only state of those two that takes detach into account. If you don't do this, then the callbacks could emit a sequential visible + invisible event.

eboudrant:
@elihart about Kotlin, I could re-write. I did not used Kotlin because client code was all in Java (actually I started in Kotlin, then realized that so converted to Java ;) ). 

However I think before moving to Kotlin we should add then same level of checking than Java hava with checkstyle. I don't think the project use `kltlint` ?

epoxy-adapter/src/test/java/com/airbnb/epoxy/EpoxyVisibilityTrackerTest.kt
```diff
@@ -0,0 +1,535 @@
...
+
+/**
+ * This class test the EpoxyVisibilityTracker by using a RecyclerView that scroll vertically. The
+ * view port height is provided by Robolectric.
+ *
+ * We are just controlling how many items are displayed with VISIBLE_ITEMS constant.
+ */
+@Config(sdk = [21], manifest = TestRunner.MANIFEST_PATH)
+@RunWith(TestRunner::class)
+class EpoxyVisibilityTrackerTest {
```
eboudrant: I just added unit tests (in Kotlin). @elihart please review and let me know what you think. I think theses are the first tests in the lib to assert on actual UI (`RecyclerView`). Btw, I was able to fix few bugs related to use case where `scrollToPosition(...)` is called.

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModel.java
```diff
@@ -155,6 +157,18 @@ public void bind(@NonNull T view, @NonNull EpoxyModel<?> previouslyBoundModel) {
   public void unbind(@NonNull T view) {
   }
 
+  public void onVisibilityStateChanged(@Visibility int visibilityState, @NonNull T view) {
```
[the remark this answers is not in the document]
eboudrant: I added lint to the related annotation and todo for link to wiki

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.java
```diff
@@ -0,0 +1,172 @@
...
+    final int id = System.identityHashCode(itemView);
+
+    EpoxyVisibilityItem vi = visibilityIdToItemMap.get(id);
+    if (vi == null) {
+      vi = new EpoxyVisibilityItem();
+      visibilityIdToItemMap.put(id, vi);
+    }
+
+    if (vi.getAdapterPosition() != epoxyHolder.getAdapterPosition()) {
+      // EpoxyVisibilityItem being re-used for a different position
```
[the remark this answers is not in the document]
eboudrant: Thanks, I will revisit the tracker. Regarding the bug you are absolutely right, I missed it. I was able to reproduce the bug via unit test (disabled for now, I will look for a later).

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.java
```diff
@@ -0,0 +1,172 @@
...
+
+/**
+ * A simple way to track visibility events on {@link com.airbnb.epoxy.EpoxyModel} within a {@link
+ * android.support.v7.widget.RecyclerView}.
+ * <p>
+ * {@link EpoxyVisibilityTracker} works with any {@link android.support.v7.widget.RecyclerView}
+ * backed by an Epoxy controller. Once attached the events will be forwarded to the Epoxy model (or
+ * to the Epoxy view when using annotations).
+ * <p>
+ * Note regarding nested lists: The visibility event tracking is not properly handled yet. This is
```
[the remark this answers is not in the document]
eboudrant: Automatically attach yes and if, I remember well, any nested recycler view will have to listen to its parent recycler view event (bind, attach, scroll). I planned to address it a separate PR as it should not impact code gen.

epoxy-processortest/src/test/java/com/airbnb/epoxy/ViewProcessorTest.kt
```diff
@@ -109,6 +109,76 @@ class ViewProcessorTest {
...
+            "must have parameter types [INT], found: [BOOLEAN] (method: onVisibilityStateChanged)")
+    }
+
+    @Test
+    fun onVisibilityStateChanged_throwsIfNoParams() {
+        assertGenerationError(
+            "OnVisibilityStateChangedView_throwsIfNoParams.java",
+            "must have exactly 1 parameter (method: onVisibilityStateChanged)")
+    }
+
```
eboudrant: Unit test for visibility annotations on @ModelView added.

PR #575, opened 2018-10-15
Follow-up for visibility tracking : react to adapter data changes

eboudrant:
Follow-up for visibility tracking (#560). This fix the case where item get added, removed or moved from the adapter. For doing it it require to add a `AdapterDataObserver` on the current adapter. The idea is to shift the adapter position maintained in the visibility item.

This chance activate `testInsertData` / `testDeleteData` tests  and add  a new test : `testMoveData`.

### DIFF
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.java
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.java
@@ -8,6 +8,7 @@ import android.support.v7.widget.RecyclerView.AdapterDataObserver;
 import android.support.v7.widget.RecyclerView.OnChildAttachStateChangeListener;
 import android.support.v7.widget.RecyclerView.OnScrollListener;
 import android.support.v7.widget.RecyclerView.ViewHolder;
+import android.util.Log;
 import android.util.SparseArray;
 import android.view.View;
 import android.view.View.OnLayoutChangeListener;
@@ -33,6 +34,10 @@ import java.util.List;
  * @see OnModelVisibilityStateChangedListener
  */
 public class EpoxyVisibilityTracker {
+
+  private static final String TAG = "EpoxyVisibilityTracker";
+
+  static final boolean DEBUG_LOG = true;
 
   /** Maintain visibility item indexed by view id (identity hashcode) */
   private final SparseArray<EpoxyVisibilityItem> visibilityIdToItemMap = new SparseArray<>();
@@ -77,7 +82,7 @@ public class EpoxyVisibilityTracker {
   /**
    * Detach the tracker
    *
-   * @param recyclerView The recyclerview that the EpoxyController has its adapter added to.
+   * @param recyclerView The recycler view that the EpoxyController has its adapter added to.
    */
   public void detach(@NonNull RecyclerView recyclerView) {
     recyclerView.removeOnScrollListener(this.listener);
@@ -95,13 +100,24 @@ public class EpoxyVisibilityTracker {
   }
 
   private void processChildren(@Nullable View detachedView, String debug) {
-    processNewAdapterInNecessary();
+
+    // On every every events lookup for a new adapter
+    processNewAdapterIfNecessary();
+
+    // Process the detached child if any
+    if (detachedView != null) {
+      processChild(detachedView, true, debug);
+    }
+
+    // Process all attached children
     final RecyclerView recyclerView = attachedRecyclerView;
     if (recyclerView != null) {
       for (int i = 0; i < recyclerView.getChildCount(); i++) {
         final View child = recyclerView.getChildAt(i);
-        if (child != null) {
-          processChild(child, child == detachedView, debug);
+        if (child != null && child != detachedView) {
+          // Is some case the detached child is still in the recycler view. Don't process it as it
+          // was already processed.
+          processChild(child, false, debug);
         }
       }
     }
@@ -111,7 +127,7 @@ public class EpoxyVisibilityTracker {
    * If there is a new adapter on the attached RecyclerView it will resister the data observer and
    * clear the current visibility states
    */
-  private void processNewAdapterInNecessary() {
+  private void processNewAdapterIfNecessary() {
     if (attachedRecyclerView != null && attachedRecyclerView.getAdapter() != null) {
       if (lastAdapterSeen != attachedRecyclerView.getAdapter()) {
         if (lastAdapterSeen != null) {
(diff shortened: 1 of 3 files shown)


epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.java
```diff
@@ -74,47 +92,114 @@ public void detach(@NonNull RecyclerView recyclerView) {
...
+      if (lastAdapterSeen != attachedRecyclerView.getAdapter()) {
+        if (lastAdapterSeen != null) {
+          // Unregister the old adapter
+          lastAdapterSeen.unregisterAdapterDataObserver(this.observer);
+        }
+        // Register the new adapter
+        attachedRecyclerView.getAdapter().registerAdapterDataObserver(this.observer);
+        lastAdapterSeen = attachedRecyclerView.getAdapter();
+        // Clear our visibility items
+        visibilityIdToItemMap.clear();
```
elihart: it may not be necessary to clear the maps - if `swapAdapter` is used on the recyclerview then the views can be reused with a new adapter.

we can't know if this is the case. the downside to not clearing them is possibly wasted memory (maybe leaks?) and longer lookups, but it might also prevent duplicate visibility events if views are reused and stay on screen

i feel like we should leave it up to the user to tell us that they want to clear view info somehow, since that's the only real way we can know the intention

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.java
```diff
@@ -155,22 +241,88 @@ public void onLayoutChange(
...
+    }
+
+    @Override
+    public void onItemRangeMoved(int fromPosition, int toPosition, int itemCount) {
+      if (DEBUG_LOG) {
+        Log.d(TAG,
+            String.format("onItemRangeMoved(%d, %d, %d)", fromPosition, toPosition, itemCount));
+      }
+      for (EpoxyVisibilityItem item : visibilityIdToItems) {
+        int position = item.getAdapterPosition();
```
elihart: other items can be affected too. moving an item from position 0 to 10 shifts items 1-9 up by one

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.java
```diff
@@ -251,13 +252,19 @@ public void onScrolled(@NonNull RecyclerView recyclerView, int dx, int dy) {
...
     public void onChildViewAttachedToWindow(View child) {
-      processChangeEvent("onChildViewAttachedToWindow");
+      processChild(child, false, "onChildViewAttachedToWindow");
     }
 
     @Override
     public void onChildViewDetachedFromWindow(View child) {
-      // On detach event send the detached view
-      processChangeEventWithDetachedView(child, "onChildViewDetachedFromWindow");
+      if (dataChanged) {
```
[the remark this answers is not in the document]
eboudrant: That's a good point, I could probably check the bounds from the visible items before assigning `dataChanged`.

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.java
```diff
@@ -74,47 +96,111 @@ public void detach(@NonNull RecyclerView recyclerView) {
...
+          // Unregister the old adapter
+          lastAdapterSeen.unregisterAdapterDataObserver(this.observer);
+        }
+        // Register the new adapter
+        attachedRecyclerView.getAdapter().registerAdapterDataObserver(this.observer);
+        lastAdapterSeen = attachedRecyclerView.getAdapter();
+        // Clear our visibility items
+        visibilityIdToItemMap.clear();
+        visibilityIdToItems.clear();
+        System.out.println("attached on " + lastAdapterSeen);
```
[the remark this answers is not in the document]
eboudrant: oops, forgot that one

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.java
eboudrant: How would you let the user in control of that ? Do you propose explicit `clear()` in the public api ?

I am not a fan of the way it discover a new adapter but I could not find a better way. And I doubt I can detect if the new adapter come from a call to `swapAdapter `

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.java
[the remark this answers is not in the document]
elihart: that seems reasonable.

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.java
elihart: Yes, I think an explicit `clear` method could work. It isn't too important though, maybe as a followup? I feel like it's better to give control than to potentially break visibility events for swapped adapters.

I don't think there is a better way to detect a new adapter, so what you're doing seems fine.

eboudrant: Ok, make sense. I added `clearVisibilityStates()`.

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.java
```diff
@@ -74,60 +96,133 @@ public void detach(@NonNull RecyclerView recyclerView) {
...
+   * {@link EpoxyVisibilityTracker#processVisibilityEvents}
+   *
+   * @param child               the view to process for visibility event
+   * @param detachEvent         true if the child was just detached
+   * @param eventOriginForDebug a debug strings used for logs
+   */
+  private void processChild(@NonNull View child, boolean detachEvent, String eventOriginForDebug) {
     final RecyclerView recyclerView = attachedRecyclerView;
     if (recyclerView != null) {
       recyclerView.getChildViewHolder(child);
```
elihart: `recyclerView.getChildViewHolder(child);` on line 169 seems to be a mistake

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.java
```diff
@@ -155,22 +250,126 @@ public void onLayoutChange(
...
     @Override
     public void onChildViewAttachedToWindow(View child) {
-      processChild(child);
+      processChild(child, false, "onChildViewAttachedToWindow");
     }
 
     @Override
     public void onChildViewDetachedFromWindow(View child) {
-      processChild(child, true);
+      if (visibleDataChanged) {
```
elihart: @eboudrant It seems like it would be possible for a multi item data change (eg several items removed) to result in several views being detached - this would only account for the first one though.

to simplify things, what if we always processed all children when a view is detached? i'm a bit concerned about the complexity in trying to optimize this, and that we won't be able to correctly handle all cases.

I'm not too familiar with how/when this is called, and it's interaction with the other callbacks, so I'll leave it up to you.

eboudrant: With _**always processed all children**_ I am just afraid of scrolling where views detach/re-attach often. I am pretty confident with the `visibleDataChanged` flag.

Issue #583, opened 2018-10-18
[epoxy-paging] Add way of clearing the model cache


kakai248:
I have a paged controller with data that comes from my database. The screen where this controller is used allows selecting multiple items. In the ViewModel I keep a set of items that are selected. To be able to diff properly when an item is selected/deselected, I keep a boolean in the object that I pass to the epoxy model. Each time an item is selected/deselected, I do `requestModelBuild()`. Since all the items are cached, there are no changes.

Model cache only reacts to updates in the PagedList, which I have no control of. It would be good if we were able to manually clear the cache before requesting another model build.

elihart:
This seems reasonable - I'd be open to a pull request if you want to attempt it yourself (or anyone else)

kakai248:
I'll try to open the PR during the weekend. Unless someone wants to do it first.

I was thinking of adding a method in the model cache that nulls all the values and a method in the controller that calls the cache method and then requests a model build (something like `requestForcedModelBuild()`, in lack of a better name).

What do you think?

elihart:
that sounds reasonable to me :)

yigit would know best, since he originally wrote it, but he probably doesn't have much time to help maintain and expand this.

baoti:
It looks like #634. You can custom `itemDiffCallback`:

```
override fun areContentsTheSame(oldItem: Any, newItem: Any) = oldItem == newItem && isSelected(oldItem) == isSelected(newItem)
```

then, pass the `itemDiffCallback` to PagedListEpoxyController.

kakai248:
@baoti Won't work. Diffs are only done after the models get rebuilt. And the models related to the PagedList elements are only rebuilt if the DataSource invalidates, otherwise they are cached. My issue is that I want to rebuild the models without the DataSource elements changing.

elihart:
Merged - thanks for the help!

NapasPayu:
@elihart May I know when this is going to release? Thanks

elihart:
@NapasPayu I'll try to release it in the next few days

NapasPayu:
@elihart I am still waiting for this. Please help push it ^^

kakai248:
It was released with 3.2.0

Issue #567, opened 2018-10-09
[epoxy-paging] Crash in the PagedListEpoxyController when trying to invoke `addTo` on a null model


rohandhruva:
While trying to implement paged list + paged controller, I run into a crash that's not reliably reproducible. The crash seems to happen around 1 in 7 times, completely anecdotally. 

I'm attaching the stack trace: note how my custom controller (that extends `PagedListEpoxyController`) is not in the stack trace because the crash happens in `PagedListEpoxyController` internals. 

I will try to get a small sample app with a reliable reproducible test case. In the meantime, some information that could be useful:

- In `MyImplementationPagedListController`, I have these two params:
  - `modelBuildingHandler = EpoxyAsyncUtil.getAsyncBackgroundHandler()`
  - `diffingHandler = EpoxyAsyncUtil.getAsyncBackgroundHandler()`
- I'm using an `RxPagedListBuilder` with `setFetchScheduler(Schedulers.io())`
- The problem reproduces more frequently with a very small page size: I used size 4 on a total list size 22

I noticed that `PagedListModelCache` holds `null` models, I am guessing one of those is likely mistakenly being passed through in `getModels()`. 

```
java.lang.NullPointerException: Attempt to invoke virtual method 'void com.airbnb.epoxy.EpoxyModel.addTo(com.airbnb.epoxy.EpoxyController)' on a null object reference
at com.airbnb.epoxy.EpoxyController.add(EpoxyController.java:465)
at com.airbnb.epoxy.paging.PagedListEpoxyController.addModels(PagedListEpoxyController.kt:78)
at com.airbnb.epoxy.paging.PagedListEpoxyController.buildModels(PagedListEpoxyController.kt:70)
at com.airbnb.epoxy.EpoxyController$1.run(EpoxyController.java:267)
at android.os.Handler.handleCallback(Handler.java:751)
at android.os.Handler.dispatchMessage(Handler.java:95)
at android.os.Looper.loop(Looper.java:154)
at android.os.HandlerThread.run(HandlerThread.java:61)
```

rohandhruva:
@yigit do you think it's a good idea to filter out nulls from the `getModels()`? I noticed that the method return type is already `List<EpoxyModel<*>>`, so it _should_ not return `null`s, so I'm a bit surprised to see how they're coming through.

yigit:
that is weird, i sample to reproduce would be nice.

I'm guessing the threading model is not working such that models are being modified in another thread.

[...]

PR #586, opened 2018-10-21
Added method to clear model cache and force a model build (#583)

kakai248:
PR for #583

### DIFF
--- a/epoxy-paging/src/main/java/com/airbnb/epoxy/paging/PagedListModelCache.kt
+++ b/epoxy-paging/src/main/java/com/airbnb/epoxy/paging/PagedListModelCache.kt
@@ -142,6 +142,12 @@ internal class PagedListModelCache<T>(
     return modelCache as List<EpoxyModel<*>>
   }
 
+  fun clearModels() {
+    (0 until modelCache.size).forEach {
+      modelCache[it] = null
+    }
+  }
+
   fun loadAround(position: Int) {
     triggerLoadAround(position)
     lastPosition = position
--- a/epoxy-paging/src/main/java/com/airbnb/epoxy/paging/PagedListEpoxyController.kt
+++ b/epoxy-paging/src/main/java/com/airbnb/epoxy/paging/PagedListEpoxyController.kt
@@ -110,6 +110,17 @@ abstract class PagedListEpoxyController<T>(
     modelCache.submitList(newList)
   }
 
+  /**
+   * Requests a model build that will run for every model, including the ones created for the paged
+   * list.
+   *
+   * Clears the current model cache to make sure that happens.
+   */
+  fun requestForcedModelBuild() {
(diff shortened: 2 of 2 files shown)


kakai248:
Yeah, I'll do it in the next days or weekend.

epoxy-paging/src/androidTest/java/com/airbnb/epoxy/paging/PagedListModelCacheTest.kt
```diff
@@ -236,6 +236,13 @@ class PagedListModelCacheTest {
     }
   }
 
+  @Test
+  fun clear() {
+    pagedListModelCache.clearModels()
```
elihart: afaik this doesn't test that models are cleared, just that models are rebuilt. I think a good way to test the behavior is to check that all models are rebuilt - maybe by adding a counter in the `modelBuilder` callback similar to how `rebuildCounter` is used

epoxy-paging/src/main/java/com/airbnb/epoxy/paging/PagedListModelCache.kt
```diff
@@ -142,6 +142,13 @@ internal class PagedListModelCache<T>(
     return modelCache as List<EpoxyModel<*>>
   }
 
+  fun clearModels() {
+    (0 until modelCache.size).forEach {
+      modelCache[it] = null
+    }
+    rebuildCallback()
```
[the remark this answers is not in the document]
kakai248: You're right. I removed the callback call.

epoxy-paging/src/androidTest/java/com/airbnb/epoxy/paging/PagedListModelCacheTest.kt
kakai248: Fixed the test. The callback call in the previous message was hiding the fact that the cache didn't even have models.

epoxy-paging/src/main/java/com/airbnb/epoxy/paging/PagedListModelCache.kt
```diff
@@ -142,6 +142,12 @@ internal class PagedListModelCache<T>(
     return modelCache as List<EpoxyModel<*>>
   }
 
+  fun clearModels() {
+    (0 until modelCache.size).forEach {
```
elihart: I think a simpler approach would be `modelCache.fill(null)`

elihart:
Thanks for updating - this looks ready.

I'm concerned that there could be race conditions by calling this clear function on a different thread at the same time models are being built.

I think it is ok to address that in a follow up task though, since we need to think about synchronization of the whole class.

kakai248:
I'm almost sure there is already a race condition in model building (see #567). That would have to be addressed as a whole.

PR #624, opened 2018-11-29
Upgrade KotlinPoet 1.0.0

shaishavgandhi05:


shaishavgandhi05:
Hmm looks like Paris is still using an older KotlinPoet and that's causing some conflicts. 😐

elihart:
Thanks for looking into this, we could try to update to update Paris soon

shaishavgandhi05:
I can try and update Paris to 1.0.0-RC3 also.

elihart:
That would be awesome

shaishavgandhi05:
Have updated the Paris one as well but haven't got a response there. If @elihart you could take a look at that one.

elihart:
Sorry for the delay, I think it best if @ngsilverman does the review on that one, but I'll follow up to make sure it happens once everyone returns from holiday break in january.

MatushkaRossiya:
KotlinPoet 1.0.1 has been released recently, maybe you can bump the version in this pull request before merging. :)

elihart:
Great - thanks!

PR #627, opened 2018-12-03
Add support for nested RecyclerView to EpoxyVisibilityTracker - Part 1

eboudrant:
The goal here is to add support for nested `RecyclerView`(epoxy managed, ex with `Carousel`) in `EpoxyVisibilityTracker`. No new features here, just internal changes validated by exiting unit tests. I preferred do send multiple smaller pull request for this big change.

So this first change update `EpoxyVisibilityTracker` and `EpoxyVisibilityItem` to support tracking in both direction : width and height. The tracker do not need to know about the scrolling direction.

### DIFF
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityItem.java
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityItem.java
@@ -31,9 +31,6 @@ class EpoxyVisibilityItem {
   @Px
   private int width;
 
-  private float percentVisibleHeight = 0.f;
-  private float percentVisibleWidth = 0.f;
-
   @Px
   private int visibleHeight;
   @Px
@@ -66,15 +63,13 @@ class EpoxyVisibilityItem {
   boolean update(@NonNull View view, @NonNull RecyclerView parent, boolean detachEvent) {
     // Clear the rect before calling getLocalVisibleRect
     localVisibleRect.setEmpty();
-    boolean visibleRect = view.getLocalVisibleRect(localVisibleRect);
-    height = view.getMeasuredHeight();
-    width = view.getMeasuredWidth();
-    viewportHeight = parent.getMeasuredHeight();
-    viewportWidth = parent.getMeasuredWidth();
-    visibleHeight = detachEvent || !visibleRect ? 0 : localVisibleRect.height();
-    visibleWidth = detachEvent || !visibleRect ? 0 : localVisibleRect.width();
-    percentVisibleHeight = detachEvent || !visibleRect ? 0 : 100.f / height * visibleHeight;
-    percentVisibleWidth = detachEvent || !visibleRect ? 0 : 100.f / width * visibleWidth;
+    boolean viewDrawn = view.getLocalVisibleRect(localVisibleRect) && !detachEvent;
+    height = view.getHeight();
+    width = view.getWidth();
+    viewportHeight = parent.getHeight();
+    viewportWidth = parent.getWidth();
+    visibleHeight = viewDrawn ? localVisibleRect.height() : 0;
+    visibleWidth = viewDrawn ? localVisibleRect.width() : 0;
     if (visibleHeight != height || visibleWidth != width) {
       fullyVisible = false;
     }
@@ -119,8 +114,11 @@ class EpoxyVisibilityItem {
 
   void handleChanged(EpoxyViewHolder epoxyHolder) {
     if (visibleHeight != lastVisibleHeightNotified || visibleWidth != lastVisibleWidthNotified) {
-        epoxyHolder.visibilityChanged(percentVisibleHeight, percentVisibleWidth, visibleHeight,
-            visibleWidth);
+      epoxyHolder.visibilityChanged(
+          100.f / height * visibleHeight,
+          100.f / width * visibleWidth,
+          visibleHeight, visibleWidth
+      );
       lastVisibleHeightNotified = visibleHeight;
       lastVisibleWidthNotified = visibleWidth;
     }
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityItem.java
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityItem.java
@@ -27,24 +27,30 @@ class EpoxyVisibilityItem {
   private int adapterPosition = RecyclerView.NO_POSITION;
 
   @Px
-  private int sizeInScrollingDirection;
+  private int height;
+  @Px
+  private int width;
 
-  private int sizeNotInScrollingDirection;
+  private float percentVisibleHeight = 0.f;
+  private float percentVisibleWidth = 0.f;
 
-  private boolean verticalScrolling;
+  @Px
+  private int visibleHeight;
+  @Px
+  private int visibleWidth;
 
-  private float percentVisibleSize = 0.f;
-
-  private int visibleSize;
-
-  private int viewportSize;
+  @Px
+  private int viewportHeight;
+  @Px
+  private int viewportWidth;
 
   private boolean fullyVisible = false;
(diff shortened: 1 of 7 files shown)


epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityItem.java
```diff
@@ -55,31 +61,24 @@
...
-      visibleSize = detachEvent || !visibleRect ? 0 : localVisibleRect.height();
-    } else {
-      sizeNotInScrollingDirection = view.getMeasuredHeight();
-      sizeInScrollingDirection = view.getMeasuredWidth();
-      viewportSize = parent.getMeasuredWidth();
-      visibleSize = detachEvent || !visibleRect ? 0 : localVisibleRect.width();
-    }
-    percentVisibleSize = detachEvent ? 0 : 100.f / sizeInScrollingDirection * visibleSize;
-    if (visibleSize != sizeInScrollingDirection) {
+    height = view.getMeasuredHeight();
```
elihart: any particular reason for using measuredHeight instead of `height` directly?

when might they be different?

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityItem.java
```diff
@@ -55,31 +61,24 @@
...
+    width = view.getMeasuredWidth();
+    viewportHeight = parent.getMeasuredHeight();
+    viewportWidth = parent.getMeasuredWidth();
+    visibleHeight = detachEvent || !visibleRect ? 0 : localVisibleRect.height();
```
elihart: since `detachEvent || !visibleRect` is checked many times, maybe it can be standardized in an local var, like `notDrawn`

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityItem.java
```diff
@@ -154,9 +150,11 @@ private boolean checkAndUpdateInvisible(boolean detachEvent) {
    * component is smaller than half the viewport, when it is fully visible.
    */
   private boolean checkAndUpdateFocusedVisible() {
-    return focusedVisible =
-        sizeInScrollingDirection >= viewportSize / 2 || (visibleSize == sizeInScrollingDirection
-            && sizeInScrollingDirection < viewportSize / 2);
+    boolean focusedVisibleHeight = height >= viewportHeight / 2 || (visibleHeight == height
+        && height < viewportHeight / 2);
```
elihart: won't `height < viewportHeight / 2` always be true here, since the `height >= viewportHeight / 2` check will have failed?

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityItem.java
```diff
@@ -27,24 +27,30 @@
...
-
-  private int sizeNotInScrollingDirection;
-
-  private boolean verticalScrolling;
+  private int height;
+  @Px
+  private int width;
 
-  private float percentVisibleSize = 0.f;
+  private float percentVisibleHeight = 0.f;
```
elihart: would it be possible to compute this as needed from `visibleHeight` and `height`, instead of storing it as a field?

seems more fragile and complex to keep things in sync if we store fields we could derive

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityItem.java
eboudrant: Just old habit, I always used `measuredHeight` ... I moved to `height`.

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityItem.java
eboudrant: Thanks for reviewing that part. After doing more research in that I finally re-implemented this focused/unfocused check using the area since it use not width and height.

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityItem.java
```diff
@@ -55,31 +58,22 @@
...
-    if (visibleSize != sizeInScrollingDirection) {
+    boolean viewDrawn = view.getLocalVisibleRect(localVisibleRect) && !detachEvent;
+    height = view.getHeight();
+    width = view.getWidth();
+    viewportHeight = parent.getHeight();
+    viewportWidth = parent.getWidth();
+    visibleHeight = viewDrawn ? localVisibleRect.height() : 0;
+    visibleWidth = viewDrawn ? localVisibleRect.width() : 0;
+    if (visibleHeight != height || visibleWidth != width) {
       fullyVisible = false;
```
elihart: I know this is from a previous PR, but it's confusing to me that `fullyVisible` is set to false here, but it is only set to true in another method - would it be possible to do all assignment in one place?

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityItem.java
```diff
@@ -165,24 +158,33 @@ private boolean checkAndUpdateFocusedVisible() {
...
   }
 
+  private boolean isInFocusVisible() {
+    final int halfViewportArea = viewportHeight * viewportWidth / 2;
+    final int totalArea = height * width;
+    final int visibleArea = visibleHeight * visibleWidth;
+    // The model has entered the focused range either if it is larger than half of the viewport
+    // and it occupies at least half of the viewport or if it is smaller than half of the viewport
+    // and it is fully visible.
+    return focusedVisible = (totalArea >= halfViewportArea)
```
elihart: it is pretty confusing and fragile that `focusedVisible` is updated here, despite also being updated from the places that call this. Tracing the side effects and intent is difficult.

would it be possible to replace all of the methods to update and check focus with an approach that only mutates `focusedVisible` in one place?

```
    void handleFocus(EpoxyViewHolder epoxyHolder, boolean detachEvent) {
        boolean previousFocusedVisible = focusedVisible;
        focusedVisible = !detachEvent && isInFocusVisible();

        if (focusedVisible != previousFocusedVisible) {
            if (focusedVisible) {
                epoxyHolder.visibilityStateChanged(VisibilityState.FOCUSED_VISIBLE);
            } else {
                epoxyHolder.visibilityStateChanged(VisibilityState.UNFOCUSED_VISIBLE);
            }
        }
    }

    private boolean isInFocusVisible() {
        final int halfViewportArea = viewportHeight * viewportWidth / 2;
        final int totalArea = height * width;
        final int visibleArea = visibleHeight * visibleWidth;
        // The model has entered the focused range either if it is larger than half of the viewport
        // and it occupies at least half of the viewport or if it is smaller than half of the viewport
        // and it is fully visible.
        return (totalArea >= halfViewportArea)
            ? (visibleArea >= halfViewportArea)
            : totalArea == visibleArea;
    }
```

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityItem.java
eboudrant: Yes, I will look into that. I am 100% is simplifying the code and this is a good opportunity now.

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityItem.java
eboudrant: This was a mistake, thanks for spotting it and for the code suggestion!

eboudrant:
Btw, @elihart you can hold off, there are few things I need to verify before. I will let you know once ready to merge.

PR #628, opened 2018-12-04
Return model class from model builder extension function

itome:
Change generated model generate extension function
**from:**
```
inline fun EpoxyController.dataBindingItem(modelInitializer: DataBindingItemBindingModelBuilder.() -> Unit) {
    DataBindingItemBindingModel_().apply  {
        modelInitializer()
    }
    .addTo(this)
}
```

**to:**
```
inline fun EpoxyController.dataBindingItem(modelInitializer: DataBindingItemBindingModelBuilder.() -> Unit): DataBindingEpoxyModel {
    val model = DataBindingItemBindingModel_().apply  {
        modelInitializer()
    }
    model.addTo(this)
    return model
}

```

Then we can get model class instance from controller
```
databindingItemModel = dataBindingItem {
    id("data binding $i")
    ...
}
```

Example use case is animating specific item from EpoxyController.


itome:
@elihart Any progress here?

itome:
Sorry for leaving this PR. I understand your idea. I close.

PR #629, opened 2018-12-05
Prepare for release 3.1.0

elihart:


elihart:
This has already been released to maven, but a test here is failing from the androidx upgrade, so I need to figure that out before merging

PR #632, opened 2018-12-09
Added a getter method for `numViewsToShowOnScreen` in Carousel

AdamMc331:
Helpful for subclasses that need to do things based on this behavior.


elihart:
Thanks!

PR #633, opened 2018-12-10
Add support for nested RecyclerView to EpoxyVisibilityTracker - Part 2

eboudrant:
This is part 2 :
- Auto register `EpoxyVisibilityTracker` on any nested recycler view.
- Any parent visibility event are forwarded to the nested `EpoxyVisibilityTracker`.

Let's discuss the implementation and I will add unit tests later...

### DIFF
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyRecyclerView.java
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyRecyclerView.java
@@ -7,7 +7,6 @@ import android.os.Build.VERSION;
 import android.util.AttributeSet;
 import android.util.SparseArray;
 import android.util.TypedValue;
-import android.view.View;
 import android.view.ViewGroup;
 
 import com.airbnb.viewmodeladapter.R;
@@ -506,18 +505,6 @@ public class EpoxyRecyclerView extends RecyclerView {
     super.swapAdapter(adapter, removeAndRecycleExistingViews);
 
     clearRemovedAdapterAndCancelRunnable();
-  }
-
-  @Override
-  public void onChildAttachedToWindow(@NonNull View child) {
-    if (child instanceof RecyclerView) {
-      RecyclerView recyclerView = (RecyclerView) child;
-      if (recyclerView.getAdapter() instanceof BaseEpoxyAdapter) {
-        // Register itself in the EpoxyVisibilityTracker. This will take care of nested list
-        // tracking (ex: carousel)
-        EpoxyVisibilityTracker.registerNestedRecyclerView(this, (RecyclerView) child);
-      }
-    }
   }
 
   @Override
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyRecyclerView.java
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyRecyclerView.java
@@ -7,6 +7,7 @@ import android.os.Build.VERSION;
 import android.util.AttributeSet;
 import android.util.SparseArray;
 import android.util.TypedValue;
+import android.view.View;
 import android.view.ViewGroup;
 
 import com.airbnb.viewmodeladapter.R;
@@ -508,6 +509,18 @@ public class EpoxyRecyclerView extends RecyclerView {
   }
 
   @Override
+  public void onChildAttachedToWindow(@NonNull View child) {
+    if (child instanceof RecyclerView) {
+      RecyclerView recyclerView = (RecyclerView) child;
+      if (recyclerView.getAdapter() instanceof BaseEpoxyAdapter) {
+        // Register itself in the EpoxyVisibilityTracker. This will take care of nested list tracking
+        // (ex: carousel)
+        EpoxyVisibilityTracker.registerNestedRecyclerView(this, (RecyclerView) child);
+      }
+    }
+  }
+
+  @Override
   public void onAttachedToWindow() {
     super.onAttachedToWindow();
 
@@ -516,10 +529,6 @@ public class EpoxyRecyclerView extends RecyclerView {
       swapAdapter(removedAdapter, false);
     }
     clearRemovedAdapterAndCancelRunnable();
-
-    // Register itself in the EpoxyVisibilityTracker. This will take care of nested list tracking
-    // (ex: carousel)
-    EpoxyVisibilityTracker.registerNestedRecyclerView(this);
   }
 
   @Override
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyRecyclerView.java
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyRecyclerView.java
@@ -513,8 +513,8 @@ public class EpoxyRecyclerView extends RecyclerView {
     if (child instanceof RecyclerView) {
       RecyclerView recyclerView = (RecyclerView) child;
       if (recyclerView.getAdapter() instanceof BaseEpoxyAdapter) {
-        // Register itself in the EpoxyVisibilityTracker. This will take care of nested list tracking
-        // (ex: carousel)
+        // Register itself in the EpoxyVisibilityTracker. This will take care of nested list
+        // tracking (ex: carousel)
         EpoxyVisibilityTracker.registerNestedRecyclerView(this, (RecyclerView) child);
       }
     }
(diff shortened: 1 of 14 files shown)


epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyRecyclerView.java
```diff
@@ -507,6 +508,18 @@ public void swapAdapter(Adapter adapter, boolean removeAndRecycleExistingViews)
     clearRemovedAdapterAndCancelRunnable();
   }
 
+  @Override
+  public void onChildAttachedToWindow(@NonNull View child) {
+    if (child instanceof RecyclerView) {
+      RecyclerView recyclerView = (RecyclerView) child;
+      if (recyclerView.getAdapter() instanceof BaseEpoxyAdapter) {
```
elihart: is `if (recyclerView.getAdapter() instanceof BaseEpoxyAdapter) {` necessary? if the child has had `attach` called on it then it will work, otherwise it is a no-op, right?

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.java
```diff
@@ -58,10 +65,46 @@
...
+   * @param nestedRecyclerView the nested recycler view
+   * @return a visibility tracker if one was attached
+   */
+  @Nullable
+  public static EpoxyVisibilityTracker registerNestedRecyclerView(
+      @NonNull RecyclerView parent,
+      @NonNull RecyclerView nestedRecyclerView
+  ) {
+    EpoxyVisibilityTracker nestedTracker = ATTACHED_TRACKERS.get(nestedRecyclerView);
+    if (nestedRecyclerView.getAdapter() instanceof BaseEpoxyAdapter) {
```
elihart: should we instead enforce this in `attach` to catch errors earlier upstream?

in general I'm not too comfortable using the adapter to check whether the recyclerview is valid, because it's possible to have a null adapter and have the adapter set later.

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.java
```diff
@@ -58,10 +65,46 @@
...
+      // Only attach a tracker if the nested RecyclerView use Epoxy and don't have a tracker yet
+      nestedTracker = ATTACHED_TRACKERS.get(nestedRecyclerView);
```
elihart: this is the same as line 91 - mistake?

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.java
```diff
@@ -169,12 +214,21 @@ private void processChild(@NonNull View child, boolean detachEvent, String event
...
       if (holder instanceof EpoxyViewHolder) {
-        processVisibilityEvents(
+        boolean changed = processVisibilityEvents(
             recyclerView,
             (EpoxyViewHolder) holder,
             detachEvent,
             eventOriginForDebug
         );
+        if (changed) {
+          if (child instanceof EpoxyRecyclerView) {
```
elihart: we are supporting any Recyclerview if they manually attach, right? should the `child instanceof EpoxyRecyclerView` check be removed?

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.java
```diff
@@ -58,10 +65,46 @@
...
+  public static EpoxyVisibilityTracker registerNestedRecyclerView(
+      @NonNull RecyclerView parent,
+      @NonNull RecyclerView nestedRecyclerView
+  ) {
+    EpoxyVisibilityTracker nestedTracker = ATTACHED_TRACKERS.get(nestedRecyclerView);
+    if (nestedRecyclerView.getAdapter() instanceof BaseEpoxyAdapter) {
+      // Only attach a tracker if the nested RecyclerView use Epoxy and don't have a tracker yet
+      nestedTracker = ATTACHED_TRACKERS.get(nestedRecyclerView);
+      if (nestedTracker == null) {
+        EpoxyVisibilityTracker parentTracker = ATTACHED_TRACKERS.get(parent);
```
elihart: this system of using static references to check attached states seems flimsy.

I see the need, but possibly have a better idea. What if we used view tags to store the tracker for a recyclerview instead? So when you call `attach` the attacher adds itself as a view tag associated with a custom id, and then for any other recyclerview in the future we can get it's attacher and know if it is set up for visibiility tracking or not?

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyRecyclerView.java
eboudrant: The idea here was if the child is a `RecyclerView` which is not managed by epoxy there is no point to have visibility tracking. But that's true that the same check is in `EpoxyVisibilityTracker.registerNestedRecyclerView` ... we could remove it from here. This way no need to hookup `onChildAttachedToWindow` method. 

When you say `attach` do you mean `EpoxyVisibilityTracker.attach` ? Because the idea of this implementation was to automatically attach nested `RecyclerView`. May be it is a bad idea and it would be better to have user explicitly call `EpoxyVisibilityTracker.attach` for nested `RecyclerView`s.

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.java
eboudrant: I like the idea of using tag better ! Let me check that...

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyRecyclerView.java
eboudrant: Actually we could use a `OnChildAttachStateChangeListener` to automatically attach nested visibility trackers from the parent. Not sure what is the best, auto-attach nested or let the the developer attach them ?

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.java
```diff
@@ -58,10 +87,46 @@
...
+  /**
+   * OnAttachStateChangeListener used to lookup for tracked RecyclerView's parent. If the parent
+   * have a visibility tracker attached we will register nested tracker.
+   */
+  private final OnAttachStateChangeListener onAttachStateChangeListener =
+      new OnAttachStateChangeListener() {
+
+        @Override
+        public void onViewAttachedToWindow(View v) {
+          if (attachedRecyclerView != null) {
```
elihart: doesn't seem like it should be possible to have `attachedRecyclerView` be null here - I'm not against the check, but also think it's not bad to crash since it should be a bug

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyRecyclerView.java
eboudrant: So as discussed I changed to use `OnChildAttachStateChangeListener` which is simpler.
Thanks @elihart

elihart:
@eboudrant this looks great to me - do you mind adding some tests before I merge?

thanks!

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.java
```diff
@@ -431,5 +448,13 @@ private void onItemMoved(int fromPosition, int toPosition) {
         }
       }
     }
+
+    /**
+     * @param recyclerView the recycler view
+     * @return true if managed by an {@link BaseEpoxyAdapter}
+     */
+    private boolean notEpoxyManaged(RecyclerView recyclerView) {
```
elihart: @Nullable?

kotlinsample/src/main/java/com/airbnb/epoxy/kotlinsample/views/CarouselNoSnap.kt
```diff
@@ -3,16 +3,11 @@ package com.airbnb.epoxy.kotlinsample.views
...
 import com.airbnb.epoxy.Carousel
-import com.airbnb.epoxy.EpoxyVisibilityTracker
 import com.airbnb.epoxy.ModelView
 import com.airbnb.epoxy.ModelView.Size
 
 @ModelView(saveViewState = true, autoLayout = Size.MATCH_WIDTH_WRAP_HEIGHT)
 class CarouselNoSnap(context: Context) : Carousel(context) {
 
-    init {
-        EpoxyVisibilityTracker().attach(this)
```
elihart: 👍

PR #656, opened 2019-01-12
Synchronize paged epoxy controller

elihart:
There have been a slew of issues with the paged controller, like NPE's and duplicate items, which point to concurrency issues. We have previously relied on model building and the paged controller's model cache using the same thread to prevent this, but it doesn't seem to be working completely.

One issue is that updates to the PagedList get dispatched on it's `notifyExecutor`, which is outside of the control of Epoxy, and we were relying on documentation to tell people to make sure this was the same thread as the model building thread. This doesn't seem like a scalable approach.

Instead, I have synchronized all access to the model cache

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,11 +1,10 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 buildscript {
 
-  ext.KOTLIN_VERSION = "1.3.20-eap-100"
+  ext.KOTLIN_VERSION = "1.2.71"
   ext.ANDROID_PLUGIN_VERSION = "3.2.1"
 
   repositories {
-    maven { url 'https://dl.bintray.com/kotlin/kotlin-eap' }
     google()
     jcenter()
   }
@@ -23,7 +22,6 @@ plugins {
 allprojects {
 
   repositories {
-    maven { url 'https://dl.bintray.com/kotlin/kotlin-eap' }
     google()
     jcenter()
   }
--- a/epoxy-adapter/build.gradle
+++ b/epoxy-adapter/build.gradle
@@ -1,5 +1,6 @@
 apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
+apply plugin: 'kotlin-kapt'
 
 android {
   compileSdkVersion rootProject.COMPILE_SDK_VERSION
@@ -22,11 +23,11 @@ dependencies {
   implementation rootProject.deps.androidAnnotations
   implementation rootProject.deps.androidRecyclerView
   implementation rootProject.deps.androidDesignLibrary
+  implementation rootProject.deps.kotlin
   api project(':epoxy-annotations')
 
-  annotationProcessor project(':epoxy-processor')
+  kapt project(':epoxy-processor')
 
-  testImplementation rootProject.deps.kotlin
   testImplementation rootProject.deps.junit
   testImplementation rootProject.deps.robolectric
   testImplementation rootProject.deps.mockito
--- a/epoxy-pagingsample/build.gradle
+++ b/epoxy-pagingsample/build.gradle
@@ -19,7 +19,7 @@ android {
 dependencies {
   implementation rootProject.deps.kotlin
 
-  implementation "org.jetbrains.anko:anko-coroutines:0.10.8"
+  implementation "org.jetbrains.anko:anko-coroutines:0.10.5"
   implementation "androidx.room:room-common:$ANDROIDX_ROOM"
   implementation "androidx.room:room-runtime:$ANDROIDX_ROOM"
   implementation "androidx.paging:paging-common:$ANDROIDX_PAGING"
--- a/epoxy-pagingsample/src/main/java/com/airbnb/epoxy/pagingsample/PagingSampleActivity.kt
+++ b/epoxy-pagingsample/src/main/java/com/airbnb/epoxy/pagingsample/PagingSampleActivity.kt
@@ -18,9 +18,11 @@ import com.airbnb.epoxy.EpoxyModel
 import com.airbnb.epoxy.ModelView
 import com.airbnb.epoxy.TextProp
 import com.airbnb.epoxy.paging.PagedListEpoxyController
-import kotlinx.coroutines.GlobalScope
-import kotlinx.coroutines.delay
-import kotlinx.coroutines.launch
+import kotlinx.coroutines.experimental.CommonPool
+import kotlinx.coroutines.experimental.delay
+import kotlinx.coroutines.experimental.launch
+import org.jetbrains.anko.coroutines.experimental.bg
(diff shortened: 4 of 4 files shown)


epoxy-paging/src/main/java/com/airbnb/epoxy/paging/PagedListModelCache.kt
```diff
@@ -35,126 +35,135 @@ import java.util.concurrent.Executor
...
+        @Synchronized
+        override fun onInserted(position: Int, count: Int) {
+            (0 until count).forEach {
+                modelCache.add(position, null)
+            }
+            rebuildCallback()
         }
-        // we have to reply on this private API, otherwise, paged list might be changed when models are being built,
-        // potentially creating concurrent modification problems.
-        builder.setMainThreadExecutor {runnable : Runnable ->
```
yigit: this should be doing the thread sync.

elihart: @yigit right, I understand that part (apologies for the indentation change from auto-format, it makes this hard to read, but most of the lines are the same.)

But as far as I can tell, this only guarantees synchronization when diffing a new paged list.

the main thing I'm trying to address with the synchronization is changes to the PagedList that get notified to the cache via the paged list's `notifyExecutor` (since the `AsyncPagedListDiffer`  uses `PagedList#addWeakCallback` to register `mPagedListCallback`)

You've said before that having people set their PagedList notifyExecutor to the same thread as model building would fix this, but that seems very fragile to me, and easy for people to miss.

Is there anything wrong with forcibly synchronizing the cache access across threads so we don't have to worry about the notifyExecutor?

elihart:
@yigit could you take a look at this new approach when you get a chance?

epoxy-paging/src/main/java/com/airbnb/epoxy/paging/PagedListModelCache.kt
```diff
@@ -31,130 +32,222 @@ import java.util.concurrent.Executor
...
+ * However, change notifications from the PageList happen on that list's notify executor which is
+ * out of our control, and we require the user to configure that properly, or an error is thrown.
+ *
+ * There are two special cases:
+ *
+ * 1. The first time models are built happens synchronously for immediate UI. In this case we don't
+ * use the model cache (to avoid data synchronization issues), but attempt to fill the cache with
+ * the models later.
+ *
+ * 2. When a list is submitted it can trigger update callbacks synchronously. Since we don't control
```
yigit: this looks like a paging bug. we should not be notifying on another thread. though paging receives just an executor so probably paging is unable to check that :/.

epoxy-paging/src/main/java/com/airbnb/epoxy/paging/PagedListModelCache.kt
```diff
@@ -31,130 +32,222 @@ import java.util.concurrent.Executor
...
-        (0 until count).forEach { _ ->
-          modelCache.add(position, null)
+    /**
+     * Changes to the paged list must happen on the same thread as changes to the model cache to
+     * ensure they stay in sync.
+     *
+     * We can't force this to happen, and must instead rely on user's configuration, but we can alert
+     * when it is not configured correctly.
+     *
+     * An exception is if the callback happens due to a new paged list being submitted, which can
```
yigit: An exception is "thrown"

epoxy-paging/src/main/java/com/airbnb/epoxy/paging/PagedListModelCache.kt
```diff
@@ -31,130 +32,222 @@ import java.util.concurrent.Executor
...
+    /**
+     * Tracks the last accessed position so that we can report it back to the paged list when models are built.
+     */
+    private var lastPosition: Int? = null
+
+    /**
+     * Set to true while a new list is being submitted, so that we can ignore the update callback
+     * thread restriction.
+     */
+    private var inSubmitList: Boolean = false
```
yigit: there is a race condition where submit might be called in thread A at the same time we are checking this value. Needs to be thread local

epoxy-paging/src/main/java/com/airbnb/epoxy/paging/PagedListModelCache.kt
```diff
@@ -31,130 +32,222 @@ import java.util.concurrent.Executor
...
+
+    @Synchronized
+    fun getModels(): List<EpoxyModel<*>> {
+        val currentList = asyncDiffer.currentList ?: emptyList<T>()
+
+        // The first time models are built the EpoxyController does so synchronously, so that
+        // the UI can be ready immediately. To avoid concurrent modification issues with the PagedList
+        // and model cache we can't allow that first build to touch the cache.
+        if (Looper.myLooper() != modelBuildingHandler.looper) {
+            val initialModels = currentList.mapIndexed { position, item ->
```
yigit: hmm we may end up building more than necessary (or too much) but maybe it is OK since paged list wont load too much.

epoxy-paging/src/main/java/com/airbnb/epoxy/paging/PagedListModelCache.kt
```diff
@@ -31,130 +32,222 @@ import java.util.concurrent.Executor
...
+            // with the models we built to avoid needing to rebuild them later.
+            modelBuildingHandler.post {
+                setCacheValues(currentList, initialModels)
+            }
+
+            return initialModels
+        }
+
+        (0 until modelCache.size).forEach { position ->
+            if (modelCache[position] != null) {
```
yigit: not sure maybe i wrot ethis code but i think it is easier to read if we just say

if (modelCache[position] == null) {
    // call build
}

to avoid return@foreach

or if we want to get kotlin fancy, we can do:

modelCache[position] ?: modelBuilder(position, list[position].also {  modelCache[position] = it}

epoxy-paging/src/main/java/com/airbnb/epoxy/paging/PagedListModelCache.kt
elihart: I think the `@Synchronized` prevent that from happening, since all access to this variable is behind synchronized functions

elihart:
Thanks for the review yigit - I think this is pretty good to go, sorry it is uglier than your original version, but I hope it fixes problems for people.

I noticed that the paging sample you setup only does model building on the main thread, and the LivePagedListBuild doesn't seem to have a way to change the notify executor, which means we can't have a sample where the model building happens asynchronously. 

Since that is probably a very common use case for people and an important thing to test it would be nice to enable that, but I couldn't figure out a great way to do it (I don't have much experience with these arch components)

elihart:
@PedroAlvarado working on it - seems to be a compatibility issue with kotlin 1.3.0+

PR #657, opened 2019-01-12
Enable recycling of views within groups

elihart:
There has been a limitation in EpoxyModelGroup that groups with the same view type had to contain the same number of models. This made it very inflexible for dynamically changing group size. Additionally, groups couldn't share subviews between each other, so view recycling was not as effective.

This changes groups to recycle child views into a shared pool among all groups - which also enables groups to have dynamic amounts of models.

This also adds kotlin to the project for the first time 🎉 

This works in the sample app, but I still need to write quite a few tests for it.

@ngsilverman @gpeal

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -48,6 +48,9 @@
       <option name="JD_KEEP_EMPTY_RETURN" value="false" />
     </JavaCodeStyleSettings>
     <JetCodeStyleSettings>
+      <option name="PACKAGES_TO_USE_STAR_IMPORTS">
+        <value />
+      </option>
       <option name="NAME_COUNT_TO_USE_STAR_IMPORT" value="2147483647" />
       <option name="NAME_COUNT_TO_USE_STAR_IMPORT_FOR_MEMBERS" value="2147483647" />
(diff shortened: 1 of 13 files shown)


epoxy-adapter/build.gradle
```diff
@@ -22,6 +22,7 @@ dependencies {
   implementation rootProject.deps.androidAnnotations
   implementation rootProject.deps.androidRecyclerView
   implementation rootProject.deps.androidDesignLibrary
+  implementation rootProject.deps.kotlin
```
eboudrant: Like Checkstyle for Java, do you want to bring ktlint for the Kotlin code ?
However here will be a bunch of code to re-format it the annotation processor that is already in Kotlin.

elihart: I think ktlint would be great - we would just have to reformat all existing kotlin code to match.

I don't feel like doing it in this PR (since there is already a lot of other kotlin code), but it would be a good thing to do soon.

Thanks for bringing it up, I'll create an issue.

elihart:
This has been updated with tests, and a few fixes, plus formatting with ktlint since it was added

PR #660, opened 2019-01-14
Add missing nullable in EpoxyVisibilityTracker

eboudrant:
Add missing `@Nullable` in `EpoxyVisibilityTracker.notEpoxyManaged`


elihart:
thanks 👍

Issue #659, opened 2019-01-14
Add Ktlint to enforce kotlin style


elihart:
Since kotlin is going to be used more and more in the project we should enforce consistent style.

I'd love some help with this if anyone is interested.

PR #661, opened 2019-01-14
Add Ktlint to enforce kotlin style

eboudrant:
Per #659 this add [ktlint](https://ktlint.github.io/) to the project. The only part I am not 100% sure is the idea config files. I am not an expert with that : `.idea/codeStyles/Project.xml`


ktlint.gradle
```diff
@@ -0,0 +1,34 @@
...
+  // see https://github.com/shyiko/ktlint#usage for more
+}
+check.dependsOn ktlint
+
+task ktlintFormat(type: JavaExec, group: "formatting") {
+  description = "Fix Kotlin code style deviations."
+  main = "com.github.shyiko.ktlint.Main"
+  classpath = configurations.ktlint
+  args "-F", "src/**/*.kt"
+}
```
shaishavgandhi05: Endline

shaishavgandhi05:
It's unfortunate that ktlint doesn't have a baseline option. 😐 
This would be quiet a large diff.

elihart:
Looks great, thanks so much for jumping on this!

The project code style file looks right - ktlint recommends editorconfig (https://editorconfig.org/#download), but I haven't used that before. It might be worth trying in the future - for now, does it seem that the Intellij autoformat matches what ktlint expects?

So usage is `./gradlew ktlintFormat` or `./gradlew ktlint`, right?

It'd be nice if this was clear to any other contributors, we can modify the readme with these commands
https://github.com/airbnb/epoxy#contributing
https://github.com/airbnb/epoxy/blob/master/CONTRIBUTING.MD

eboudrant:
`./gradlew ktlintFormat` -> auto format all `.kt` files (when it is possible).
`./gradlew ktlint` -> check the formatting (similar to what `./gradlew checkstyle` is to Java)

I will check with the `editorconfig`, I am not yet not familiar with it.

eboudrant:
> Looks great, thanks so much for jumping on this!
> 
> The project code style file looks right - ktlint recommends editorconfig (https://editorconfig.org/#download), but I haven't used that before. It might be worth trying in the future - for now, does it seem that the Intellij autoformat matches what ktlint expects?
> 
> So usage is `./gradlew ktlintFormat` or `./gradlew ktlint`, right?
> 
> It'd be nice if this was clear to any other contributors, we can modify the readme with these commands
> https://github.com/airbnb/epoxy#contributing
> https://github.com/airbnb/epoxy/blob/master/CONTRIBUTING.MD

So @elihart actually the usage is still reformat from the IDE and `./gradlew check` so I am not sure it is worth to update the `CONTRIBUTING.MD` as it is still valid.

Or if you prefer I can write : 

# Epoxy is an Open Source Project

Pull requests are welcome! We'd love help improving this library.

We have a code style setting for the project (checkstyle for Java, ktlint for Kotlin). Please run `Reformat Code` in Android Studio (or Intellij) on changed files before pushing them. Alternatively for Kotlin you can use [ktlint](https://ktlint.github.io/) tasks: check Kotlin code formatting with`./gradlew ktlint` and reformat all Kotlin code with `./gradlew ktlintformat`.

Also, run `./gradlew check` locally to make sure that style checks and tests pass.

If you update the model annotation processor you may find the `update_processor_test_resources.rb` script very useful for updating the existing tests with your changes.

elihart:
@eboudrant that change to the contributing doc seems good to me, if you don't mind making the change

eboudrant:
Ok cool, I will update it.



Now for the `.editorconfig` I don't think we need it and I don't think ktlint recommend it. They just mention they just added support in ktlint to process it, I would guess it is for project that already use  `editorconfig` :

> * Starting from 0.8.0 value of indent_size specified under [*{kt,kts}] section in .editorconfig takes precedence (if any). Official recommendation is to use 4 spaces, though. (see #43 for details)

So I would not add it unless you really want it. When it is not present ktlint will simply use Kotlin Style Guide indentation style.

elihart:
looks great - thanks!!!

PR #671, opened 2019-01-31
Changed the section name for the timer class to be in the start method

L7ColWinters:
This follows the more standard android.os.Trace class which allows to make traces or also use Firebase performance monitoring with an overrode DebugTimer class in a local project.

### DIFF
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/DebugTimer.java
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/DebugTimer.java
@@ -6,6 +6,7 @@ class DebugTimer implements Timer {
 
   private final String tag;
   private long startTime;
+  private String sectionName;
 
   DebugTimer(String tag) {
     this.tag = tag;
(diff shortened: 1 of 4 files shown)


epoxy-adapter/src/main/java/com/airbnb/epoxy/DebugTimer.java
```diff
@@ -14,10 +15,11 @@
 
   private void reset() {
     startTime = -1;
+    sectionName = null;
   }
 
   @Override
-  public void start() {
+  public void start(String sectionName) {
```
elihart: this isn't used, did you mean to assign the field?

L7ColWinters: whoops, was copying over from other repo

L7ColWinters:
It would be nice to have an API but then on every EpoxyController it would need to pass in an optional Timer instance. What i'm doing is overriding the DebugTimer class under the airbnb package to inject my own implementation. One could also make a static method to set the instance but I'm not a fan of static methods for DI. Do you think it would make sense to add a static method on the EpoxyController class?

elihart:
I would be fine with a static method to set a global change to the Timer, we already have some other static methods for global defaults.

You're right that it's not ideal, but imo it's better than overriding the class.

Issue #672, opened 2019-02-01
EpoxyModelGroup with ViewStub


eboudrant:
I am trying to layout a model group using `ViewStub`, here the xml :

```
<?xml version="1.0" encoding="utf-8"?>
<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
    android:layout_width="match_parent"
    android:layout_height="wrap_content"
    android:orientation="horizontal">

    <!-- First element -->
    <ViewStub
        android:layout_width="60dp"
        android:layout_height="60dp" />

    <!-- Elastic space -->
    <View
        android:layout_width="0dp"
        android:layout_height="match_parent"
        android:layout_weight="1" />

    <!-- And up to 3 elements -->
    <ViewStub
        android:layout_width="60dp"
        android:layout_height="60dp" />
    <ViewStub
        android:layout_width="60dp"
        android:layout_height="60dp" />
    <ViewStub
        android:layout_width="60dp"
        android:layout_height="60dp" />

</LinearLayout>
```

I am expecting :

<img width="417" alt="screen shot 2019-02-01 at 1 45 34 pm" src="https://user-images.githubusercontent.com/773377/52151579-767d9600-2628-11e9-8287-6f8606de393d.png">

But I am seeing this :

<img width="415" alt="screen shot 2019-02-01 at 1 46 19 pm" src="https://user-images.githubusercontent.com/773377/52151590-82695800-2628-11e9-8db5-f6747a8f34c8.png">

Is there any issue with `EpoxyModelGroup` / `ViewStub` or is it a miss-use from me ?

Full code is here : https://github.com/eboudrant/epoxy/pull/9

eboudrant:
Looks like the stub are removed too early from the layout :

https://github.com/airbnb/epoxy/blob/2de36ba0fdb2223dcb51900efd41da952b5cd1cf/epoxy-adapter/src/main/java/com/airbnb/epoxy/ModelGroupHolder.kt#L67

The position passed in the `ViewStubData` would not allow to re-position correctly the nested model.

elihart:
Thanks for reporting. good catch, this is definitely wrong.

I'm traveling now, but can fix it early next week - sorry for the issue!

Maybe an easy fix could be to have `getNextViewStub` add the count of existing stubs to the index of the newly created stub data - I think that would give the correct adjusted position while not requiring a change to the view removal - I'll have to think about it a bit more though.

elihart:
[...]

Issue #675, opened 2019-02-05
IndexOutOfBoundsException in EpoxyModelGroup binding


eboudrant:
Hi,

We use `EpoxyModelGroup` in our app and since then it crash on `IndexOutOfBoundsException` (see stack-trace below). Unfortunately I cannot publish our code but I will try to repro in a sample app. The crash happen when a model group get updated and when the nested models increase. 

It crash on this line in `EpoxyModelGroup.java`
```
EpoxyModel<?> previousModel = previousGroup.models.get(modelIndex);` 
```

Before they were 4 models in the group and after the update there were 11 models.

```
java.lang.IndexOutOfBoundsException: Index: 4, Size: 4
 	at java.util.ArrayList.get(ArrayList.java:437)
 	at com.airbnb.epoxy.EpoxyModelGroup$3.onModel(EpoxyModelGroup.java:144)
	at com.airbnb.epoxy.EpoxyModelGroup.iterateModels(EpoxyModelGroup.java:197)
 	at com.airbnb.epoxy.EpoxyModelGroup.bind(EpoxyModelGroup.java:139)
 	at com.airbnb.epoxy.EpoxyViewHolder.bind(EpoxyViewHolder.java:57)
	at com.airbnb.epoxy.BaseEpoxyAdapter.onBindViewHolder(BaseEpoxyAdapter.java:97)
 	at com.airbnb.epoxy.EpoxyControllerAdapter.onBindViewHolder(EpoxyControllerAdapter.java:16)
	at com.airbnb.epoxy.BaseEpoxyAdapter.onBindViewHolder(BaseEpoxyAdapter.java:15)
 	at androidx.recyclerview.widget.RecyclerView$Adapter.bindViewHolder(RecyclerView.java:6823)
 	at androidx.recyclerview.widget.RecyclerView$Recycler.tryBindViewHolderByDeadline(RecyclerView.java:5752)
 	at androidx.recyclerview.widget.RecyclerView$Recycler.tryGetViewHolderForPositionByDeadline(RecyclerView.java:6019)
 	at androidx.recyclerview.widget.RecyclerView$Recycler.getViewForPosition(RecyclerView.java:5858)
 	at androidx.recyclerview.widget.RecyclerView$Recycler.getViewForPosition(RecyclerView.java:5854)
 	at androidx.recyclerview.widget.LinearLayoutManager$LayoutState.next(LinearLayoutManager.java:2230)
[...]

PR #676, opened 2019-02-05
Fix model group issues

elihart:
Should fix https://github.com/airbnb/epoxy/issues/672 and https://github.com/airbnb/epoxy/issues/675

Tests were updated to catch these issues

### DIFF
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/ModelGroupHolder.kt
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/ModelGroupHolder.kt
@@ -58,40 +58,31 @@ class ModelGroupHolder : EpoxyHolder() {
     }
 
     private fun createViewStubData(viewGroup: ViewGroup): List<ViewStubData> {
-        val stubs = ArrayList<ViewStubData>(4)
+        return ArrayList<ViewStubData>(4).apply {
 
-        while (true) {
-            val stubData = getNextViewStub(viewGroup)
-            if (stubData != null) {
-                stubs.add(stubData)
-                stubData.removeView()
-            } else {
-                break
+            collectViewStubs(viewGroup, this)
+
+            if (isEmpty()) {
+                throw IllegalStateException(
+                    "No view stubs found. If viewgroup is not empty it must contain ViewStubs."
+                )
             }
         }
-
-        if (stubs.isEmpty()) {
-            throw IllegalStateException(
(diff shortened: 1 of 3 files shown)


epoxy-adapter/src/main/java/com/airbnb/epoxy/ModelGroupHolder.kt
```diff
@@ -58,40 +58,31 @@ class ModelGroupHolder : EpoxyHolder() {
     }
 
     private fun createViewStubData(viewGroup: ViewGroup): List<ViewStubData> {
-        val stubs = ArrayList<ViewStubData>(4)
+        return ArrayList<ViewStubData>(4).apply {
```
eboudrant: Just curious, why don't you use Kotlin list : `mutableListOf<ViewStubData>()` ?

epoxy-adapter/src/main/java/com/airbnb/epoxy/ModelGroupHolder.kt
```diff
@@ -217,7 +210,12 @@ private class ViewStubData(
         }
     }
 
-    fun removeView() {
+    fun resetStub() {
```
eboudrant: `private` ?

epoxy-adapter/src/main/java/com/airbnb/epoxy/ModelGroupHolder.kt
elihart: I believe it was so I could give an initial capacity value

epoxy-adapter/src/main/java/com/airbnb/epoxy/ModelGroupHolder.kt
elihart: it's actually used in the holder class in the same file, but ViewStubData is private

eboudrant: ah, could reduce to `internal` then ;)

epoxy-adapter/src/main/java/com/airbnb/epoxy/ModelGroupHolder.kt
eboudrant: Ok, why 4 ?

elihart: it's a somewhat arbitrary guess - groups are generally small, but non zero size

PR #688, opened 2019-02-14
Disabled the auto-detach behavior on Carousels by default.

mradzinski:
Hey @elihart, I was implementing some SharedElementTransition's on a Carousel and found out under certain conditions some may fail due to its adapter being detached, causing models to be unbound and hence failing to execute the exit part of a SharedElementTransition properly (adapter is reattached, scroll position reset). 

I solved this by extending Carousel and calling `setRemoveAdapterWhenDetachedFromWindow` on `init`, but I guess exposing a ModelProp for it on the Carousel itself shouldn't cause any side effects.

Edit:
Also exposed `setDelayMsWhenRemovingAdapterOnDetach` to keep consistency with `EpoxyRecyclerView`

### DIFF
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/Carousel.java
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/Carousel.java
@@ -494,18 +494,6 @@ public class Carousel extends EpoxyRecyclerView {
     super.setModels(models);
   }
 
-  @ModelProp
-  @Override
-  public void setRemoveAdapterWhenDetachedFromWindow(boolean removeAdapterWhenDetachedFromWindow) {
-    super.setRemoveAdapterWhenDetachedFromWindow(removeAdapterWhenDetachedFromWindow);
-  }
-
-  @ModelProp
-  @Override
-  public void setDelayMsWhenRemovingAdapterOnDetach(int delayMsWhenRemovingAdapterOnDetach) {
-    super.setDelayMsWhenRemovingAdapterOnDetach(delayMsWhenRemovingAdapterOnDetach);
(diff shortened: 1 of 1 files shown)


elihart:
Hey there, thanks for putting this up. I'm not opposed to this, but I'm wondering what conditions are causing this problem exactly? The default delay of 2 seconds would be more than enough time to allow the shared element transition to finish, I would have thought.

In general I think the auto detaching doesn't make as much sense for carousels, since they will be detached when their parent recyclerview is. It might be a good idea to make the Carousel disable the auto detach behavior by default.

mradzinski:
@elihart 

> The default delay of 2 seconds would be more than enough time to allow the shared element transition to finish, I would have thought.

Not using fragments here but views. In my scenario the delay works fine for the shared element transition to finish its way in, but once the view is detached the adapter will detach too. The adapter will later re-attach when navigating back causing the carousel to reset its scroll position, meaning the exit transition will not work for some items (specifically those items which aren't yet bound... last carousel item for example). Due to this, I'm disabling the automatic adapter removal for both carousel and parent EpoxyRecyclerView and handling it myself.

> In general I think the auto detaching doesn't make as much sense for carousels, since they will be detached when their parent recyclerview is. It might be a good idea to make the Carousel disable the auto detach behavior by default.

Agreed. I can disable this as part of the same PR if needed.

elihart:
Thanks for explaining your use case.

Can you change this so that instead of exposing these options as props, you make ever carousel disable this behavior for itself automatically?

mradzinski:
Done @elihart

elihart:
Thanks!

PR #750, opened 2019-05-19
Update kotlin to 1.3.31

miga9:
Updating kotlin from 1.2.71 to the latest, 1.3.31.

### DIFF
--- a/blessedDeps.gradle
+++ b/blessedDeps.gradle
@@ -34,6 +34,7 @@ rootProject.ext.ANDROID_ARCH_TESTING = "1.1.1"
 rootProject.ext.ANDROID_TEST_RUNNER = "1.0.2"
 rootProject.ext.SQUARE_JAVAPOET_VERSION = "1.11.1"
 rootProject.ext.SQUARE_KOTLINPOET_VERSION = "1.0.1"
+rootProject.ext.KOTLIN_COROUTINES_VERSION = "1.2.1"
 rootProject.ext.KOTLIN_METADATA_VERSION = "1.4.0"
 
 rootProject.ext.ASSERTJ_VERSION = "1.7.1"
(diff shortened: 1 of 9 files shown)


elihart:
Thanks for doing this, but it is failing with
```
e: error: Epoxy Processor Exception:  Caused by BadClassFile: bad class file: /home/travis/build/airbnb/epoxy/epoxy-processor/build/libs/epoxy-processor-3.5.0.jar(com/airbnb/epoxy/KotlinUtilsKt$buildAnnotationSpecs$1.class)
    bad enclosing method attribute for class com.airbnb.epoxy.KotlinUtilsKt$buildAnnotationSpecs$1
    Please remove or make sure it appears in the correct subdirectory of the classpath.       Stacktrace:      com.airbnb.epoxy.ConfigManager.processConfigurations(ConfigManager.java:90)       com.airbnb.epoxy.EpoxyProcessor.processRound(EpoxyProcessor.java:185)       com.airbnb.epoxy.EpoxyProcessor.process(EpoxyProcessor.java:165)
```

Not sure why this is happening, but if you can fix it that would be fantastic :)

miga9:
@elihart I was able to get it to work after a couple changes. 

1: Removed the default function parameters. I don't understand how this causes the bad class file issue. Would be curious to hear some theories/ideas if any.

2: Replaced the usages of experimental coroutines in anko with the stable coroutines. It was throwing the error: `java.lang.IllegalStateException: Backend Internal error: Exception during code generation`

elihart:
Oh great! really appreciate you fixing that up - I have no idea why the default param values are causing problems, but nice work figuring it out. Also nice to have coroutines updated :)

josephyanks:
@elihart The kotlin upgrade is a blocker for us upgrading, would it be possible to do a minor release to 3.5.1 so we can use an official version?

elihart:
@josephyanks sure, I just pushed a 3.5.1 version with the change - it should be available in a few minutes

josephyanks:
Thank you!

miga9:
@elihart Thank you for addressing this so quickly!

PR #593, opened 2018-10-24
Preload and Preload sample

KirkBushman:
I'm committing the preload module and a working sample, check if all is working as you want.

There were some visibility issues, for example, BaseEpoxyAdapter wasn't visible. I just changed to public  for now to leave that last. How do you want to proceed?  Do you want me to inject everything into the same package with a gradle task?

I've set gradle version to 4.8.0, that uses support libs 27.1.1, while epoxy uses 27.1.0, this was causing ./gradlew check to fail, I then set transitive to false, to make it pass. This is not ideal, do you want to step up to 27.1.1?

This PR is the most, I'll do another commit with the changes for these details.

PR #766, opened 2019-06-14
Preloader system with glide extensions

elihart:
Updated version of https://github.com/airbnb/epoxy/pull/593

This creates a generic preloading system, with integration into EpoxyRecyclerView.

It also adds an optional module for easy prefetching with glide.

Usage is as simple as
```
recycler_view.addGlidePreloader(
            Glide.with(this),
            preloader = glidePreloader { requestManager, model: ImageModel_, _ ->
                requestManager.loadImage(model.imageUrl)
            }
        )
```

Once this is merged I'll push a new release with it and add a usage guide to the wiki.

If anyone is interested in adding support for other image libraries, like Picasso, feel free to open a PR and copy the glide implementation here. It is pretty straightforward.


elihart:
There is an issue where running the annotation processor in tests fails fails to properly compile because of issues with the new kotlin code in epoxy-adapter. Can't merge until that is figured out

PR #775, opened 2019-06-27
 Add a method to request visibility check externally

eboudrant:
Add a method to request visibility check on the `EpoxyVisibilityTracker`. 

An example is when you nest a horizontal Epoxy backed RecyclerView in a non Epoxy vertical RecyclerView. When the vertical RecyclerView scroll you want to notify the visibility tracker attached on the horizontal RecyclerView.

Unfortunately we need it in our app as some screens are a combination of Epoxy and non Epoxy UI.


elihart:
Do you need a new release with this soon?

eboudrant:
Thanks for quick review ! If you ask then we would need it in a about week ;)

PR #778, opened 2019-07-09
Refactor EpoxyItemspacingDecorator getItemOffsets 

apurbalal:
Refactor `EpoxyItemspacingDecorator.getItemOffsets(...)`
```
Current implementation 
- padding = pxBetweenItems / 2
- set padding on left, right, top and bottom

Which can be reduced to
- set pxBetweenItems on left and top
```
This will reduce the amount of statements executed by half.

### DIFF
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyItemSpacingDecorator.java
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyItemSpacingDecorator.java
@@ -61,29 +61,18 @@ public class EpoxyItemSpacingDecorator extends RecyclerView.ItemDecoration {
     calculatePositionDetails(parent, position, layout);
 
     boolean left = useLeftPadding();
-    boolean right = useRightPadding();
     boolean top = useTopPadding();
-    boolean bottom = useBottomPadding();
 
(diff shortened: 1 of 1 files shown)


elihart:
Thank you for spending time to submit this, but I don't think it should be done for a few reasons:
- there is no unit test in place to catch regressions or issues
- the current implementation isn't a performance concern
- I think it is more accurate to have item spacing on all sides, and am wary of shifting it to two sides because that may affect cases I am not aware of

apurbalal:
@elihart Thanks for taking a look. I tested it on the example samples and here are the screenshots. 
<table>
<tr>
<td>Before</td>
<td>After</td>
</tr>
<tr>
<td></td>
<td></td>
</tr>
</table>
Both looks same. I am not sure why "Add button" height got stretched on master branch.

Right now, for each item we are checking 4 directions. So if we have 400 items, it's about 1600 checks which can be reduced to 800 checks IMO when layout is not reversed.

elihart:
```
So if we have 400 items, it's about 1600 checks which can be reduced to 800 checks IMO when layout is not reversed.
```

are you actually showing all 400 items? Decorations are only applied for items on screen, which is likely much less than that. Even so, these are just boolean logic checks and basic arithmitic which is not going to impact performance

PR #784, opened 2019-07-13
Fix item spacing decorator

### LINKED ISSUE #373: EpoxyItemSpacingDecorator adding extra bottom spacing for certain items
See the following screenshots. The items in the middle of the row have extra bottom padding for some reason.

[Screenshot 1](https://user-images.githubusercontent.com/227486/34468656-8c6c8ea4-ef05-11e7-84d6-5437ec86bbe7.png)
[Screenshot 2](https://user-images.githubusercontent.com/227486/34468657-8c824b90-ef05-11e7-8588-685632bc5425.png)

Mori-Atsushi:
It fix #373

EpoxyItemSpacingDecorator has checked the item in the edge, and added / removed padding.
However,it has caused the problem that the items of girdview are not equally spaced.

This solves them by adding negative padding to the recycler view.
It is calculated to hold default padding.


elihart:
Hi, thanks for the attempt but I don't prefer this approach because:
1. using negative padding isn't a good pattern imo
2. Ideally the item space decorator would work on all recyclerviews, not just EpoxyRecyclerView
3. Ideally the item spacing would be solved in just the decorator

Mori-Atsushi:
Thank you for your comment.
I will propose in PR if I come up with an idea that does not use negative padding.

Issue #818, opened 2019-09-05
Visibility values after rebuild


AgneseBussone:
Hello!
I need to keep track of the first time a some data is fully visible to the user (it doesn't matter if it's hidden and shown again multiple time later) and I'm using a `OnModelVisibilityChangedListener` added to the model like this
```
new OfferModel_().id(offer.getId())
.onVisibilityChanged((model, view, percentVisibleHeight, percentVisibleWidth, heightVisible, widthVisible) -> {
  if (percentVisibleHeight == 100 && percentVisibleWidth == 100) {
    callbacks.resultViewFullyVisible(offer.getId());
  }
}).addTo(this);
```
The first time the recyclerview is shown everything is working as expected. The problem is that when the data set is changed and the models are rebuilt, some views have an initial visibility different from the final, like 100% and 7% a few ms later. In this case, the new data are tracked as visible when in reality they're only partially visible and shouldn't be tracked.
I think it's due to the recycling of the views, but I don't know if this is the intended behaviour and how to avoid this.

I did some logging on the visibility listener to better explain this. The logs contain 
- `id` related to the data shown in the view, 
- `heightVisible`, 
- `percentVisibleHeight`,
- `view.hashCode()` (to check if the view was used before, not sure if it's the right thing to use for that though).
```
First time the data are shown: looks right
14:18:23.053  D/TEST - onVisibilityChangedListener: b12cxEGT, height=290 (100%) hash=62627501
14:18:23.054  D/TEST - onVisibilityChangedListener: 2c27QurT, height=290 (100%) hash=134837730
14:18:23.055  D/TEST - onVisibilityChangedListener: bf4fFYoT, height=290 (100%) hash=213794163
14:18:23.056  D/TEST - onVisibilityChangedListener: c30bVurT, height=290 (100%) hash=229336112
14:18:23.057  D/TEST - onVisibilityChangedListener: 1e10QYoT, height=126 (43%) hash=127957161

The recyclerview is hidden before the new data set is loaded, so every view becomes invisible. Seems right.
14:18:48.681  D/TEST - onVisibilityChangedListener: 1e10QYoT, height=0 (0%) hash=127957161
14:18:48.684 D/TEST - onVisibilityChangedListener: c30bVurT, height=0 (0%) hash=229336112
14:18:48.686 D/TEST - onVisibilityChangedListener: bf4fFYoT, height=0 (0%) hash=213794163
[...]

PR #826, opened 2019-09-11
Fix for #818 - visibility tracking with recycling

eboudrant:
The bug is because, when a view is recycled, the `onChildViewAttachedToWindow` was send an event right away. On some recycling use case it happen before the view get laid out to the new data making the event is wrong as it still contains attributes of the recycled view.

I was able to reproduce the issue in unit test (Robolectric) by removing the item animator. That makes the recycler poll also used during tests. The idea of the new test is to rebuild the data set several time and make sure than always 1 visibility changed event is sent (as no scroll).

This should fix https://github.com/airbnb/epoxy/issues/818


eboudrant:
I will close the PR for now, it need more testing.

Issue #164, opened 2017-03-22
Support for sticky headers


elihart:
We are working on this, and hope to release support in a 2.x release.

There are two ways to support this.
1. An item decoration to copy the view of a model. This is simpler, but you lose touch functionality. This would be what we initially support.
2. Using a custom layout manager to create a fully functional duplicate view. There are some libraries that use this approach, but I am hesitant to pursue it because layout managers are very complex and using Linear/GridLayoutManager is the safest/optimized approach. I have an idea on how we might extend LinearLayoutManager though, which I can hopefully work on over the next few months.

If anyone is interested in helping with these, please let me know!

VisheshVadhera:
Hey @elihart, we just chatted over on reddit. 

I think the second approach definitely sounds interesting and can be a good starting point. Can you please elaborate a bit more on the second approach?

elihart:
Hi! Sure thing. I'm basically thinking that a custom layout manager can create the view that we want to sticky, and keep it positioned at the top of the screen. It would scroll away as another sticky view pushed it off screen.

There are similar approaches out there. https://github.com/TonicArtos/SuperSLiM seems like the best one. The problems I have with this are: 

1. It's not immediately usable by epoxy since we have our own models to interface with. This means we would have to fork it and manage it
2. It's completely custom so we lose all of the default behavior of the LinearLayoutManager.

I think the Android team has done a great job with the default layout managers. Those classes are very complicated and handle a ton of situations for us. I really don't want a custom layout manager because I think it would be worse and harder to maintain.

So, ideally we can make a layout manager that extends LinearLayoutManager, and just adds a little bit of functionality to figure out which view should be stuck to the top and keep it there. As far as I can tell so far we could need a custom view to do this - right now I am thinking a framelayout with a nested RecyclerView. Our layoutmanager would get the sticky view from the recyclerview and add it to the framelayout. It can't add it to the recyclerview because that would mess with the LinearLayoutManager internals which expects the child views to be a certain way.

Then our layoutmanager would simply translate the header as needed to scroll it off and replace it with the next header. An interface on a model would indicate that it wants to make its view sticky.

I played around with this idea for a day and it seems possible. You can see my (very rough) code at https://github.com/airbnb/epoxy/compare/eli/sticky

I think this approach can work fairly nice. Attaching the view to a framelayout is a bit hacky, but as far as I can tell it will work.

[...]

PR #842, opened 2019-10-09
Add support for Sticky Headers

AkshayChordiya:
**Description**
- Add sticky header linear layout manager to the library
- Add sample for sticky header using `EpoxyAdapter`
- Implement sticky header functions in `BaseEpoxyAdapter` and `EpoxyController` with no-op as default behaviour so the sticky headers are available for both `EpoxyController` and `EpoxyAdapter`, YAY 🎉 
- Update the `StickyHeaderLinearLayoutManager` to use `BaseEpoxyAdapter` so it can be used with Epoxy controller or adapter
- Update the sample app to showcase sticky header with `EpoxyController`

Solves issue #164, please take a look at the issue thread to understand different solutions. 

**Note**
- The `experimental` flag was added for `Parcelize`. It feels a bit overkill only for 1 use-case, but let me know your views

-----

Kindly feel free to provide suggestions to improve this or point potential issues. I'd also love to update the Wiki with a new section for sticky header once the PR is merged.

Inspired from [StickyHeadersLinearLayoutManager](https://github.com/Doist/RecyclerViewExtensions/blob/master/StickyHeaders/src/main/java/io/doist/recyclerviewext/sticky_headers/StickyHeadersLinearLayoutManager.java)


Nekromancer:
Hi,

Are there any perspectives to adapt this to `EpoxyController` in the near future?

nksaroj:
Can this be merged please?

akvashi:
isn't EpoxyAdapter now Deprecated? probably need a way to declare this from the controller?

Nekromancer:
@akvashi you can look at my implementation few posts earlier. It works pretty well

akvashi:
@Nekromancer so sorry i appearently didn't read anything before i made my comment. @AkshayChordiya

AkshayChordiya:
@akvashi 
Epoxy Adapter is not deprecated, it's a legacy component and not recommended by Epoxy team [AFAIK]. 
But we still use it internally for fine control over the list, to do things which are not possible by `EpoxyController` due to it's design.

@Nekromancer I can take your gist and add to my PR so we have sticky header for both the things. WDYT?

akvashi:
@AkshayChordiya i always took legacy to mean deprecated since i can't see why they would keep support beyond a certain point. I am curious what sort of things you'd use adapter over controller for

Nekromancer:
@AkshayChordiya sure, if you're fine with my solution you could merge it

akvashi:
@Nekromancer i think that maybe its worth having the interface be have something like `var isStickyHeader: Boolean`

my usecase is having multiple of the same epoxy models in the list. but one of them becomes sticky. 

this way you dont have duplicate a model. one that is sticky and one that isn't with all the other attributes being the same :)

Nekromancer:
@akvashi yes, this seems reasonable

AkshayChordiya:
@akvashi We use for notifying our models about small changes kinda like `notifyItemChanged(payload)` so it just binds the changed thing. We tried the same with controller but the performance was not 60 fps.

@Nekromancer I'll take it as an inspiration and update the PR.

Thanks for suggestions

akvashi:
@AkshayChordiya not sure if this useful as the other attribute i placed on IStickyHeader, but I also have added a "isSticking" which i flip true/false during setup/teardown of the sticky header. 

this way using the EpoxyModel itself still holds all the code for transforming the view once it becomes a header. 

usecase: have a view all of the same epoxy model type. some of them have isStickyHeader set to true. once they become a header needed to slightly change some things on the model (in my case was removal of a click listener)

I can post a gist if that would helpful :)

akvashi:
might be beyond the scope of this PR. since i could implement using just the base classes provided with a mix of the code you and @Nekromancer provided

kssc0112:
@Nekromancer Thanks for your code for the controller integration. I have taken your gist and tested on my project which is currently using EpoxyController. What I am seeing is that the header is correctly sticking to the top and gets replaced by next header item as I scroll. The only issue is that the header gets hidden by the other non-IStickyHeader models (It is on the very top but sits behind other items in the recyclerview).

AkshayChordiya:
@kssc0112 I think you need to provide an elevation to your header model. 

Use the `setupStickyHeaderView` function to set the elevation to the view and the `teardownStickyHeaderView` to remove it.

AkshayChordiya:
I updated the PR with support for `EpoxyController` and `EpoxyAdapter` 🎉 💃

nksaroj:
Can this me merged soon please?

AkshayChordiya:
@nksaroj I think @elihart is looking into it, the reason it was delayed because of ownership of the `StickyHeaderLinearLayoutManager` for bug fixes, improvements, etc.

jvillad1:
Is there any update regarding the merge? This is an awesome feature and would be great if this was part of Epoxy, thanks!

elihart:
I'll try to get to this as soon as I can

premacck:
+1 on this feature 😅

epoxy-adapter/src/main/java/com/airbnb/epoxy/stickyheader/StickyHeaderLinearLayoutManager.kt
```diff
@@ -0,0 +1,621 @@
...
+    private var scrollOffset = 0
+
+    override fun onAttachedToWindow(recyclerView: RecyclerView) {
+        super.onAttachedToWindow(recyclerView)
+        recyclerView.adapter?.let(this::setAdapter)
+    }
+
+    override fun onAdapterChanged(oldAdapter: RecyclerView.Adapter<*>?, newAdapter: RecyclerView.Adapter<*>?) {
+        super.onAdapterChanged(oldAdapter, newAdapter)
+        newAdapter?.let(this::setAdapter)
```
elihart, line 51: should anything be cleared when this newAdapter is null?

epoxy-adapter/src/main/java/com/airbnb/epoxy/stickyheader/StickyHeaderLinearLayoutManager.kt
```diff
@@ -0,0 +1,621 @@
...
+import android.view.ViewGroup
+import android.view.ViewTreeObserver
+import androidx.recyclerview.widget.LinearLayoutManager
+import androidx.recyclerview.widget.RecyclerView
+import com.airbnb.epoxy.BaseEpoxyAdapter
+import kotlinx.android.parcel.Parcelize
+
+/**
+ * Adds sticky headers capabilities to your [RecyclerView.Adapter].
+ * The adapter must implement [HasStickyHeader] to indicate which items are
```
elihart, line 17: re `The adapter must implement [HasStickyHeader]` is this necessary to call out since epoxy base classes already implement it? is it confusing to mention this?

epoxy-adapter/src/main/java/com/airbnb/epoxy/stickyheader/HasStickyHeader.kt
```diff
@@ -0,0 +1,39 @@
+package com.airbnb.epoxy.stickyheader
+
+import android.view.View
+import androidx.recyclerview.widget.RecyclerView
+
+/**
+ * Adds sticky headers capabilities to any [RecyclerView.Adapter]
+ * combined with [StickyHeaderLinearLayoutManager].
+ */
+interface HasStickyHeader {
```
elihart, line 10: I think `HasStickyHeader` naming doesn't feel quite right, since this interface does a bit more than that. maybe something like `StickyHeaderCallbacks`

epoxy-adapter/src/main/java/com/airbnb/epoxy/stickyheader/StickyHeaderLinearLayoutManager.kt
AkshayChordiya: Good point. Now that I think about it, yes we should be clearing few things if the adapter is null.
I'll update it 👍

epoxy-adapter/src/main/java/com/airbnb/epoxy/stickyheader/StickyHeaderLinearLayoutManager.kt
AkshayChordiya: Makes sense. I'll reword it. Initially the base classes were not implementing it hence it was mentioned.

epoxy-adapter/src/main/java/com/airbnb/epoxy/stickyheader/HasStickyHeader.kt
AkshayChordiya: Yeah it's not the name which I would prefer as first choice. `StickyHeaderCallbacks` seem good to me, any other suggestion which come to your mind?

AkshayChordiya:
> @AkshayChordiya thanks for all your work on this and being patient. finally ready to merge it, just have some minor questions first

Thank you so much @elihart for reviewing this. I'll address the comments and update the PR.
I have a question if I should update the Wiki too after this PR is merged?

epoxy-adapter/src/main/java/com/airbnb/epoxy/stickyheader/HasStickyHeader.kt
elihart: nope, I think that's great

elihart:
@AkshayChordiya great, thanks! I'll make a new release soon after this is merged once you get the updates in.

 it's not possible for external contributors to edit the wiki, but if you'd like to share content here that'd you like to have in the wiki for it I'm happy to copy it in, otherwise I'll write something up myself. I'm planning to have a dedicated page for it linked from the wiki sidebar

epoxy-adapter/src/main/java/com/airbnb/epoxy/stickyheader/HasStickyHeader.kt
AkshayChordiya: So should I keep it `HasStickyHeader`?

elihart: sorry, I meant `StickyHeaderCallbacks` seems good

epoxy-adapter/src/main/java/com/airbnb/epoxy/stickyheader/StickyHeaderLinearLayoutManager.kt
AkshayChordiya: Turns out, the adapter / controller needs to implement the `isStickyHeader` function so I'll keep it comment as it is and just tweak it a bit and add a example usage in the code docs

epoxy-adapter/src/main/java/com/airbnb/epoxy/stickyheader/StickyHeaderLinearLayoutManager.kt
```diff
@@ -43,20 +43,25 @@ class StickyHeaderLinearLayoutManager @JvmOverloads constructor(
...
-        headerPositionsObserver.onChanged()
+    private fun setAdapter(newAdapter: RecyclerView.Adapter<*>?) {
+        adapter?.unregisterAdapterDataObserver(headerPositionsObserver)
+        if (newAdapter is BaseEpoxyAdapter) {
+            adapter = newAdapter
+            adapter?.registerAdapterDataObserver(headerPositionsObserver)
+            headerPositionsObserver.onChanged()
+        } else {
+            adapter = null
+            headerPositions.clear()
```
elihart, line 72: `headerPositionsObserver.onChanged()` doesn't need to be called when clearing the positions?

elihart:
Thanks for the clean ups!

epoxy-adapter/src/main/java/com/airbnb/epoxy/stickyheader/StickyHeaderLinearLayoutManager.kt
AkshayChordiya: It essentially clears the `headerPositions` and does some extra things which are not necessary when the adapter is null. Hence it's safe to just directly call `headerPositions.clear()`

parthm1015:
calls to onbind and onunbind is not working as expected for model immediately below sticky header. If we scroll down and model below sticky header is off the screen, then also calls to onBind and onUnbind of that model keeps happening. Does anyone else facing the same issue?

PR #848, opened 2019-10-22
Update Gradle plugin and wrapper

friederbluemle:
Two quick updates to the latest stable versions:

- Android Gradle plugin `3.5.1`
- Gradle wrapper `5.6.3`


gradle.properties
```diff
@@ -16,7 +16,6 @@ POM_INCEPTION_YEAR=2016
 android.useAndroidX=true
 android.enableJetifier=false
 
-org.gradle.daemon=true
```
friederbluemle: Note: This _defaults_ to true. There is no need to set it (it was only required for an older Gradle version).

PR #864, opened 2019-11-05
Remove executable bit from non-executable files

friederbluemle:
A bunch of non-executable files in the repo erroneously had the executable bit set. This fixes is by changing the file mode from `755` to `644`.


elihart:
Good catch, thanks!

PR #891, opened 2019-12-14
EpoxyVisibilityTracker: detach detached child recycler view from tracker

Ethan1983:


elihart:
Can you post a description of the issue, and ideally update the tests to catch this issue?

PR #898, opened 2019-12-23
Fix incorrect license attributes in POM file

chibatching:
License attributes in POM are null from 3.8.0 to 3.9.0.
https://repo1.maven.org/maven2/com/airbnb/android/epoxy/3.9.0/epoxy-3.9.0.pom

This is caused by incorrect gradle property names for gradle-maven-publish-plugin.
ref: https://github.com/vanniktech/gradle-maven-publish-plugin/blob/master/gradle.properties

I matched properties with correct names.


elihart:
thanks for catching this!

PR #899, opened 2019-12-30
Rename ResultCallack to ResultCallback in AsyncEpoxyDiffer

HealGaren:
I think `ResultCallack` is a typo.
If the word meant `call acknowledgement`, it would be right to write `ResultCallAck`.. and it seems to mean callback.


elihart:
Yes, this was a typo, thanks for catching it!

PR #900, opened 2019-12-31
Add support full span for staggered grid layout manager

dinhlamvn:
At the moment, **epoxy** only support for **GridLayoutManager**, but sometime we use **StaggeredLayoutManager** and we would like to render some model is a full span model.

In this pull request, I have added a class to support to set **Model View** is full span when we use **StaggeredLayoutManager**


epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModelStaggeredGrid.java
```diff
@@ -0,0 +1,60 @@
...
+   */
+  private boolean isFullSpanModel = false;
+
+  /**
+   * The method to set a model view can be full span or not.
+   *
+   * @param isFullSpanModel if true, the model view will be full span.
+   * @return The model view.
+   */
+  public EpoxyModelStaggeredGrid<T> setFullSpanModel(boolean isFullSpanModel) {
```
elihart, line 28: I think this needs to be marked as a model prop, otherwise toggling it will not trigger a view update

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModelStaggeredGrid.java
```diff
@@ -0,0 +1,60 @@
...
+    this.isFullSpanModel = isFullSpanModel;
+    return this;
+  }
+
+  @Override
+  public void bind(@NonNull T view) {
+    if (isFullSpanModel) {
```
elihart, line 35: do you need to clear the params change when this is false, to undo it?

dinhlamvn:
@elihart Thanks for your review. Finally, for better case I think we should make the model view is full span or not when we create view holder for it. So I have updated my PR, please check it.

dinhlamvn:
@elihart Alright, you are right, after thinking I also feel that is not a good practice. I will update my PR.

Issue #886, opened 2019-12-03
Possible memory leak in ActivityRecyclerPool when EpoxyRecyclerView is created with a ContextWrapper


rashadsookram:
I came across this when going through an app with Leak Canary 2. This was happening with an `EpoxyRecyclerView` being inflated within a `DialogFragment` using the inflater passed to `onCreateView`. This `LayoutInflater` had its `Context` set as a `ContextThemeWrapper` because it uses the [`Dialog`'s `Context`](https://android.googlesource.com/platform/frameworks/support/+/2d95dad0be72c1aa45c5a0844c67bb727403bcea/fragment/fragment/src/main/java/androidx/fragment/app/DialogFragment.java#408), which is a `ContextThemeWrapper` (see [1](https://android.googlesource.com/platform/frameworks/support/+/2d95dad0be72c1aa45c5a0844c67bb727403bcea/fragment/fragment/src/main/java/androidx/fragment/app/DialogFragment.java#455) and [2](https://github.com/aosp-mirror/platform_frameworks_base/blob/ad3e08aa693ee0e5936ce6cce2b6036a20554c31/core/java/android/app/Dialog.java#L181-L191)).

Looking at Epoxy's source, it seems to be because of [this code](https://github.com/airbnb/epoxy/blob/3.8.0/epoxy-adapter/src/main/java/com/airbnb/epoxy/ActivityRecyclerPool.kt#L53) which results in the `LifecycleObserver` that cleans things up not being registered.

Let me know if you need any more details, but I think this could be fixed by changing it to recursively unwrap the `Context` to try and find the `LifecycleOwner` / `Activity`.

## Leak Canary Trace

The TL;DR of the trace below is:

`EpoxyRecyclerView.ACTIVITY_RECYCLER_POOL` (static) retains
a `ActivityRecyclerPool` which retains
a `ViewHolder` which retains
a `Context` (specifically the `ContextThemeWrapper` which is wrapping the Activity)

### Full Trace

  ```
  612733 bytes retained
┬
├─ android.os.FileObserver$ObserverThread
│    Leaking: NO (PathClassLoader↓ is not leaking)
│    Thread name: 'FileObserver'
│    GC Root: Input or output parameters in native code
│    ↓ thread FileObserver$ObserverThread.contextClassLoader
├─ dalvik.system.PathClassLoader
│    Leaking: NO (Object[]↓ is not leaking and A ClassLoader is never leaking)
[...]

PR #906, opened 2020-01-08
Fix leak in ActivityRecyclerPool when EpoxyRecyclerView is created with a ContextWrapper

rashadsookram:
Fixes https://github.com/airbnb/epoxy/issues/886 by unwrapping `Context`s to find the `LifecycleOwner` / `Activity` as appropriate.


rishirajrandive:
@elihart This issue is causing a leak in our app as well. Any idea when will be the next version released with this fix?

Issue #682, opened 2019-02-12
Carousel numViewsToShowOnScreen and config changes


mradzinski:
Hey @elihart, I've been trying to set `numViewsToShowOnScreen` on a Carousel to a percentage of the display width (which gets updated upon config changes). To achieve this, what I'm doing is quite simple:

```
numViewsToShowOnScreen((17 * screenWidthPx / 100) / 100) // Where 17 is the percentage visible
```

This works well upon rendering the first time, but it breaks as soon as I rotate the device... and if I scroll through the carousel it seems to fix itself (probably due to recycling of the modes). I'm attaching a short GIF of what's happening.

![carousel_small](https://user-images.githubusercontent.com/4214209/52650839-c0d50180-2ec9-11e9-8ca5-718802d5e197.gif)

Any idea of what might be happening? I've tried invalidating the RecyclerView and the Carousel itself (during onBind), also assigning `Random` id's to force Epoxy to rebuild the models, but none of this seem to solve the issue.

mradzinski:
I ended up solving this through using `WRAP_WIDTH_WRAP_HEIGHT` on the `@Model` and setting a padding on the `CarouselModel_`. 

Still not sure if `numViewsToShowOnScreen` should behave this way though. If this behavior is to be expected then feel free close the issue @elihart.

elihart:
Thanks for the video - are you handling the configuration change yourself, or is the activity being recreated on rotation? if it is being recreated I don't see how the views would remember their old width

mradzinski:
@elihart I don't use activities, but the view containing the RV re-inflates and has a different instance, so I think it would count as "being recreated". I'm quite sure of this given the instances of the RV on config changes are different too (hash is different). 

I think it has something to do with recycling/rendering. The moment I scroll and "force" the items to recycle the issue is gone.

elihart:
In a situation where rotation causes an activity recreation this should work as expected, and in general `numViewsToShowOnScreen` should update across recreation.

I am not familiar with how you are using it without an activity though and don't know how that affects view recreation. if you'd like to look into it you can see how Carousel#numViewsToShowOnScreen is being called, since that has the view resizing logic

mradzinski:
Agreed. I'll check the class and will try finding the reason behind this based on that. I'll let you know if I find anything reproducible using activities or fragments. In the meantime let's close this issue to keep the backlog clean, I'll open a new one if needed.

Thanks again for your help Eli and cheers!

[...]

PR #915, opened 2020-02-04
Carousel - Use measured width/height if it exists.

ekimmel-ta:
My carousel component is not full width.  The current carousel implementation was defaulting to the full screen width because layout had not completed yet.

This PR adds measured width / measured height to the getTotalWidthPx and getTotalHeightPx functions in Carousel to help prevent a situation where the carousel is not using the intending width.

### DIFF
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/Carousel.java
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/Carousel.java
@@ -239,6 +239,10 @@ public class Carousel extends EpoxyRecyclerView {
       return view.getWidth();
     }
 
+    if (view.getMeasuredWidth() > 0) {
+      return view.getMeasuredWidth();
+    }
+
     // Fall back to assuming we want the full screen width
     DisplayMetrics metrics = view.getContext().getResources().getDisplayMetrics();
     return metrics.widthPixels;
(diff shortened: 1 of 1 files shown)


elihart:
Sounds good to me :)

andybdahl:
This merge created the issue mentioned in the issue #682 . A proper layout should be done by only using the view's width / measured height as long as the orientation stays the same, and should default to the screen width OR measure the Carousel once again after configuration changes has been made.

elihart:
@andybdahl thanks for catching this. are you interested in contributing that fix?

andybdahl:
@EliHart
We fixed it locally in our app already, I'll be putting up a merge request
when I get the time for it soon.

ekimmel-ta:
Hi, could you explain how this PR affects your issue?  Epoxy was already using this.width previously, and as you said that value is wrong during configuration changes.  

I did find that Epoxy reusing a Carousel model with a different width, even without configuration changes, is still broken.  Width is the old value and measuredWidth is the correct, updated value.  I don't know how we could detect that width is out of date.

ekimmel-ta:
Slight update: Using width before measuredWidth is incorrect if this function is called during onMeasure (meaning during onChildAttachedToWindow as well).  I don't know if that means we should be using measuredWidth only or if we need to flag when we're in the measurement phase.

Issue #847, opened 2019-10-21
Carousel building with Kotlin DSL


eboudrant:
Regarding the nice Kotlin  DSL we have to build model, we still need to use the model constructor for carousels (ie: `PhotoViewModel_()`) : 
```
fun buildModels() {

   val photoModels = mutableListOf<PhotoViewModel_>()
   photos.forEach {
      photoModels.add(PhotoViewModel_().apply {
        id(photo.id())
        url(photo.url))
      }
   }

   carousel {
        id("carousel")
        numViewsToShowOnScreen(5)
        models(photoModels)
    }
}
```

Does a DSL for building carousel was ever considered or is there a known limitation? This could be much more simple to write/read : 

```
fun buildModels() {
   carousel {
        id("carousel")
        numViewsToShowOnScreen(5)
        photos.forEach {
            photoView {
                 id(photo.id())
                 url(photo.url))
            }
        }
    }
}
```

elihart:
Yes, this would be ideal and has been brought up before. It is possible - the problem is that the extension function to create the model uses an EpoxyController as a receiver, so we need a syntax that changes the receiver to be a separate EpoxyController that we then pull the models off of.

It would be more feasible to do something like this
```
   carousel {
        id("carousel")
        numViewsToShowOnScreen(5)
        models(buildModels {
           photos.forEach {
             photoView {
                 id(photo.id())
                 url(photo.url))
            }
        }
      }
    }
```

This allows us to change the receiver. The only way to do it in your example would be to create a special extension function `carousel` that uses that receiver. This would conflict with the default `carousel` unless the annotation processor was able to replace the default, which is possible but just gets more complex

elihart:
I think https://github.com/airbnb/epoxy/pull/850 accomplishes the syntax you demonstrated.

It should work - tasks left are
- Make sure it doesn't break existing usages
- Decide whether to ship with a different function name than `carousel`, or update our code replace the generated carousel function with this custom one somehow
- Think about providing an easy pattern for users with custom carousels (or other nested models, like possibly EpoxyModelGroup) to also apply this approach

eboudrant:
Nice, I will check it out !

apramana:
Hi @elihart - are there any plans to revive this improvement? Also, would it make sense for Carousel to use `withModels` with `@ModelProp` so that callbacks are used through the entire chain? Or maybe there's a technical constraint I'm missing here.

elihart:
[...]

PR #967, opened 2020-05-12
Carousel building with Kotlin DSL - Step 1

eboudrant:
This is a refresh from @elihart's pull request #850 about issue issue #847.

While we certainly need a more robust api so we could use the DSL for custom carousel (and Epoxy model) group I propose we do it step by step. 

Where the first step is we merge this change : 
- Create a `ModelCollector` with required change in code-gen
- Add an example of builder in the sample app `EpoxyCarouselNoSnapBuilder.kt`
- Update the sample app to use the carousel DSL

```
carouselNoSnap { 
  id(...)
  models(mutableListOf<CarouselItemCustomViewModel_>().apply {
    for (...) {
      add(
        CarouselItemCustomViewModel_()
          .id(...)
        )
      }
  })
}
```

```
carouselNoSnapBuilder {
  id(...)
  for (...) {
    carouselItemCustomView {
      id(...)
    }
  }
}
```


ameyab10:
@eboudrant Thanks for the PR :)

PR #972, opened 2020-05-15
Incremental annotation processor and processor optimizations

elihart:
Fixes https://github.com/airbnb/epoxy/issues/423

- Adds incremental annotation support via Aggregrating mode. Isolating is not possible without removing functionality
- Removes Litho, since it was never used afaik, was very long out of date, and not worth our time to maintain it
- The single EpoxyProcessor annotation processor class is split into a separate processor for each type of annotation
- Several optimizations and memoizations were applied to the processors to speed up their processing time
- A new annotation processor argument `logEpoxyTimings` can be set to get a detailed breakdown of how long the processors took and where they spent their time (off by default)
- Another new argument `enableParallelEpoxyProcessing` can be set to true to have the annotation processor process annotations and generate files in parallel (via coroutines). This can greatly speed up processing time, but given the nature of parallel processing it is still incubating

With these optimizations and parallel processing, clean builds of a model with 165 @ModelView classes went from 4 seconds to 0.5 seconds. A module with 70 @EpoxyModelClass annotated classes also had a 27% decrease in processing time, also now to 0.5 seconds. We don't yet have any profiling for how much difference the incremental compilation makes, but these other optimizations are likely more impactful.

I will release this as a beta with a major release number


epoxy-processor/src/main/java/com/airbnb/epoxy/processor/BaseProcessor.kt
```diff
@@ -0,0 +1,166 @@
...
+            val processorName = this@BaseProcessor::class.java.simpleName
+            // We wait until the very end to log errors so that all the generated classes are still
+            // created.
+            // Otherwise the compiler error output is clogged with lots of errors from the generated
+            // classes  not existing, which makes it hard to see the actual errors.
+            logger.measure("validateAttributesImplementHashCode") {
+                validateAttributesImplementHashCode(generatedModels)
+            }
+
+            logger.writeExceptions()
```
BenSchwab, line 125: Probably not necessary but just a thought - it seems like normal exceptions are pretty tractable and likely can be gracefully lazily collected. 

Did you find any super weird output with concurrent exceptions like cme's while working on the parallelization? I could imagine that those might leave the program in a weird state and create tons of exceptions. Specially handling them might also be nice to print a message like "CME found, consider disabling experimental parallelization feature and file a ticket with info).

epoxy-processor/src/main/java/com/airbnb/epoxy/processor/BaseProcessorWithPackageConfigs.kt
```diff
@@ -0,0 +1,64 @@
...
+import javax.annotation.processing.RoundEnvironment
+import javax.lang.model.element.Element
+import kotlin.reflect.KClass
+
+abstract class BaseProcessorWithPackageConfigs : BaseProcessor() {
+
+    abstract val usesPackageEpoxyConfig: Boolean
+    abstract val usesModelViewConfig: Boolean
+
+    override fun supportedAnnotations(): List<KClass<*>> = mutableListOf<KClass<*>>().apply {
```
BenSchwab, line 14: nit: Maybe a little safer to make this final, and have an abstract `additionalSupportedAnnotations` as it seems like we can't use `@CallSuper` or something like that

epoxy-processor/src/main/java/com/airbnb/epoxy/processor/BaseProcessorWithPackageConfigs.kt
```diff
@@ -0,0 +1,64 @@
...
+            add(PackageEpoxyConfig::class)
+        }
+        if (usesModelViewConfig) {
+            add(PackageModelViewConfig::class)
+        }
+    }
+
+    /**
+     * Returns all of the package config elements applicable to this processor.
+     *
```
BenSchwab, line 25: Bunch of extra space here

epoxy-processor/src/main/java/com/airbnb/epoxy/processor/GeneratedModelInfo.kt
```diff
@@ -0,0 +1,279 @@
...
+     * view programmatically
+     * instead of via xml layout resource inflation.
+     */
+    var layoutParams = ModelView.Size.NONE
+
+    /**
+     * The elements that influence the generation of this model.
+     * eg base model class for @EpoxyModelClass, view class for @ModelView, etc
+     */
+    open fun originatingElements(): List<Element> {
```
BenSchwab, line 64: Similar - might be nice to keep this private and call into an abstract `originatingElements` methods

epoxy-processor/src/main/java/com/airbnb/epoxy/processor/BaseProcessor.kt
elihart: interesting idea. I did find `CME` sometimes, or a specific exception from ClassReader. i like the idea of checking for that to tell users to disable parallelization and to report the stacktrace. i didn't find any times where the processor got into a weird state because of it where there were downstream errors

elihart:
I've pushed a `4.0.0-beta1` release if anyone wants to use this right now

TaxistSamael:
Thank you a lot for this optimization!

PR #973, opened 2020-05-15
Introduce partial impression visibility states

### LINKED ISSUE #838: Handle partial visibility state
Requirements for our project have visibility tracking of elements at 50% visibility. Since both `VisibilityState.FOCUSED_VISIBLE` and `VisibilityState.FULL_IMPRESSION_VISIBLE` do not cover this state, we were thinking of extending the visibility tracker.

The idea would be to offer some api to define a threshold (range 0..1) for when a visibility state would be triggered. This could either be applied to the `VisibilityState.FULL_IMPRESSION_VISIBLE`, which would redefine what a full impression is in the app context, or preferably introduce a new state `VisibilityState.PARTIAL_IMPRESSION_VISIBLE`. Implementation would be very similar to `VisibilityItem#handleFullImpressionVisible`.

Now we could probably handle this using the `visibilityChanged`, but it would be more involved to not double-count the impression and it would trigger quite often (so we would not be able to disable the noisy onChanged (`visibilityChangedEnabled`).

I think there might be more use-cases which would warrant to have a single-event impression for a certain threshold of visibility. If you think it might be useful, I will whip up a PR within the next weeks.

mediavrog:
fixes #838

- introduces `PARTIAL_IMPRESSION_VISIBLE` and `PARTIAL_IMPRESSION_INVISIBLE` visibility states
- threshold for partial impression can be set via `EpoxyVisibilityTracker#setPartialImpressionThresholdPercentage`
- the threshold will be automatically applied to nested trackers as well if they do not have their own tracker set
- no event is fired if the threshold cannot be met due to the element being larger than the view port and the set threshold too high to trigger. This is consistent with the behavior of `FULL_IMPRESSION_VISIBLE`
- updated `:kotlinsample` to reflect the addition

### Tests
- I added some tolerance to the AssertHelper:
- TOLERANCE_PIXELS (value=1) was currently not being tolerant, as it was compared using `<` instead of `<=`
- for visibleHeight and visibleWidth I added a delta to cover the 49% case to test for an element that got VISIBLE, but not PARTIAL_IMPRESSION_VISIBLE

As for tests, since setting the threshold will add a new dimension to the test matrix, I was not sure if I should adjust the whole testsuite to cope for this. Basically all of

testDataAttachedToRecyclerView, testScrollToPosition, testMoveDataDown, testInsertData, testScrollBy, testDeleteData, testMoveDataUp

could be tested once with a threshold set, and once without.
Please let me know if I should cope for this.

### Wiki Content Suggestion
Visibility Events

- **Partial Impression Visible**: Triggered when the percentage of the visible area of a view is
   at least the specified threshold. This threshold can be set via `EpoxyVisibilityTracker#setPartialImpressionThresholdPercentage`. This event does not fire if the threshold is set to `null` or if the threshold cannot be met due to the element being larger than the viewport and the set threshold too high.
- **Partial Impression Invisible**: Triggered when the percentage of the visible area of a view is
   no longer at least the specified threshold. Also see **Partial Impression Visible**.

### DIFF
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.java
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.java
@@ -114,6 +114,7 @@ public class EpoxyVisibilityTracker {
       @IntRange(from = 0, to = 100) int thresholdPercentage
   ) {
     partialImpressionThresholdPercentage = thresholdPercentage;
+    String heelo = "";
   }
 
   /**
@@ -310,6 +311,7 @@ public class EpoxyVisibilityTracker {
     EpoxyVisibilityTracker tracker = getTracker(childRecyclerView);
     if (tracker == null) {
       tracker = new EpoxyVisibilityTracker();
+      tracker.setPartialImpressionThresholdPercentage(partialImpressionThresholdPercentage);
       tracker.attach(childRecyclerView);
     }
     nestedTrackers.put(childRecyclerView, tracker);
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.java
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.java
@@ -82,6 +82,8 @@ public class EpoxyVisibilityTracker {
 
   private boolean onChangedEnabled = true;
 
+  private int partialImpressionThreshold = 0;
+
   /** All nested visibility trackers */
   private Map<RecyclerView, EpoxyVisibilityTracker> nestedTrackers = new HashMap<>();
 
@@ -98,6 +100,15 @@ public class EpoxyVisibilityTracker {
    */
   public void setOnChangedEnabled(boolean enabled) {
     onChangedEnabled = enabled;
+  }
+
+  /**
+   * Set the threshold of percentage visible area to identify the partial impression view state.
+   *
+   * @param threshold Percentage of visible area of an element. 0..100.
+   */
+  public void setPartialImpressionThreshold(int threshold) {
+    partialImpressionThreshold = threshold;
   }
 
   /**
@@ -280,6 +291,7 @@ public class EpoxyVisibilityTracker {
       // View is measured, process events
       vi.handleVisible(epoxyHolder, detachEvent);
       vi.handleFocus(epoxyHolder, detachEvent);
+      vi.handlePartialImpressionVisible(epoxyHolder, detachEvent, partialImpressionThreshold);
       vi.handleFullImpressionVisible(epoxyHolder, detachEvent);
       changed = vi.handleChanged(epoxyHolder, onChangedEnabled);
     }
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.java
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.java
@@ -114,7 +114,6 @@ public class EpoxyVisibilityTracker {
       @IntRange(from = 0, to = 100) int thresholdPercentage
   ) {
     partialImpressionThresholdPercentage = thresholdPercentage;
-    String heelo = "";
   }
 
   /**
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.java
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.java
@@ -13,6 +13,7 @@ import java.util.List;
 import java.util.Map;
 
 import androidx.annotation.IdRes;
+import androidx.annotation.IntRange;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.recyclerview.widget.RecyclerView;
@@ -82,7 +83,7 @@ public class EpoxyVisibilityTracker {
 
   private boolean onChangedEnabled = true;
 
-  private int partialImpressionThreshold = 0;
+  private int partialImpressionThresholdPercentage = 0;
 
   /** All nested visibility trackers */
   private Map<RecyclerView, EpoxyVisibilityTracker> nestedTrackers = new HashMap<>();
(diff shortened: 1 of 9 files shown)


epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.java
```diff
@@ -82,6 +83,8 @@ private static void setTracker(
 
   private boolean onChangedEnabled = true;
 
+  private int partialImpressionThresholdPercentage = 0;
```
elihart, line 86: should it default to something, or default to null in which case `handlePartialImpressionVisible` is not called?

mediavrog: Good point. Let's default to `null` and not call the method as you described.

elihart: I guess I see that it shortcircuits to doing nothing when set to 0, in which case the listeners are never called. i think this is a little confusing since a threshold of `0` seems like it would always trigger it to be called. maybe use -1 or null instead?

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.java
```diff
@@ -100,6 +103,19 @@ public void setOnChangedEnabled(boolean enabled) {
     onChangedEnabled = enabled;
   }
 
+  /**
+   * Set the threshold of percentage visible area to identify the partial impression view state.
```
elihart, line 107: maybe not that the default makes partial impression events never called?

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.java
mediavrog: The other idea is to just not call `handlePartialImpressionVisible` in case of `0`

mediavrog: If using -1 to disable the events, would you prefer a separate method to disable the events then in regards to public API? Or just annotate `setPartialImpressionThresholdPercentage`'s parameter with @IntRange(from = -1, to = 100) (seems a bit odd)

mediavrog: A value of `0` remains a special case, as it would have to fire if at least one pixel of the element is visible on the screen (would act the same as `VISIBLE`).

mediavrog: How about b6d43a5b10b06c43115a6942fd84d2c9962fffb1

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.java
mediavrog: adressed also in b6d43a5b10b06c43115a6942fd84d2c9962fffb1

epoxy-adapter/src/test/java/com/airbnb/epoxy/EpoxyVisibilityTrackerTest.kt
```diff
@@ -687,6 +882,7 @@ class EpoxyVisibilityTrackerTest {
     fun setup() {
         Robolectric.setupActivity(Activity::class.java).apply {
             setContentView(EpoxyRecyclerView(this).apply {
+                epoxyVisibilityTracker.setPartialImpressionThresholdPercentage(50)
```
eboudrant, line 885: That's nice to update all the tests with a `partialImpressionThresholdPercentage` value !

I would have let the existing tests as "it is" and added more test to validate the different `partialImpressionThresholdPercentage` scenarios.

kotlinsample/src/main/java/com/airbnb/epoxy/kotlinsample/models/OnVisibilityEventDrawable.kt
```diff
@@ -12,7 +12,8 @@ import android.graphics.drawable.Drawable
  * Drawable for sample app that draw the current visibility state :
  * - circle #1 : visible
  * - circle #2 : focused
- * - circle #3 : full impression
+ * - circle #3 : partial impression
+ * - circle #4 : full impression
```
eboudrant, line 16: Thanks for updating the sample app too 👍

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.java
```diff
@@ -100,6 +103,19 @@ public void setOnChangedEnabled(boolean enabled) {
...
+   *
+   * @param thresholdPercentage Percentage of visible area of an element. 0..100. A value of 0
+   *                            effectively disables
+   *                            {@link VisibilityState#PARTIAL_IMPRESSION_VISIBLE} events.
+   */
+  public void setPartialImpressionThresholdPercentage(
+      @IntRange(from = 0, to = 100) int thresholdPercentage
```
eboudrant: I am wondering if `0` should be considered as `null`. Is it even a valid threshold for this use case?

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.java
eboudrant: How about `@IntRange(from = 1, to = 99)`, possible values would be `null` or any `int` from 1 to 99
Btw, if we have `100` we could use the full impression event so `100` is not really needed.

epoxy-adapter/src/test/java/com/airbnb/epoxy/EpoxyVisibilityTrackerTest.kt
mediavrog: Yeah that's why I asked about the test matrix. If necessary, I can write some code to dupe the cases once with and once without partial imps.

> As for tests, since setting the threshold will add a new dimension to the test matrix, I was not sure if I should adjust the whole testsuite to cope for this. Basically all of
> testDataAttachedToRecyclerView, testScrollToPosition, testMoveDataDown, testInsertData, testScrollBy, testDeleteData, testMoveDataUp
> could be tested once with a threshold set, and once without.

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.java
mediavrog: See this discussion 
https://github.com/airbnb/epoxy/pull/973#discussion_r426092299

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.java
mediavrog: I would leave this decision to you as the maintainers of the library as it is more about the public API and what is expected here.

for sure the extremes will basically duplicate existing events:
- `0` acts like `VISIBLE`
- `100` acts like `FULL_IMPRESSION_VISIBLE` (with an added `_INVISIBLE` state)

I think having the full range from 0 to 100 grew on me. I think of it this way: If other visibility events were to drop out and be removed, this one should still be fully functional and in that case 0 and 100 are valid use cases.

eboudrant: I don't have a strong opinion on it and since you handle `0` and `100` having `@IntRange(from = 0 to = 100)` is fine.

Issue #986, opened 2020-05-28
[epoxy-paging] IndexOutOfBoundsException


denis-bezrukov:
We encountered the following crash in our app:
```
java.lang.IndexOutOfBoundsException: Index: 81, Size: 81
    at androidx.paging.PagedStorage.get(PagedStorage.java:138)
    at androidx.paging.PagedList.get(PagedList.java:410)
    at com.airbnb.epoxy.paging.PagedListModelCache.getModels(PagedListModelCache.kt:200)
    at com.airbnb.epoxy.paging.PagedListEpoxyController.buildModels(PagedListEpoxyController.kt:74)
    at com.airbnb.epoxy.EpoxyController$1.run(EpoxyController.java:276)
    at android.os.Handler.handleCallback(Handler.java:815)
    at android.os.Handler.dispatchMessage(Handler.java:104)
    at android.os.Looper.loop(Looper.java:207)
    at android.os.HandlerThread.run(HandlerThread.java:61)
```

After a quick look at the lines from the stacktrace I found that access to `modelCache` in `PagedListModelCache` may be non thread safe.
```
(0 until modelCache.size).forEach { position ->
    if (modelCache[position] == null) { // <- crash here, so it's concurrent modification issue
        modelCache[position] = modelBuilder(position, currentList[position])
    }
}
```

functions that accesses to `modelCache` are marked by `@Synchronized` but synchronization  performed on different objects:
`getModels` is synchronized on PagedListModelCache instance while `onRemoved` is synchronized on `updateCallback` object. So there is no synchronization between these two functions.

Fix will be to use the same monitor object for synchronization these functions. e.g. by making 
```
override fun onRemoved(position: Int, count: Int) = synchronized(this@PagedListModelCache) {
    assertUpdateCallbacksAllowed()
    (0 until count).forEach {
        modelCache.removeAt(position)
     }
     rebuildCallback()
}
```

elihart:
thanks for the report! are you interested in putting up the fix?

PR #987, opened 2020-05-28
Synchronize ListUpdateCallback and PagedListModelCahe functions

denis-bezrukov:
Fix for #986 

I'm not familiar with all the code, but also see potential RC between `PagedListEpoxyController.addModels` fun that supposed to be overriden and ModelCache. Because addModels uses exposed mutable list without synchronization with PagedListModelCache.


elihart:
Thanks for this! good eye. I'll be including this in an upcoming 4.0 beta release

Issue #988, opened 2020-05-29
4.0.0.beta1 generating duplicate method layout(int) 


markchristopherng:
Hi 

The 4.0.0.beta1 version is generating duplicate methods on the builder. When i downgrade to 3.11.0 it goes away.

In the console i get an error like this

 method layout(int) is already defined in class EpoxyOfflineArticles_
  public EpoxyOfflineArticles_ layout(@LayoutRes int layoutRes) {

When i open up the generated class you can see the duplicated method

public interface EpoxyArticlesBuilder {
  EpoxyArticlesBuilder onBind(OnModelBoundListener<EpoxyArticles_, TrackItemRowView> listener);

  EpoxyArticlesBuilder onUnbind(OnModelUnboundListener<EpoxyArticles_, TrackItemRowView> listener);

  EpoxyArticlesBuilder onVisibilityStateChanged(
      OnModelVisibilityStateChangedListener<EpoxyArticles_, TrackItemRowView> listener);

  EpoxyArticlesBuilder onVisibilityChanged(
      OnModelVisibilityChangedListener<EpoxyArticles_, TrackItemRowView> listener);

  EpoxyArticlesBuilder consignmentWrapper(
      @Nullable NonEqualWrapper<Consignment> consignmentWrapper);

  EpoxyArticlesBuilder itemViewListener(@Nullable ActionableTrackItemListener itemViewListener);

  EpoxyArticlesBuilder layout(@LayoutRes int layoutRes);

  EpoxyArticlesBuilder id(long id);

  EpoxyArticlesBuilder id(@androidx.annotation.Nullable Number... ids);

  EpoxyArticlesBuilder id(long id1, long id2);

  EpoxyArticlesBuilder id(@androidx.annotation.Nullable CharSequence key);

  EpoxyArticlesBuilder id(@androidx.annotation.Nullable CharSequence key,
      @androidx.annotation.Nullable CharSequence... otherKeys);

  EpoxyArticlesBuilder id(@androidx.annotation.Nullable CharSequence key, long id);

  EpoxyArticlesBuilder layout(@LayoutRes int layoutRes);

  EpoxyArticlesBuilder spanSizeOverride(
      @androidx.annotation.Nullable EpoxyModel.SpanSizeOverrideCallback spanSizeCallback);
}

markchristopherng:
Hi just a little bit more information the affect classes seem to extend EpoxyModelWithView

E.g.

@EpoxyModelClass
[...]

PR #991, opened 2020-06-01
Fix duplicate layout method

elihart:
Fixes https://github.com/airbnb/epoxy/issues/988

Also updates AGP to 4.0.

Got rid of checkstyle as something broke with it and the AGP upgrade. couldn't be bothered to figure out the fix since all our new code should be kotlin anyway


lizhuogui:
Same issuse when override layout without annotation or different params name
```
    override fun layout(layoutRes: Int): EpoxyModel<TextView> {
        return super.layout(layoutRes)
    }
```
```
    override fun layout(@LayoutRes layout: Int): EpoxyModel<TextView> {
        return super.layout(layout)
    }
```

elihart:
I've updated it to ignore param name and just look at type. annotation shouldn't affect it

Issue #979, opened 2020-05-24
EpoxyModelGroup DSL improvement


sanogueralorenzo:
Hello everyone 👋 , 

I came up with a working solution for EpoxyModelGroup DSL.

I checked https://github.com/airbnb/epoxy/issues/800 but the solution involved some workaround on building them first and then getting them. Not sure if there are shortcomings to this approach so it would be great to get some 👀 on it.

I have a simple EpoxyModelGroup, just a cardview with a linear layout like shown in the [grouping models example](https://github.com/airbnb/epoxy/wiki/Grouping-Models#usage) where I fill that layout up:

![image](https://user-images.githubusercontent.com/5630683/82761737-8ab34c80-9db1-11ea-9381-7392471fa0b4.png)

My Kotlin class looks like:
``` kotlin
class CardViewEpoxyModelGroup(vararg models: EpoxyModel<*>) :
    EpoxyModelGroup(R.layout.container_card_view, *models)
```

Usage in current version is something like this  (non dsl way)
``` kotlin
override fun controller() = simpleController(viewModel) { state ->
        CardViewEpoxyModelGroup(
            TextRowModel_()
                .id("text")
                .body("Hello")
                .style(TextRow.TextStyle.BODY)
        // Other models...
        ).addTo(this)
    }
```

I was looking into the generated `EpoxyModelKotlinExtensions.kt` and for all the models it automatically adds itself
``` kotlin
inline fun EpoxyController.textRow(modelInitializer: TextRowModelBuilder.() -> Unit) {
    TextRowModel_().apply { modelInitializer() }.addTo(this)
}
```

Just to try it out I created manually an equivalent, appending Model at the end and the only difference is that it returns the Model_ instead of adding it
``` kotlin
inline fun EpoxyController.textRowModel(modelInitializer: TextRowModelBuilder.() -> Unit): TextRowModel_ =
 TextRowModel_().apply { modelInitializer() }
```

So the new usage becomes something like this:
``` kotlin
    override fun controller() = simpleController(viewModel) { state ->
        CardViewEpoxyModelGroup(
            textRowModel {
                id("text")
                body("Hello")
                style(TextRow.TextStyle.BODY)
            }
        ).addTo(this)
    }
```

Are there any disadvantages to this (apart from having to write the model function manually)?
Would it be a good idea to autogenerate these methods as well?

I was testing it with different components, static and also ones where the model changed a lot (text input layout) and I didn't see any issues 

The last few questions are:
[...]

PR #1012, opened 2020-06-30
Model group building with Kotlin DSL

eboudrant:
Proposal for DSL usage in with `EpoxyModelGroup` (Issue #979), the change is based on the `ModelCollector.

It require modification in code generation. Any epoxy model class that implements `ModelCollector` will have an associated builder that implements `ModelCollector` too.

It also require a change in `EpoxyModelGroup as we need the model / layout to be lazily set (require an empty constructor). 

To do : 
- Add a unit test for code generation

Also:
- Should we make the builder implements `ModelCollector` only if the model is an `EpoxyModelGroup` ?
- Should `GroupModel.kt` added to `epoxy-adapter` ? It is currently in `kotlinsample` app.

I like how it is now as it give some flexibility (any model can be a collector). Maybe we could also improve carousel DSL so the helper extensions are not needed (which could happen in a subsequent pull request).

So here the usage in sample app : 
```
group {
    id("epoxyModelGroupDsl")
    layout(R.layout.vertical_linear_group)

    coloredSquareView {
        id("coloredSquareView 1")
        color(Color.DKGRAY)
    }

    coloredSquareView {
        id("coloredSquareView 2")
        color(Color.GRAY)
    }

    coloredSquareView {
        id("coloredSquareView 3")
        color(Color.LTGRAY)
    }
}
```

image


epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModelGroup.java
```diff
@@ -59,9 +59,9 @@
 @SuppressWarnings("rawtypes")
 public class EpoxyModelGroup extends EpoxyModelWithHolder<ModelGroupHolder> {
 
-  protected final List<? extends EpoxyModel<?>> models;
-  /** By default we save view state if any of the models need to save state. */
-  private final boolean shouldSaveViewState;
+  protected final List<EpoxyModel<?>> models;
+
+  private boolean shouldSaveViewState = false;
```
elihart, line 64: before the DSL, the recommended approach to customize this was to subclass the EpoxyModelGroup and override `shouldSaveViewState`. Since that isn't possible with a DSL what do you think of just making this not private?

We will want to make sure that the default handling of setting this based on added models does not override a custom setting, so we can do something like
```
private boolean defaultShouldSaveViewState = false;
public Boolean shouldSaveViewState = null;
````

and then the value we use is `shouldSaveViewState ?: defaultShouldSaveViewState` (i wish this file was in kotlin :P )

File: epoxy-processor/src/main/java/com/airbnb/epoxy/processor/ModelBuilderInterfaceWriter.kt [hunk 2]
```diff
@@ -74,6 +74,14 @@ class ModelBuilderInterfaceWriter(
             addTypeVariables(modelInfo.typeVariables)
             addMethods(interfaceMethods)
 
+            for (implements in modelInfo.superClassElement.interfaces) {
+                if (implements.toString() == ClassNames.MODEL_COLLECTOR.toString()) {
```
elihart, line 78: it would be nice to memoize `ClassNames.MODEL_COLLECTOR.toString()`, or better yet, memoize on the class level via `memoizer.implementsModelCollector(modelInfo.superClassElement)`

The `memoizer` is a shared instance of Memoizer in the processor module that can be passed into this class

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModelGroup.java
eboudrant: Ok, I missed it, how about this : 

```
group {
  id(...)
  layout(...)
  shouldSaveViewState(true)
  ...
}
```

It should be possible with :

```
  private boolean shouldSaveViewStateDefault = false;

  @Nullable
  public Boolean shouldSaveViewState = null;

  @NonNull
  public EpoxyModelGroup shouldSaveViewState(boolean shouldSaveViewState) {
    onMutation();
    this.shouldSaveViewState = shouldSaveViewState;
    return this;
  }

  @Override
  public boolean shouldSaveViewState() {
    // By default state is saved if any of the models have saved state enabled.
    // Override this if you need custom behavior.
    if (shouldSaveViewState != null) {
      return shouldSaveViewState;
    } else {
      return shouldSaveViewStateDefault;
    }
  }
```

ps: I was really close to migrate the class to Kotlin :)

epoxy-processor/src/main/java/com/airbnb/epoxy/processor/ModelBuilderInterfaceWriter.kt
eboudrant: Done. 

Are you worried about performance in general? Theses was simple `.toString()` and iteration on a likely small collections.

epoxy-processor/src/main/java/com/airbnb/epoxy/processor/Memoizer.kt
```diff
@@ -355,6 +355,16 @@ class Memoizer(
...
         }
     }
+
+    private val implementsModelCollectorMap = mutableMapOf<GeneratedModelInfo, Boolean>()
+    fun implementsModelCollector(modelInfo: GeneratedModelInfo): Boolean {
+        return synchronized(typeMap) {
+            val implementsModelCollector = modelInfo.superClassElement.interfaces.firstOrNull {
+                it.toString() == ClassNames.MODEL_COLLECTOR.toString()
+            } != null
+            implementsModelCollectorMap.getOrPut(modelInfo) { implementsModelCollector }
```
elihart, line 365: `implementsModelCollector` is always calculated, even when the map already contains the value.

Could you remove the `implementsModelCollector` intermediate property and just put the whole calculation in the `getOrPut` lambda?

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModelGroup.java
elihart: That sounds great

epoxy-processor/src/main/java/com/airbnb/epoxy/processor/ModelBuilderInterfaceWriter.kt
elihart: @eboudrant thanks! I am a bit worried about performance - I recently reworked a lot of code to memoize hot spots and it improved performance several times (we dropped processing in one module from 4 seconds to half a second).

The `toString` functions here are a bit deceiving, one is in the ClassName class, which calls `emit` to a code writer to format the code for display which isn't trivial. It is internally cached, which I forgot about, but that isn't synchronized, which we have to worry about now since I added parallel processing support. The interface `toString` function is on the javac class `ClassType` which potentially loads the class by reading it from file which is not thread safe, so the memoizer helps us to synchronize.

epoxy-processor/src/main/java/com/airbnb/epoxy/processor/Memoizer.kt
```diff
@@ -355,6 +355,16 @@ class Memoizer(
             }
         }
     }
+
+    private val implementsModelCollectorMap = mutableMapOf<GeneratedModelInfo, Boolean>()
+    fun implementsModelCollector(modelInfo: GeneratedModelInfo): Boolean {
+        return synchronized(typeMap) {
+            val implementsModelCollector = modelInfo.superClassElement.interfaces.firstOrNull {
```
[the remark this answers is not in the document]
eboudrant: I also need to remove `!= null` ... `.firstOrNull { ... } != null` -> `.any { ... }`

eboudrant:
> What do you think of adding `ModelCollector` to the original EpoxyModelGroup, and shipping a `modelGroup` builder function with the core library?

@elihart I have `EpoxyModelGroup` implements `ModelCollector` now and I also added this simple class in the core library so `group { ... }` can be used as DSL out of the box : 

```
@EpoxyModelClass
abstract class GroupModel : EpoxyModelGroup()
```

It works in sample app but what is scary is unit tests, theses one started to fails : 
```
> Task :epoxy-integrationtest:testDebugUnitTest

com.airbnb.epoxy.AutoModelIntegrationTest > implicitlyAddingAutoModels FAILED
    java.lang.AssertionError at AutoModelIntegrationTest.java:61

com.airbnb.epoxy.AutoModelIntegrationTest > implicitlyAddingAutoModels2 FAILED
    java.lang.AssertionError at AutoModelIntegrationTest.java:69

com.airbnb.epoxy.AutoModelIntegrationTest > implicitlyAddingAutoModels3 FAILED
    java.lang.AssertionError at AutoModelIntegrationTest.java:77
```
```
Expected :[Model_{value=1}Model_{id=-1, viewType=2131427371, shown=true, addedToAdapter=false}, Model_{value=0}Model_{id=1, viewType=2131427371, shown=true, addedToAdapter=false}, Model_{value=2}Model_{id=2, viewType=2131427371, shown=true, addedToAdapter=false}, Model_{value=2}Model_{id=-2, viewType=2131427371, shown=true, addedToAdapter=false}, Model_{value=3}Model_{id=3, viewType=2131427371, shown=true, addedToAdapter=false}, Model_{value=0}Model_{id=-4, viewType=2131427371, shown=true, addedToAdapter=false}]
Actual   :[Model_{value=0}Model_{id=1, viewType=2131427371, shown=true, addedToAdapter=false}, Model_{value=2}Model_{id=2, viewType=2131427371, shown=true, addedToAdapter=false}, Model_{value=3}Model_{id=3, viewType=2131427371, shown=true, addedToAdapter=false}, Model_{value=0}Model_{id=-4, viewType=2131427371, shown=true, addedToAdapter=false}]
```

Do you have any idea? Theses tests don't seems to use model group.

eboudrant:
> It works in sample app but what is scary is unit tests, theses ones started to fails :

It seems test failures are related to code gen and are un-detected in pull request check, is the build incremental on CI ?
Somehow `getConfigurationForElement(controller.controllerClassElement).implicitlyAddAutoModels` always return `false` and break auto-model tests by not adding the statement `"setControllerToStageTo(controller.\$L, controller)"`.

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModelGroup.java
```diff
@@ -57,11 +58,14 @@
...
 
-  protected final List<? extends EpoxyModel<?>> models;
-  /** By default we save view state if any of the models need to save state. */
-  private final boolean shouldSaveViewState;
+  protected final List<EpoxyModel<?>> models;
+
+  private boolean shouldSaveViewStateDefault = false;
+
+  @Nullable
+  public Boolean shouldSaveViewState = null;
```
elihart, line 68: should this not be public? do we want to force usage of `shouldSaveViewState` so the mutation check is always done?

epoxy-adapter/src/main/java/com/airbnb/epoxy/GroupModel.kt
```diff
@@ -0,0 +1,7 @@
+package com.airbnb.epoxy
+
+/**
+ * An [EpoxyModelGroup] usable in DSL.
```
elihart, line 4: could be nice to give a small example of usage (or we can link to the wiki docs here later)

epoxy-processor/src/main/java/com/airbnb/epoxy/processor/Memoizer.kt
```diff
@@ -359,6 +359,17 @@ class Memoizer(
             }
         }
     }
+
+    private val implementsModelCollectorMap = mutableMapOf<GeneratedModelInfo, Boolean>()
```
elihart, line 363: to actually get good cache hits on this the parameter of the function should be `classElement: TypeElement`
then the map can be `mutableMapOf<Name, Boolean>()`, with `getOrPut(classElement.qualifiedName)` as the key to make sure it is consistent across type instances

epoxy-processor/src/main/java/com/airbnb/epoxy/processor/Memoizer.kt
```diff
@@ -359,6 +359,17 @@ class Memoizer(
...
+    fun implementsModelCollector(modelInfo: GeneratedModelInfo): Boolean {
+        return synchronized(typeMap) {
+            implementsModelCollectorMap.getOrPut(modelInfo) {
+                modelInfo.superClassElement.interfaces.any {
```
elihart, line 367: Should this check the super class recursively? If there are multiple subclasses a superclass up the chain could have the interface instead.

maybe `interfaces.any {...}  || implementsModelCollector(superClassElement.superclass)` (psuedocode)

elihart:
@eboudrant I pulled your branch but can't reproduce the test failures. it passes for me with `./gradlew clean testDebug`

epoxy-adapter/src/main/java/com/airbnb/epoxy/GroupModel.kt
eboudrant: Ok, also wdyt to add the `ModelCollector` interface / `add` method in this small class instead of in `EpoxyModelGroup` class, like : 

```
@EpoxyModelClass
abstract class GroupModel : EpoxyModelGroup(), ModelCollector {

    override fun add(model: EpoxyModel<*>) {
        super.addModel(model)
    }
}
```

Because in any case `EpoxyModelGroup` can't be used directly in the DSL. It allow to separate responsibilities.

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModelGroup.java
eboudrant: Good catch, I can't remember why I changed it to `public`

eboudrant:
> @eboudrant I pulled your branch but can't reproduce the test failures. it passes for me with `./gradlew clean testDebug`

`./gradlew clean testDebug` pass for me too, then when I make a modification to `ControllerWithImplicitlyAddedModels3` and run `./gradlew testDebug`, it fails. Ex, here the modification I do : 

image

```
> Task :epoxy-integrationtest:testDebugUnitTest

com.airbnb.epoxy.AutoModelIntegrationTest > implicitlyAddingAutoModels FAILED
    java.lang.AssertionError at AutoModelIntegrationTest.java:61

com.airbnb.epoxy.AutoModelIntegrationTest > implicitlyAddingAutoModels2 FAILED
    java.lang.AssertionError at AutoModelIntegrationTest.java:69

com.airbnb.epoxy.AutoModelIntegrationTest > implicitlyAddingAutoModels3 FAILED
    java.lang.AssertionError at AutoModelIntegrationTest.java:77

100 tests completed, 3 failed

> Task :epoxy-integrationtest:testDebugUnitTest FAILED
```

Weird

elihart:
@eboudrant this is very likely a bug with incremental processing, probably because of the kapt issue where package-info.java files are not properly included in incremental kapt processing. This leads to the configuration info from `@PackageEpoxyConfig` to not be included, which then causes the failure you see.

This is unrelated to your change, so let's just ignore it for now. The workaround I will have to apply l later is similar to https://github.com/airbnb/epoxy/pull/1020

eboudrant:
@elihart how close are we from a 4.0.0 out of beta? Is there a list of things to do apart from the package info issue?

elihart:
@eboudrant I would say it is very close. We have been using it in production for a long time with no issues. The package-info issue is the only known problem, and that is only going to be a sporadic incremental issue that requires a clean (or sometimes deleting local build cache), and only if you happen to be using that config annotation.

My current plan is to fix the package-info problem, and then I can make a normal release

plnice:
@elihart is this available in any beta/snapshot build?

elihart:
@plnice not yet, it will be in the next beta, whenever I have time to do that

elihart:
this is in the 4.0.0 release

elihart:
I've updated the wiki to mention this dsl https://github.com/airbnb/epoxy/wiki/Grouping-Models#kotlin-dsl

PR #1026, opened 2020-08-05
Allow specifying layout in GroupModel constructor

plnice:
This is a small, but useful improvement to #1012. It allows us to generate and use predefined DSLs which do not require referencing layout each time the group DSL is used. This is especially useful when we don't want to access another module's R class.

As in example, by introducing abstract class that extends `GroupModel` and annotating it with `@EpoxyModelClass`:
```kotlin
@EpoxyModelClass
abstract class DecoratedLinearGroupModel : GroupModel(R.layout.decorated_linear_group)
```
we can use following DSL:
```kotlin
decoratedLinearGroup {
    id("epoxyModelGroupWithLayoutDsl")
    ...
}
```
instead of:
```kotlin
group {
    id("epoxyModelGroupWithLayoutDsl")
    layout(R.layout.decorated_linear_group)
    ...
}
```

### DIFF
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModelGroup.java
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModelGroup.java
@@ -110,9 +110,10 @@ public class EpoxyModelGroup extends EpoxyModelWithHolder<ModelGroupHolder> {
   /**
    * Constructor use for DSL
    */
-  protected EpoxyModelGroup() {
+  protected EpoxyModelGroup(@LayoutRes int layoutRes) {
     models = new ArrayList<>();
     shouldSaveViewStateDefault = false;
+    layout(layoutRes);
   }
 
   protected void addModel(@NonNull EpoxyModel<?> model) {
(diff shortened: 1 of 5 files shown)


epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModelGroup.java
```diff
@@ -110,9 +110,10 @@ private EpoxyModelGroup(@LayoutRes int layoutRes, List<EpoxyModel<?>> models) {
   /**
    * Constructor use for DSL
    */
-  protected EpoxyModelGroup() {
+  protected EpoxyModelGroup(@LayoutRes int layoutRes) {
```
elihart, line 113: isn't this breaking existing usages by removing the empty constructor? should this be a new, separate constructor from the empty one?

plnice: The empty constructor was introduced in #1012, so it was not released yet and couldn't be used by library users. However, I've restored it to not introduce confusion that you have to provide some layout resource even if it's not necessary at this point.

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModelGroup.java
```diff
@@ -110,9 +110,10 @@ private EpoxyModelGroup(@LayoutRes int layoutRes, List<EpoxyModel<?>> models) {
...
     models = new ArrayList<>();
     shouldSaveViewStateDefault = false;
+    layout(layoutRes);
```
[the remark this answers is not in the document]
plnice: I've checked and 0 passed as a layout res is ignored by `EpoxyModel`, but you're right, that's something that may change in the future. Fix in 76bbea7

PR #1030, opened 2020-08-11
fix #1003

### LINKED ISSUE #1003: Sticky header doesn't respect view margins
I'm using Epoxy 3.11.0, and trying out these new sticky headers. Everything seems to work fine, except that the margins I have on my sticky header item disappear as soon as it becomes "sticky".

Video showing the issue: https://drive.google.com/file/d/1qQLTnNYhTtakHPc_wRxiVbzMtzcLZv3L/view?usp=sharing

Bilue404:
hi , this  pr is layout the StickyHeader's margin .
left and right's margin on VERTICAL .
top and bottom's marign on HORIZONTAL.

### DIFF
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/stickyheader/StickyHeaderLinearLayoutManager.kt
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/stickyheader/StickyHeaderLinearLayoutManager.kt
@@ -326,9 +326,10 @@ class StickyHeaderLinearLayoutManager @JvmOverloads constructor(
      */
     private fun measureAndLayout(stickyHeader: View) {
         measureChildWithMargins(stickyHeader, 0, 0)
+        val lp = stickyHeader.layoutParams as RecyclerView.LayoutParams
         when (orientation) {
-            VERTICAL -> stickyHeader.layout(paddingLeft, 0, width - paddingRight, stickyHeader.measuredHeight)
-            else -> stickyHeader.layout(0, paddingTop, stickyHeader.measuredWidth, height - paddingBottom)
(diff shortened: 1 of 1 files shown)


elihart:
looks good to me on paper, but I have a hard time saying whether this is correct for the layout manager. @AkshayChordiya what do you think?

@Bilue404 thanks for putting this up. could you perhaps provide screenshots to demonstrate how the fix looks?

Bilue404:
Thanks for your review 

if my HeaderModel  has  a margin like this.



it will flash to (0,0) when it sticks header,  like this .



After this pr , it will still layout on (marginLeft,0)  when it  sticks header,  like this 




is has the same effect as StickyHeaderLinearLayoutManager.setStickyHeaderTranslationX() . 
but translationX works on whole LayoutManager.
margin can work on each model.

DennisMarcu:
@Bilue404 Nice! My team and I were recently looking at this issue and we were going to apply a similar change to fix our StickyHeaders. The only thing that I would suggest is that you apply all the margins for both vertical and horizontal. The reason for this is if you want to create a 'floating' sticky header your solution will still 'snap' up.

Bilue404:
thanks very much @DennisMarcu 

i saw `StickyHeaderLinearLayoutManager` had` setStickyHeaderTranslationX/Y` API .
maybe we can use **setStickyHeaderTranslationY on vertical** and **setStickyHeaderTranslationX on horizontal** 
to change the direction's start point if we need a 'floating' sticky header.

i think this has the least impact on those who are using it,
so i had't changed their start point on they direction

AkshayChordiya:
Thank you for the PR. I'm also not entirely sure about this. 
You could always customize the margin in the `setupStickyHeaderView` function provided by sticky header feature in the epoxy controller and adapter so I don't see the need to do it on the layout manager level 🤔 

@DennisMarcu You could use the same function to either use translation or elevation to provide the floating effect

Bilue404:
@AkshayChordiya get it ,  thanks ~

PR #1039, opened 2020-08-19
Add an epoxy-paging3 module with source-compatibility fixes

dlam:
Adding a new module so that it can be published in parallel, since it's expected that Paging3 will be in alpha for awhile.

This essentially bumps the lower bound of generics from <T : Any?> to <T : Any>, which was previously disallowed by Paging2 via runtime checks, but now that Paging3 is in Kotlin, we're able to put the nullity directly in the type system.

Paging3 deprecates `PagedList` in favor of `PagingData`, but I thought it might be to prudent to setup a smaller PR to setup this module first before looking to add a `PagingData` based `ModelCache` / `EpoxyController`


dlam:
Eventually, the goal would be to deprecate PagedList* for PagingData* in this module.

dlam:
Hmm I realized my Studio setup added license headers randomly.. will fix that

dlam:
Or rather.. looks like they were copied over from existing epoxy-paging, so maybe I should leave them?

epoxy-paging3/src/main/java/com/airbnb/epoxy/paging3/PagingEpoxyController.java
```diff
@@ -0,0 +1,279 @@
...
+ * PagedList.
+ * <p>
+ * {@link #buildModels(List)} will be called with the subset of items that should have models built
+ * for them.
+ *
+ * @param <T> The type of item in the list
+ *
+ * @deprecated Use {@link PagedListEpoxyController} instead.
+ */
+@Deprecated()
```
elihart, line 40: I think we can delete this file. It has been deprecated for a long time - I don't see a reason to bring it along to this new module, seems like a good time to finally delete it

epoxy-paging3/src/main/java/com/airbnb/epoxy/paging3/SimplePagingEpoxyController.java
```diff
@@ -0,0 +1,28 @@
...
+import java.util.List;
+
+/**
+ * Use this version of {@link PagingEpoxyController} if you don't have any other model types to
+ * build (eg header, footer, etc). This will build one model for each item in the list by calling
+ * {@link #buildModel(Object)}
+ *
+ * @param <T> The list type
+ *
+ * @deprecated Use {@link PagedListEpoxyController} instead.
```
elihart, line 16: also delete this one

epoxy-paging3/src/main/java/com/airbnb/epoxy/paging3/PagingEpoxyController.java
dlam: SGTM!

dlam:
Thanks for taking the time to review!

linhvnguyen9:
@dlam are there any plan to support paging3 anytime soon?

PR #1046, opened 2020-09-06
4.0.0 release

elihart:
## New
- Incremental annotation processing for faster builds
- A new annotation processor argument `logEpoxyTimings` can be set to get a detailed breakdown of how long the processors took and where they spent their time (off by default)
- Another new argument `enableParallelEpoxyProcessing` can be set to true to have the annotation processor process annotations and generate files in parallel (via coroutines).

You can enable these processor options in your build.gradle file like so:
```
project.android.buildTypes.all { buildType ->
  buildType.javaCompileOptions.annotationProcessorOptions.arguments =
      [
          logEpoxyTimings  : "true",
          enableParallelEpoxyProcessing     : "true"
      ]
}
```

Parallel processing can greatly speed up processing time (moreso than the incremental support), but given the hairy nature of parallel processing it is still incubating.
Please report any issues or crashes that you notice.
(We are currently using parallel mode in our large project at Airbnb with no problems.)

- Add options to skip generation of functions for getters, reset, and method overloads to reduce generated code
    - New annotation processor options are:
        - epoxyDisableGenerateOverloads
        - epoxyDisableGenerateGetters
        - epoxyDisableGenerateReset

## Fixes
- Synchronize ListUpdateCallback and PagedListModelCache functions (#987)
- Avoid generating bitset checks in models when not needed (reduces code size)

## Breaking

- Annotations that previously targeted package elements now target types (classes or interfaces).
  This includes: `EpoxyDataBindingPattern`, `EpoxyDataBindingLayouts`, `PackageModelViewConfig`, `PackageEpoxyConfig`
  This was necessary to work around an incremental annotation processor issue where annotation on package-info elements are not properly recompiled

- In order to enable incremental annotation processing a change had to be made in how the processor of
  `@AutoModel` annotations work. If you use `@AutoModel` in an EpoxyController the annotated Model types
  must be either declared in a different module from the EpoxyController, or in the same module in the same java package.

  Also make sure you have kapt error types enabled.

  However, generally `@AutoModel` is considered legacy and is not recommended. It is a relic of Java Epoxy usage
  and instead the current best practice is to use Kotlin with the Kotlin model extension functions to build models.

- Removed support for generating Epoxy models from Litho components


eboudrant:
Thanks for releasing it, my 2 comments: 
- Is Kotlin 1.4 a prerequisite for Epoxy 4.0.0 ? Looks like it update Kotlin stdlib to 1.4.
- Should the release note mention the model group dsl ? (I should also update the doc on the wiki)

elihart:
@eboudrant thanks for the reminder, sorry I missed that. I have updated the release notes to include it https://github.com/airbnb/epoxy/releases/tag/4.0.0 and also yes we should update the wiki. I think only I have write access, but I can do it now.

About kotlin 1.4.0, I believe that with both 1.4.x and 1.3.x standard libraries on the classpath gradle will emit a warning, but it would still compile. If it is a problem you should be able to have gradle force a resolution to 1.3.x (I think)

Honestly I didn't think about it too much, and hopefully its not a problem for people. I'm not sure what the recommended approach to updating to kotlin 1.4 in libraries is, while making it safe for consumers - if you have any suggestions please let me know. Also if 1.4 is a problem I should also be able to make a 4.0.0 compat release that uses kotlin 1.3.72 instead (we still can't update to 1.4.0 ourselves...)

eboudrant:
We still have some perf issue with Kotlin 1.4 so we are waiting for the next version to update. To be extra safe ex excluded Kotlin from Epoxy dependency (`exclude group: "org.jetbrains.kotlin"`). Seems no problem so far.

elihart:
great, thanks for the details. glad its working like that

PR #1051, opened 2020-09-11
Update installation template with proper version

aruke:
Thanks for the 4.0.0 release!! 🎉 

Since the release of version 4.0.0, the template `3.x.y` doesn't make sense, so replaced it with a common variable.


elihart (approved):
great, thanks for catching this :)

Issue #1050, opened 2020-09-11
Tests in epoxy-integrationtest are flaky


pmecho:
While working on a PR I had a failure running `./gradlew check`. I thought this might be related to my bumping of `androidx.test:core` to `1.3` but I also get the same flakiness on `1.2`. One example is that they were successful 3 times in a row and the 4th time is when they failed.

One stacktrace is:
```
java.lang.NullPointerException
	at com.airbnb.epoxy.integrationtest.KotlinViewWithDefaultParamsModel_.bind(KotlinViewWithDefaultParamsModel_.java:104)
	at com.airbnb.epoxy.integrationtest.KotlinViewWithDefaultParamsModel_.bind(KotlinViewWithDefaultParamsModel_.java:33)
	at com.airbnb.epoxy.EpoxyViewHolder.bind(EpoxyViewHolder.java:43)
	at com.airbnb.epoxy.BaseEpoxyAdapter.onBindViewHolder(BaseEpoxyAdapter.java:118)
	at com.airbnb.epoxy.BaseEpoxyAdapter.onBindViewHolder(BaseEpoxyAdapter.java:106)
	at com.airbnb.epoxy.ControllerLifecycleHelper.bindModels(ControllerLifecycleHelper.java:30)
	at com.airbnb.epoxy.ControllerLifecycleHelper.bindModels(ControllerLifecycleHelper.java:23)
	at com.airbnb.epoxy.ControllerLifecycleHelper.buildModelsAndBind(ControllerLifecycleHelper.java:19)
	at com.airbnb.epoxy.KotlinDefaultParamTest$TestController.buildWithModel(KotlinDefaultParamTest.kt:36)
	at com.airbnb.epoxy.KotlinDefaultParamTest.defaultParamIsUsed(KotlinDefaultParamTest.kt:53)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
[...]

PR #1052, opened 2020-09-11
Add view visibility checks to EpoxyVisibilityItem and decouple RecyclerView

pmecho:
## Problems

1. If the view transitions to a visibility of GONE, it will retain a non-zero width and height. This causes the methods inside `EpoxyVisibilityItem` to continue to return true even though the view is not visible on the screen. 
2. The `EpoxyVisibilityItem` is strongly coupled with a `RecyclerView`. We have code in our app that uses epoxy outside of a `RecyclerView` and wish to incorporate the visibility tracking with that.

## Solutions
1. Store the visibility value of the view during the `update` call and check it when the `handle...` methods are called.
2. We can just change the `RecyclerView` argument to use a `ViewGroup`. This will support the existing use case as well as allowing it to be used outside of `RecyclerView`s. A default constructor was added out of convenience but is by no means needed. The existing constructor really just sets the adapter position which isn't used if not using a `RecyclerView`


blessedDeps.gradle
```diff
@@ -68,7 +68,8 @@ rootProject.ext.deps = [
     androidArchCoreTesting : "android.arch.core:core-testing:$ANDROID_ARCH_TESTING",
     androidTestRunner      : "com.android.support.test:runner:$ANDROID_TEST_RUNNER",
     androidAnnotations     : "androidx.annotation:annotation:$ANDROIDX_ANNOTATION",
-    androidTestCore        : "androidx.test:core:1.2.0",
+    androidTestCore        : "androidx.test:core:1.3.0",
```
pmecho, line 71: I needed to update this to run the activity scenario tests. Otherwise the tests would crash.

epoxy-annotations/src/main/java/com/airbnb/epoxy/OnVisibilityStateChanged.java
```diff
@@ -10,10 +10,10 @@
  * with this annotation will be called when the visibility state is changed.
  * <p>
  * Annotated methods should follow this signature :
- * `@OnVisibilityStateChange
```
pmecho, line 13: These references were incorrect so I updated them

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityItem.java
```diff
@@ -139,11 +145,15 @@ boolean handleChanged(EpoxyViewHolder epoxyHolder, boolean visibilityChangedEnab
     boolean changed = false;
     if (visibleHeight != lastVisibleHeightNotified || visibleWidth != lastVisibleWidthNotified) {
       if (visibilityChangedEnabled) {
-        epoxyHolder.visibilityChanged(
-            100.f / height * visibleHeight,
-            100.f / width * visibleWidth,
-            visibleHeight, visibleWidth
-        );
+        if (viewVisibility == View.GONE) {
+          epoxyHolder.visibilityChanged(0f, 0f, 0, 0);
```
elihart, line 149: do we need to check that `viewVisibility` changed? could it have already been gone but had the size change?

elihart (approved):
@pmecho looks great to me, just had one question.

Thanks for all the clean tests too!

eboudrant:
Nice addition! We are also starting to have models outside of `RecyclerView`. 

What I am missing where how do you use it? Don't you need to update/refactor the `EpoxyVisibilityTraker` which can only be attached to a `RecyclerView`.

pmecho:
@eboudrant Eli was spot on. We're using an internal class for the visibility tracking when used outside of a `RecyclerView`. That's still sort of a work in progress but I totally agree that we should upstream it, along with how we use epoxy outside of `RecylcerView`s. Once I finish up with the tracking I can make another PR if that works.

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityItem.java
pmecho: Nice catch! I'd agree it makes sense to also check that. I'll make that change.

eboudrant:
Well now I can't wait these other PRs 👍

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityItem.java
pmecho: @elihart While testing this I realized the size won't change while it's gone since the layout pass won't re-measure. I did find a case that makes us need this last visibility notified check though. If the view transitions from gone to visible the size won't change so we would need that check. I've added it and this should be good to go now.

pmecho:
Oof looks like CI is failing due to the flaky integration tests noted here: https://github.com/airbnb/epoxy/issues/1050

elihart:
@pmecho if you rebase the tests should be fixed! then we can merge and get a release out

eboudrant:
> @eboudrant Eli was spot on. We're using an internal class for the visibility tracking when used outside of a `RecyclerView`. That's still sort of a work in progress but I totally agree that we should upstream it, along with how we use epoxy outside of `RecylcerView`s. Once I finish up with the tracking I can make another PR if that works.

Hi @pmecho ... when do you think it is a good time work on a new change to upstream the `EpoxyViewBinder` (outside of RV) and the visibility  tracking? We use `EpoxyViewBinder` without visibility tracking but we may need it soon and I don't want to re-invent the wheel :)

pmecho:
@eboudrant heh it keeps getting bumped down my priority list. This may be the nudge I needed though. I will work on it this next week and maybe have something by end of week or the beginning of the next. When are you thinking you'll need it?

eboudrant:
@pmecho the timing you propose works for me. I don't want to change your priorities list. I would say we would starts to need visibility in about a month. Thanks!

Issue #954, opened 2020-04-30
Can't call notifyItemChanged(position) to trigger revert swipe


semanticer:
As mentioned here https://stackoverflow.com/questions/31787272/android-recyclerview-itemtouchhelper-revert-swipe-and-restore-view-holder

probably the only way to trigger revert swipe animation is to call `notifyItemChanged` on your adapter directly. The problem is that this is not possible with epoxy since you get `java.lang.IllegalStateException: You cannot notify item changes directly. Call requestModelBuild instead.`

The issue is that `requestModelBuild` doesn't trigger redraw since the backing model didn't really change. 

Here is the link to the exact place where I need to use notifyItemChanged but can't https://github.com/toggl/android/blob/5e0c1882ab3ebcba4c7f83486168e300653cbd59/timer/src/main/java/com/toggl/timer/log/ui/TimeEntriesLogFragment.kt#L138

Do you have some workaround for this?
Thanks

semanticer:
The only way I could find so far is to reattach touch helper: 
```kotlin
itemTouchHelper.attachToRecyclerView(null);
itemTouchHelper.attachToRecyclerView(recyclerView);
```
mentioned in https://stackoverflow.com/a/37342327/509681

This will immediately force the list item to its original position but without the revert animation that we need. Is there really no way to do this properly? 

Thanks for your answer and keep up the good work 👍

elihart:
We can't expose that adapter API in EpoxyController, but I think what you can do is have a boolean prop on your model, like `isSwiped` and when you want to revert it you can set it to false when you rebuild models. This will cause Epoxy to notify an item change on that item.

of course, that won't happen until the next frame, and I'm not sure if that's a problem for your use case. if so the only solution I can see is an Epoxy internal change to allow this to happen synchronously, possible via the touch helper API. i suppose allowing access to notify item changes is pretty safe (as opposed to notifying insertions or removals)

semanticer:
Thanks for the answer. I've experimented with this `isSwiped` approach and it still doesn't solve the problem, unfortunately 🤔. I isolated this issue in a [separate controller with debugging turned on](https://github.com/toggl/android/blob/epoxy-log-is-swiped/timer/src/main/java/com/toggl/timer/log/ui/TimeEntryController.kt).  
[...]

PR #1063, opened 2020-09-30
Add notify model changed method for issue 954

berendjan:
Allows notifying that a model has changed on a position, purpose is to allow for revert swiping a model.
Issue https://github.com/airbnb/epoxy/issues/954


elihart:
Looks good to me, thanks for contributing!

PR #1090, opened 2020-11-19
Convert EpoxyVisibilityTracker code to Kotlin

eboudrant:
This change only convert `EpoxyVisibilityTracker` and `EpoxyVisibilityItem` to kotlin. 

It may be easier to review by commit : 
- IDE convert : 1f63da3
- Cleanup generated code from `!!` : 90de635
- Code formatting : c0b84ac
- Fix a warning : f9f561d
- Fix a unit test running from another module : 4d22fd1

I'd like this code to be in Kotlin for a future change I am about to propose.
Thanks


elihart:
Great :) I'll try to review and merge this soon

pmecho:
@eboudrant awesome! I was just thinking about converting this last night. I'm working on adding `EpoxyModelGroup` support to this so I'll base my work off of this branch.

eboudrant:
> @eboudrant awesome! I was just thinking about converting this last night. I'm working on adding `EpoxyModelGroup` support to this so I'll base my work off of this branch.

Nice! This is also my motivation: add support to group that have a nested carousel. 
I will try to show you my proposal too. Still early but works for us.

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.kt
```diff
@@ -0,0 +1,506 @@
...
+
+    private fun processChildRecyclerViewAttached(childRecyclerView: RecyclerView) {
+        // Register itself in the EpoxyVisibilityTracker. This will take care of nested list
+        // tracking (ex: carousel)
+        var tracker = getTracker(childRecyclerView)
+        if (tracker == null) {
+            tracker = EpoxyVisibilityTracker()
+            tracker.setPartialImpressionThresholdPercentage(partialImpressionThresholdPercentage)
+            tracker.attach(childRecyclerView)
+        }
```
pmecho, line 304: We could shorten this a bit by doing:
```suggestion
        val tracker = getTracker(childRecyclerView) ?: EpoxyVisibilityTracker().apply {
            setPartialImpressionThresholdPercentage(partialImpressionThresholdPercentage)
            attach(childRecyclerView)
        }
```

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.kt
```diff
@@ -0,0 +1,506 @@
...
+        /**
+         * For all items after the inserted range shift each [EpoxyVisibilityTracker] adapter
+         * position by inserted item count.
+         */
+        override fun onItemRangeInserted(positionStart: Int, itemCount: Int) {
+            if (notEpoxyManaged(attachedRecyclerView)) {
+                return
+            }
+            if (DEBUG_LOG) {
+                Log.d(TAG, String.format("onItemRangeInserted(%d, %d)", positionStart, itemCount))
```
pmecho, line 389:
```suggestion
                Log.d(TAG, "onItemRangeInserted($positionStart, $itemCount)")
```

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.kt
```diff
@@ -0,0 +1,506 @@
...
+        /**
+         * For all items after the removed range reverse-shift each [EpoxyVisibilityTracker]
+         * adapter position by removed item count
+         */
+        override fun onItemRangeRemoved(positionStart: Int, itemCount: Int) {
+            if (notEpoxyManaged(attachedRecyclerView)) {
+                return
+            }
+            if (DEBUG_LOG) {
+                Log.d(TAG, String.format("onItemRangeRemoved(%d, %d)", positionStart, itemCount))
```
pmecho, line 408:
```suggestion
                Log.d(TAG, "onItemRangeRemoved($positionStart, $itemCount)")
```

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.kt
```diff
@@ -0,0 +1,506 @@
...
+
+        private fun onItemMoved(fromPosition: Int, toPosition: Int) {
+            if (notEpoxyManaged(attachedRecyclerView)) {
+                return
+            }
+            if (DEBUG_LOG) {
+                Log.d(
+                    TAG,
+                    String.format("onItemRangeMoved(%d, %d, %d)", fromPosition, toPosition, 1)
+                )
```
pmecho, line 442:
```suggestion
                Log.d(TAG, "onItemRangeMoved($fromPosition, $toPosition, 1)"
```

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityItem.kt
```diff
@@ -0,0 +1,208 @@
...
+
+    /** Store last value for de-duping  */
+    private var lastVisibleHeightNotified = NOT_NOTIFIED
+    private var lastVisibleWidthNotified = NOT_NOTIFIED
+    private var lastVisibilityNotified = NOT_NOTIFIED
+
+    constructor() {}
+    constructor(adapterPosition: Int) {
+        reset(adapterPosition)
+    }
```
pmecho, line 59: We could also move the empty constructor to the class definition but no strong preference either way here. That'd looks something like 
```
class EpoxyVisibilityItem() {
    ...
    constructor(adapterPosition: Int) : this() {
        reset(adapterPosition)
    }
```

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityItem.kt
```diff
@@ -0,0 +1,208 @@
...
+        return changed
+    }
+
+    private fun isVisible(): Boolean {
+        return viewVisibility == View.VISIBLE && visibleHeight > 0 && visibleWidth > 0
+    }
+
+    // The model has entered the focused range either if it is larger than half of the viewport
+    // and it occupies at least half of the viewport or if it is smaller than half of the viewport
+    // and it is fully visible.
```
pmecho, line 170: Extraneous comment 
```suggestion
```

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityItem.kt
```diff
@@ -0,0 +1,208 @@
...
+        get() {
+            val halfViewportArea = viewportHeight * viewportWidth / 2
+            val totalArea = height * width
+            val visibleArea = visibleHeight * visibleWidth
+            // The model has entered the focused range either if it is larger than half of the viewport
+            // and it occupies at least half of the viewport or if it is smaller than half of the viewport
+            // and it is fully visible.
+            return viewVisibility == View.VISIBLE &&
+                if (totalArea >= halfViewportArea) visibleArea >= halfViewportArea else totalArea == visibleArea
+        }
```
pmecho, line 181: Should we just make this a function?
```suggestion
    private fun isInFocusVisible(): Boolean {
        val halfViewportArea = viewportHeight * viewportWidth / 2
        val totalArea = height * width
        val visibleArea = visibleHeight * visibleWidth
        // The model has entered the focused range either if it is larger than half of the viewport
        // and it occupies at least half of the viewport or if it is smaller than half of the viewport
        // and it is fully visible.
        return viewVisibility == View.VISIBLE &&
            if (totalArea >= halfViewportArea) visibleArea >= halfViewportArea else totalArea == visibleArea
    }
```

pmecho (changes requested):
@eboudrant This looks great to me with one exception. We need to cast to a nullable in `EpoxyVisibilityTracker.getTracker` otherwise we'll crash. Added some optional suggestions to clean up/fix some of the things missed by the auto converter as well.

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.kt
```diff
@@ -0,0 +1,506 @@
...
+
+    /**
+     * Don't call this method directly, it is called from
+     * [EpoxyVisibilityTracker.processVisibilityEvents]
+     *
+     * @param child               the view to process for visibility event
+     * @param detachEvent         true if the child was just detached
+     * @param eventOriginForDebug a debug strings used for logs
+     */
+    private fun processChild(child: View, detachEvent: Boolean, eventOriginForDebug: String) {
```
pmecho, line 212: @eboudrant wanted to put this in a separate review so we could have threaded comments. My potentially naive solution was to replace this method with the following. Basically it checks if the child is a model group and processes each one, taking care to also add nested trackers for recycler views. Curious what your thoughts are here or if you have a more robust solution.

```
    private fun processChild(child: View, detachEvent: Boolean, eventOriginForDebug: String) {
        val recyclerView = attachedRecyclerView ?: return
        // Preemptive check for child's parent validity to prevent `IllegalArgumentException` in
        // `getChildViewHolder`.
        val isParentValid = child.parent == null || child.parent === recyclerView
        val viewHolder = if (isParentValid) recyclerView.getChildViewHolder(child) else null
        if (viewHolder is EpoxyViewHolder) {
            val epoxyHolder = viewHolder.holder
            if (epoxyHolder is ModelGroupHolder) {
                processModelGroupChildren(epoxyHolder, detachEvent, eventOriginForDebug)
            } else {
                processChild(child, detachEvent, eventOriginForDebug, viewHolder)
            }
        }
    }

    /**
     * Loop through the children of the model group and process visibility events on each one in
     * relation to the model group's layout. This will attach or detach trackers to any nested
     * [RecyclerView]s.
     *
     * @param epoxyHolder         the [ModelGroupHolder] with children to process
     * @param detachEvent         true if the child was just detached
     * @param eventOriginForDebug a debug strings used for logs
     */
    private fun processModelGroupChildren(
        epoxyHolder: ModelGroupHolder,
        detachEvent: Boolean,
        eventOriginForDebug: String
    ) {
        // Iterate through models in the group and process each of them instead of the group
        for (groupChildHolder in epoxyHolder.viewHolders) {
            // Since the group is likely using a ViewGroup other than a RecyclerView we need to
            // handle the potential of a nested RecyclerView.
            if (groupChildHolder.itemView is RecyclerView) {
                if (detachEvent) {
                    processChildRecyclerViewDetached(groupChildHolder.itemView)
                } else {
                    processChildRecyclerViewAttached(groupChildHolder.itemView)
                }
            }
            processChild(
                groupChildHolder.itemView,
                detachEvent,
                eventOriginForDebug,
                groupChildHolder
            )
        }
    }

    /**
     * Process visibility events for a view and propagate to a nested tracker if the view is a 
     * [RecyclerView].
     */
    private fun processChild(
        child: View,
        detachEvent: Boolean,
        eventOriginForDebug: String,
        viewHolder: EpoxyViewHolder
    ) {
        val recyclerView = attachedRecyclerView ?: return
        val changed = processVisibilityEvents(
            recyclerView,
            viewHolder,
            detachEvent,
            eventOriginForDebug
        )
        if (changed && child is RecyclerView) {
            val tracker = nestedTrackers[child]
            tracker?.processChangeEvent("parent")
        }
    }
```

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.kt
```diff
@@ -0,0 +1,506 @@
...
+
+        @IdRes
+        private val TAG_ID = R.id.epoxy_visibility_tracker
+
+        /**
+         * @param recyclerView the view.
+         * @return the tracker for the given [RecyclerView]. Null if no tracker was attached.
+         */
+        private fun getTracker(recyclerView: RecyclerView): EpoxyVisibilityTracker? {
+            return recyclerView.getTag(TAG_ID) as EpoxyVisibilityTracker
```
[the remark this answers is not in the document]
eboudrant: Yes, missed it because the nested visibility test is disabled. I will re-enable it. And I don't remember why I disabled it :(

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityItem.kt
eboudrant: Oh, that was generated by the IDE

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.kt
```diff
@@ -0,0 +1,506 @@
...
+                    // Is some case the detached child is still in the recycler view. Don't process it as it
+                    // was already processed.
+                    processChild(child, false, debug)
+                }
+            }
+        }
+    }
+
+    /**
+     * If there is a new adapter on the attached RecyclerView it will resister the data observer and
```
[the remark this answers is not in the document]
eboudrant: resist! :)

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.kt
```diff
@@ -0,0 +1,506 @@
...
+import android.view.View
+import androidx.annotation.IdRes
+import androidx.annotation.IntRange
+import androidx.recyclerview.widget.RecyclerView
+import com.airbnb.viewmodeladapter.R
+import java.util.ArrayList
+import java.util.HashMap
+
+/**
+ * A simple way to track visibility events on [com.airbnb.epoxy.EpoxyModel] within a [ ].
```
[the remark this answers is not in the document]
eboudrant: Not sure what I wanted to link here, I will remove the `within a`.

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityItem.kt
eboudrant: > Should we just make this a function?

Interesting, Kotlin conversion converted a java method to `val`, good catch.

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.kt
eboudrant: Agree, let's convert first to Kotlin and create new review for everything related to group / visibility. but looking at your proposal, I like it.

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.kt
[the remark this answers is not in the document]
eboudrant: I will fix the crash in this PR and open a new PR for re-enabling the unit test.

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityItem.kt
pmecho: Heh, yeah I figured :-) Auto conversion can do some odd stuff sometimes.

pmecho (approved):
LGTM!

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityItem.kt
```diff
@@ -0,0 +1,207 @@
...
+
+    fun handleChanged(epoxyHolder: EpoxyViewHolder, visibilityChangedEnabled: Boolean): Boolean {
+        var changed = false
+        if (visibleHeight != lastVisibleHeightNotified || visibleWidth != lastVisibleWidthNotified || viewVisibility != lastVisibilityNotified) {
+            if (visibilityChangedEnabled) {
+                if (viewVisibility == View.GONE) {
+                    epoxyHolder.visibilityChanged(0f, 0f, 0, 0)
+                } else {
+                    epoxyHolder.visibilityChanged(
+                        100f / height * visibleHeight,
```
elihart, line 153: named parameters might make this a little more readable

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityItem.kt
```diff
@@ -0,0 +1,207 @@
...
+    @Px
+    private var viewportWidth = 0
+    private var partiallyVisible = false
+    private var fullyVisible = false
+    private var visible = false
+    private var focusedVisible = false
+    private var viewVisibility = View.GONE
+
+    /** Store last value for de-duping  */
+    private var lastVisibleHeightNotified = NOT_NOTIFIED
```
elihart, line 54: could null be used instead of -1?

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.kt
```diff
@@ -0,0 +1,498 @@
...
+    /**
+     * Enable or disable visibility changed event. Default is `true`, disable it if you don't need
+     * (triggered by every pixel scrolled).
+     *
+     * @see OnVisibilityChanged
+     *
+     * @see OnModelVisibilityChangedListener
+     */
+    fun setOnChangedEnabled(enabled: Boolean) {
+        onChangedEnabled = enabled
```
elihart, line 75: is this for backwards compatibility? kind of a shame to have this function when we could just make the var public - wouldn't the java interop be the same with just the var?

elihart: same for `setPartialImpressionThresholdPercentage` I guess

elihart (approved):
@eboudrant changes look great, just had some minor kotlin style suggestions.

Thanks for this!

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityItem.kt
eboudrant: Pb is `EpoxyViewHolder` is still in Java, or do you meant to also covert it? I prefer to focus this change on visibility code.

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.kt
eboudrant: Same ... it is IDE convert :)

eboudrant: @elihart okay for a breaking change? `setPartialImpressionThresholdPercentage` won't be available from Kotlin, what is the general policy for this?

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.kt
eboudrant: @pmecho I tried your code and it seems to work. One change I needed is to remove the `else` in `processChild` : 
```
            processChild(child, detachEvent, eventOriginForDebug, viewHolder)
            if (epoxyHolder is ModelGroupHolder) {
                processModelGroupChildren(epoxyHolder, detachEvent, eventOriginForDebug)
            }
```

How about you open a new pull request with your change and we iterate from there? I'd like to do more testing.

pmecho: @eboudrant Oh nice catch. Probably makes sense to still call the visibility methods on the model group itself. I made that change and pushed a draft PR: https://github.com/airbnb/epoxy/pull/1091

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityItem.kt
elihart: ah, my mistake. 🙏

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyVisibilityTracker.kt
elihart: @eboudrant I am in favor of a small breaking change like that (easy for people to fix) in order to set us up with a better API going forward :) thanks!

elihart:
Looks great. Appreciate all the tests you had in place already too, makes this much safer :)

Issue #1094, opened 2020-11-24
The ModelGroupHolder should use the pool from its parent instead of creating it


eboudrant:
When you configure your `EpoxyRecyclerView` to not use shared activity pool with : 
```
override fun shouldShareViewPoolAcrossContext() = false
```

If you add a group in it, the group will still try to get a shared pool. The reason is `ModelGroupHolder` never check the parent but call directly : 
```
poolReference = ACTIVITY_RECYCLER_POOL.getPool(itemView.context) { UnboundedViewPool() }
```

PR #1095, opened 2020-11-24
ModelGroupHolder to get the pool from parent RecyclerView

eboudrant:
Fix for #1094

This change remove `PoolReference` from `ModelGroupHolder`. Instead it just need a `RecyclerView.RecycledViewPool`. 

This new pool is passed in the constructor by the `EpoxyModelGroup` as it is the only place we have access to the parent `RecyclerView`.

`EpoxyModelGroup` will try to find a pool from the closest `RecyclerView` parent. If the group is not in a `RecyclerView` then it will use a local pool.

### DIFF
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModelGroup.java
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModelGroup.java
@@ -9,7 +9,6 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
-import java.util.Objects;
 
 import androidx.annotation.CallSuper;
 import androidx.annotation.LayoutRes;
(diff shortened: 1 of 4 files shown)


eboudrant:
@elihart let me know what you think (or if there is a better way of doing it). I had to update the `EpoxyModelGroup` which I was not expecting at first, kind of required to grab the parent's view pool.

elihart (approved):
@eboudrant this looks reasonable given the constraints. I'm a _little_ wary of putting the view pool in the model group as an intermediary (since technically the model could be reused in a separate recyclerview), but see why you had to do it that way.

The only other approach I can think of involves modifying `EpoxyViewHolder` to take a reference to the parent in `BaseEpoxyAdapter#onCreateViewHolder` and then passing that parent on to `EpoxyModelWithHolder#createNewHolder` when the holder is first created in `bind` (and then nulling out the parent to be safe after bind?)

This is a larger change, but is potentially more explicit and cleaner. I don't feel too strongly one way or another though so I'll leave it up to you

eboudrant:
> since technically the model could be reused in a separate recyclerview
Can you elaborate a bit on this? I though a model (`EpoxyModelGroup`) is built for one controller/recyclerview (via `buildModels`) and never re-used.

elihart:
>  I though a model (EpoxyModelGroup) is built for one controller/recyclerview (via buildModels) and never re-used. 

That's not a requirement anywhere, and technically it could be reused for performance optimizations. I highly doubt anyone is doing that though, as the kotlin dsl pushes away from that approach.

The paging epoxy controller is one example that does cache and reuse models, but it doesn't generally reuse them across different recyclerviews (unless you happened to manually switch the controller to a different recyclerview)

eboudrant:
Closing as a cleaner solution was implemented here #1097.

PR #1097, opened 2020-11-27
ModelGroupHolder to get the pool from parent - another solution

eboudrant:
Fix for #1094

This change remove PoolReference from ModelGroupHolder. Instead it just need a RecyclerView.RecycledViewPool.
`ModelGroupHolder` will find the pool from the `ViewParent` passed in constructor. 

- `EpoxyModelWithHolder` updated to add the `parent` in `createNewHolder`.
- `ModelGroupHolder` will use the `parent` to retrieve a view pool.
- `EpoxyHolder` have 2 constructors, one empty, one with the parent (to prevent painful breaking change). The code generation will invoke the constructor with parent if it exists.
- As an example, I updated the `ImageModel.ImageHolder` in sample app. If we have the parent we don't need `by lazy` to initialize glide (03e0e4a).
- With unit test

The change have changed in generated code. I plan to update the processor unit tests after the code review to prevent noise in diff.

There is one more change I want to address but it can go in a separate pull request which is related to `Carousel`. I think a carousel should always get the view pool from parent. Currently it get it from the logic we have in `EpoxyRecyclerView`. If the share pool across context is false we would have really inefficient pooling on nested lists (unfortunately Hilt force us to set it to false).

@elihart this is the solution you proposed #1095 which I think it cleaner too.

### DIFF
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModelWithHolder.java
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModelWithHolder.java
@@ -1,5 +1,7 @@
 package com.airbnb.epoxy;
 
+import android.view.ViewParent;
+
 import com.airbnb.epoxy.VisibilityState.Visibility;
 
 import java.util.List;
@@ -22,7 +24,7 @@ public abstract class EpoxyModelWithHolder<T extends EpoxyHolder> extends EpoxyM
   }
 
   /** This should return a new instance of your {@link com.airbnb.epoxy.EpoxyHolder} class. */
-  protected abstract T createNewHolder();
+  protected abstract T createNewHolder(@NonNull ViewParent parent);
 
   @Override
   public void bind(@NonNull T holder) {
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyViewHolder.java
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyViewHolder.java
@@ -20,6 +20,8 @@ public class EpoxyViewHolder extends RecyclerView.ViewHolder {
   private List<Object> payloads;
   private EpoxyHolder epoxyHolder;
   @Nullable ViewHolderState.ViewState initialViewState;
+
+  // Once the EpoxyHolder is created parent will be set to null.
   private ViewParent parent;
 
   public EpoxyViewHolder(ViewParent parent, View view, boolean saveInitialState) {
@@ -48,6 +50,7 @@ public class EpoxyViewHolder extends RecyclerView.ViewHolder {
     if (epoxyHolder == null && model instanceof EpoxyModelWithHolder) {
       epoxyHolder = ((EpoxyModelWithHolder) model).createNewHolder(parent);
       epoxyHolder.bindView(itemView);
+      parent = null;
     }
 
     if (model instanceof GeneratedModel) {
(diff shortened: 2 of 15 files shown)


epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyViewHolder.java
```diff
@@ -43,7 +46,7 @@ public void bind(@SuppressWarnings("rawtypes") EpoxyModel model,
     this.payloads = payloads;
 
     if (epoxyHolder == null && model instanceof EpoxyModelWithHolder) {
-      epoxyHolder = ((EpoxyModelWithHolder) model).createNewHolder();
+      epoxyHolder = ((EpoxyModelWithHolder) model).createNewHolder(parent);
```
elihart, line 49: what do you think about nulling out `parent` after this `if` statement so we don't leak a reference? If the recycled view pool is used across multiple recyclerviews (like with the activity wide pool) then there could be multiple parents, and if one is destroyed i wouldn't want it to leak here.

It's a little funky, but I think it would be safe to null it out

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModelWithHolder.java
```diff
@@ -22,7 +24,7 @@ public EpoxyModelWithHolder(long id) {
   }
 
   /** This should return a new instance of your {@link com.airbnb.epoxy.EpoxyHolder} class. */
-  protected abstract T createNewHolder();
+  protected abstract T createNewHolder(@NonNull ViewParent parent);
```
elihart, line 27: I suppose this would be a breaking change for a few people and they would have to update. I'm not opposed to that though

elihart (approved):
@eboudrant thanks for this. it does seem better, and helpful for other cases though, but sorry it ended up being so much more complicated. The changes seem clean though! Lgtm once the tests are updated. Agree that the carousel change followup would be a good idea too

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModelWithHolder.java
eboudrant: Yes, you're right, I am not opposed too.

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyViewHolder.java
eboudrant: Ok, because it is little funky I will add a comment.

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyViewHolder.java
```diff
@@ -43,8 +48,9 @@ public void bind(@SuppressWarnings("rawtypes") EpoxyModel model,
...
       epoxyHolder.bindView(itemView);
+      parent = null;
```
elihart, line 53: @eboudrant I think this needs to go outside of the braces so that it is nulled out for models that don't have holders as well?

eboudrant: Fixed, good catch!

elihart: Thanks for the quick fix :) I'll merge this, then if you want to get the carousel fix in I can push a release afterwards

eboudrant: I am not sure when I will have the carousel fix so having a release seen would be nice (visibility on group and this fix) :)

eboudrant:
@elihart I tried the update, if you use a lot of `EpoxyModelWithHolder` the update can be a little painful as you need to update every  `createNewHolder` methods. Better to document this in the release note.

PR #1099, opened 2020-12-02
[proposal] linking the view pool to activities to lifecycle owner

eboudrant:
This is a proposal and will require an update to Hilt, see https://github.com/google/dagger/issues/2202#issuecomment-735984477.

The idea of theses changes in to Epoxy and Hilt are to fix a memory leak. It happen because Epoxy link the view pool to the activity. However, when a view get inflated in a Hilt fragment, the view's context is a wrapper and hold a reference on the fragment. The fragment leak for the whole activity lifecycle.

So why Epoxy link the view pool arbitrary to the `Activity`? Instead, could we use the view's context and find the closest `LifecycleOwner`. When this lifecycle get destroyed we can clear the pool.

It is a proposal, happy to discuss here. Using the `LifecycleOwner` instead of `Activity` looks natural. Is it possible to use Epoxy outside of an `Activity` that is not a `LifecycleOwner`?

### DIFF
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyRecyclerView.kt
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyRecyclerView.kt
@@ -248,7 +248,7 @@ open class EpoxyRecyclerView @JvmOverloads constructor(
         }
 
         setRecycledViewPool(
-            ACTIVITY_RECYCLER_POOL.getPool(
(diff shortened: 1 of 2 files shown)


elihart (commented):
I think this makes a lot of sense, and is cleaner anyway. I can't think of any cases it wouldn't work in, although I also don't know when the context wouldn't be an Activity (possible a Compose context? but epoxy wouldn't really be used there anyway, unless we have some interop)

PR #1101, opened 2020-12-02
Fix ANR and view pool resolution in nested group

eboudrant:
Well, I completely forgot there was an infinite loop with a nested group with Epoxy 4.3.0. First, there is a bug in the `findViewPool` on the exit condition. If no parent found it would run on an infinite loop. Happen consistently with a nested group.

So the second fix is to make sure the nested group finds the right pool. I updated the `HelperAdapter` so it knows about the `ModelGroupHolder`'s parent. (the `parent` passed in the `HelperAdapter.createViewHolder` is not yet attached so `findViewPool` would not work as intended).

I added a nested group in the unit test coverage.

### DIFF
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/ModelGroupHolder.kt
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/ModelGroupHolder.kt
@@ -9,7 +9,7 @@ import androidx.recyclerview.widget.RecyclerView
 import com.airbnb.viewmodeladapter.R
 import java.util.ArrayList
 
-class ModelGroupHolder(parent: ViewParent) : EpoxyHolder() {
+class ModelGroupHolder(private val modelGroupParent: ViewParent) : EpoxyHolder() {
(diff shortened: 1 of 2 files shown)


epoxy-adapter/src/main/java/com/airbnb/epoxy/ModelGroupHolder.kt
```diff
@@ -9,12 +9,12 @@ import androidx.recyclerview.widget.RecyclerView
 import com.airbnb.viewmodeladapter.R
 import java.util.ArrayList
 
-class ModelGroupHolder(parent: ViewParent) : EpoxyHolder() {
+class ModelGroupHolder(private val modelGroupParent: ViewParent) : EpoxyHolder() {
```
elihart, line 12: I'd like to follow up later to see if we can avoid storing this parent, as theoretically this holder can get recycled to be used by another recyclerview

elihart (approved):
thanks for the fix and tests, will make a 4.3.1 release with this

PR #1114, opened 2020-12-21
[ModelView] - Add code to support set layout params manually from init

dinhlamvn:
Currently, we just create `LayoutParams` of model by type of size:
```kotlin
enum Size {
    NONE,
    WRAP_WIDTH_WRAP_HEIGHT,
    WRAP_WIDTH_MATCH_HEIGHT,
    MATCH_WIDTH_WRAP_HEIGHT,
    MATCH_WIDTH_MATCH_HEIGHT
  }
```
The layout params will be automatic created base on that.

But sometime, we need to create layout params of view manually when we init a view. In this PR, I've added some code to do that. 

Please help me check and hope it will helpful.


kotlinsample/src/main/java/com/airbnb/epoxy/kotlinsample/models/ItemTitleTextView.kt
```diff
@@ -0,0 +1,30 @@
...
+        ).apply {
+            setMargins(50, 50, 50, 50)
+        }
+    }
+
+    @TextProp
+    fun setTitle(title: CharSequence?) {
+        text = title
+    }
+}
```
duchuyctlk, line 30: @dinhlamvn please add a new empty line here, it may help your PR pass the CI check :smile:

elihart:
I believe the current way to manually set layout properties would be via an xml file that the view is inflated from. Can you justify why this should be added on top of that? Having two ways to do something is generally more confusing and should be avoided unless there is good reason.

dinhlamvn:
This case happend when I try inflate my view from xml with `merge` tag, that will merge child layout to parent, then epoxy `override` my layout params when build view if I use `autoLayout`.

leffsu:
@elihart, Greetings!

Could you give this PR more attention?

My use case is that I need to set a height of the element depending on the data from the server. When using ModelView.Size.MATCH_WIDTH_WRAP_HEIGHT, I see that the Recycler is shaking because the height of the elements is changing.

epoxy-annotations/src/main/java/com/airbnb/epoxy/ModelView.java
```diff
@@ -24,6 +24,7 @@
    */
   enum Size {
     NONE,
+    MANUAL,
```
elihart, line 27: This needs to have javadoc describing its behavior and requirements

epoxy-processor/src/main/java/com/airbnb/epoxy/processor/GeneratedModelWriter.kt
```diff
@@ -484,12 +493,13 @@ class GeneratedModelWriter(
         return methods
     }
 
-    private fun getLayoutDimensions(modelInfo: GeneratedModelInfo): Pair<CodeBlock, CodeBlock> {
+    private fun getLayoutDimensions(modelInfo: GeneratedModelInfo): Pair<CodeBlock?, CodeBlock?> {
```
elihart, line 496: use a nullable Pair instead of making the contents nullable
```suggestion
    private fun getLayoutDimensions(modelInfo: GeneratedModelInfo): Pair<CodeBlock, CodeBlock>? {
```

kotlinsample/src/main/java/com/airbnb/epoxy/kotlinsample/models/ItemTitleTextView.kt
```diff
@@ -0,0 +1,30 @@
...
+
+import android.content.Context
+import android.util.AttributeSet
+import android.view.ViewGroup
+import androidx.appcompat.widget.AppCompatTextView
+import com.airbnb.epoxy.ModelView
+import com.airbnb.epoxy.TextProp
+
+@ModelView(autoLayout = ModelView.Size.MANUAL)
+class ItemTitleTextView @JvmOverloads constructor(
```
elihart, line 11: Name this for what it is meant to be an example for 
```suggestion
class ManualLayoutParamsView @JvmOverloads constructor(
```

elihart (commented):
I can merge this once the comments are addressed.

Also, please add tests that assert the correct generated code is produced (ie add a test to ViewProcessorTest class)

dinhlamvn:
@elihart I have updated the PR, please help me verify that. Thanks you !!

epoxy-processor/src/main/java/com/airbnb/epoxy/processor/GeneratedModelWriter.kt
```diff
@@ -470,12 +470,19 @@ class GeneratedModelWriter(
...
-                )
+                getLayoutDimensions(modelInfo)?.let { (layoutWidth, layoutHeight) ->
+                    addStatement(
+                        "v.setLayoutParams(new \$T(\$L, \$L))",
+                        ClassNames.ANDROID_MARGIN_LAYOUT_PARAMS, layoutWidth, layoutHeight
+                    )
+                } ?: run {
+                    beginControlFlow("if (v.getLayoutParams() == null)")
+                        .addStatement(
+                            "throw new \$T(\"Layout params is required set for Size.MANUAL\")",
```
elihart, line 481:
```suggestion
                            "throw new \$T(\"Layout params is required to be set for Size.MANUAL\")",
```

epoxy-annotations/src/main/java/com/airbnb/epoxy/ModelView.java
```diff
@@ -20,10 +20,12 @@
   /**
    * Use with {@link #autoLayout()} to declare what layout parameters should be used to size your
    * view when it is added to a RecyclerView. This maps to the LayoutParams options {@code
-   * layout_width} and {@code layout_height}.
+   * layout_width} and {@code layout_height}. If you want to set the LayoutParams manually, you can use
+   * {@link Size#MANUAL} and set it in the ModelView's constructor.
```
elihart, line 24:
```suggestion
   * {@link Size#MANUAL} and set the params in the View's constructor when it is initialized (a runtime exception will be thrown if layout params are not set during view instantiation when MANUAL is used).
```

elihart (approved):
Thanks for the tests and example. Just need to polish the documentation

elihart:
also looks like the testAutoLayoutManualLayoutParams test is failing

dinhlamvn:
@elihart I have updated the documentation, please help me check it. Thanks you!

elihart:
@dinhlamvn thanks, the documentation looks good. only issue is the test is failing because of slightly different output. it expects java 11, and you may be on an older version which results in different behavior. (should be https://github.com/airbnb/epoxy/commit/e44ef43af771452c58248a4a4767600ede625849)

I can merge this and then merge the fix, since I don't think I can push to your branch directly

PR #1117, opened 2021-01-02
Update KotlinPoet to 1.7.2

plnice:
I recently ran into some strange issues with conflicting KotlinPoet versions between Epoxy and Moshi; updating KotlinPoet version in Epoxy fixes the problem. Additionally, in KotlinPoet 1.6.0, the issue with not escaping class name that contains `$` has been fixed.


elihart (approved):
Looks great, thanks!

PR #1144, opened 2021-03-03
Provide support for snapshot() function in PagingDataEpoxyController

anhanh11001:
The seems to be the only function in `paging3` lib I can find to provide access to all loaded data. I think it's useful.

A use case can be that as a developer, I want to track all of the IDs of the loaded posts in a screen, then this one can be used.

### DIFF
--- a/epoxy-paging3/src/main/java/com/airbnb/epoxy/paging3/PagedDataModelCache.kt
+++ b/epoxy-paging3/src/main/java/com/airbnb/epoxy/paging3/PagedDataModelCache.kt
@@ -4,6 +4,7 @@ import android.os.Handler
 import android.os.Looper
 import androidx.paging.AsyncPagingDataDiffer
 import androidx.paging.CombinedLoadStates
+import androidx.paging.ItemSnapshotList
 import androidx.paging.PagingData
 import androidx.recyclerview.widget.DiffUtil
 import androidx.recyclerview.widget.ListUpdateCallback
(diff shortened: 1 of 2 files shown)


anhanh11001:
@elihart You mind checking this out. It's a small change.

I'm also wondering if you think it's necessary to:
1. Should there be a paging3 epoxy sample
2. paging3 provides this function for the adapter https://developer.android.com/reference/kotlin/androidx/paging/PagingDataAdapter#withloadstateheaderandfooter which is convenient to add a loading indicator to the top/bottom of the list. Should there be something similar to the `epoxy-paging3`

If you find them useful and have any suggestions, I can start looking into them as I have some free time to contribute.

elihart (approved):
@anhanh11001 this change looks good to me, sorry for the delay in reviewing.

I think a paging3 sample could be helpful to people, so feel free to put up a PR if you are interested :) It could go in the same module as the paging2 sample https://github.com/airbnb/epoxy/tree/master/epoxy-pagingsample

As far as headers and footers, I think it is fairly easy for people to add those already by overriding the `addModels` function. While we could add functions explicitly to override to provide header or footer models to save them a few lines I don't think the extra complexity is worth the small savings and would rather leave the current API to maximize flexibility.

anhanh11001:
I see, thanks for replying.

> As far as headers and footers, I think it is fairly easy for people to add those already by overriding the addModels function. While we could add functions explicitly to override to provide header or footer models to save them a few lines I don't think the extra complexity is worth the small savings and would rather leave the current API to maximize flexibility.

Make sense to me

anhanh11001:
Oh can you also add this to the next release

elihart:
> Should I remove the paging2 sample because I don't think we can have 2 paging deps in the same module.?

ah, right. maybe it is best to create a new paging3 sample module for now then, since it doesn't have a stable release yet.

I'll try to make a new release with this change soon 👍

marksr12 (commented):
Pull requests

PR #1150, opened 2021-03-17
Fix for generated interfaces regression from #1142

vinaygaba:
This is a follow up to my previous PR(https://github.com/airbnb/epoxy/commit/fa61cfbc4059f88a3e89644ab21b3f8d70aa7c57) that tried to update the package name of generated model files for interface class. 

The appropriate interface wasn't being correctly referenced in the generated code that implemented this interface classes and so this PR adds that functionality. 

Thanks to @elihart for pair programming on this and helping accelerate this fix!


epoxy-processor/src/main/java/com/airbnb/epoxy/processor/ModelBuilderInterfaceWriter.kt
```diff
@@ -50,7 +50,12 @@ class ModelBuilderInterfaceWriter(
             val interfaceMethods = getInterfaceMethods(modelInfo, methods, interfaceName)
 
             if (modelInfo is ModelViewInfo) {
-                modelInfo.generatedViewInterfaceNames.forEach { viewInterface ->
```
elihart, line 53: can we delete the property `generatedViewInterfaceNames` now?

elihart (approved):

epoxy-processor/src/main/java/com/airbnb/epoxy/processor/ModelBuilderInterfaceWriter.kt
vinaygaba: makes sense!

PR #1162, opened 2021-04-03
Explicitly add public modifier

### LINKED ISSUE #1133: Support Explicit API Mode for Kotlin
Compiling with `strict` enabled causes a compile error. See: https://kotlinlang.org/docs/whatsnew14.html#explicit-api-mode-for-library-authors

Example from Epoxy output on an example project:

```kotlin
inline fun ModelCollector.highlightCard(modelInitializer: HighlightCardModelBuilder.() -> Unit) {
  add(
  HighlightCardModel_().apply {
    modelInitializer()
  }
  )
}
```

This would have to have visibility modifiers explicitly stated in the declaration.

osipxd:
Fixes #1133


elihart (approved):
Thanks!

PR #1165, opened 2021-04-12
EpoxyViewBinder update - add epoxyView / optionalEpoxyView to ViewGroup

eboudrant:
We have use case where the activity or the fragment is unaware of what model we want to bind, in this case we want to bind the model from any view (`ViewGroup` actually).

This PR also add `optionalEpoxyView` to activity.

### DIFF
--- a/epoxy-viewbinder/src/main/java/com/airbnb/epoxy/EpoxyViewBinder.kt
+++ b/epoxy-viewbinder/src/main/java/com/airbnb/epoxy/EpoxyViewBinder.kt
@@ -5,7 +5,7 @@ import android.util.Log
 import android.view.View
 import android.view.ViewGroup
 import androidx.annotation.IdRes
-import androidx.appcompat.app.AppCompatActivity
+import androidx.core.app.ComponentActivity
 import androidx.core.view.ViewCompat
 import androidx.core.view.children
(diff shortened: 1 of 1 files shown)


pmecho (approved):

eboudrant:
@pmecho I added the `optionalEpoxyView` to activity and view in the same pull request, let me know you that's okay. I made a minor code refactor/extraction so we can have less code dup.

epoxy-viewbinder/src/main/java/com/airbnb/epoxy/EpoxyViewBinder.kt
```diff
@@ -265,24 +288,77 @@ fun AppCompatActivity.epoxyView(
 fun Fragment.optionalEpoxyView(
     @IdRes viewId: Int,
     modelProvider: EpoxyController.() -> Unit,
-    initializer: (LifecycleAwareEpoxyViewBinder.() -> Unit)? = null,
+    initializer: (LifecycleAwareEpoxyViewBinder.() -> Unit) = { },
```
pmecho, line 291: Oh yeah that's way better 😄

pmecho (approved):
@eboudrant nice call on extracting the logic to `epoxyViewInternal` methods!

PR #1166, opened 2021-04-12
Fix generated code consistency in builder interfaces

elihart:
The generated files of the builder interfaces could be different across builds because of methods being generated in different orders. This can lead to different cache keys and cache misses.


vinaygaba (approved):
Great addition 👍

PR #1168, opened 2021-04-14
[View Binder] Add visibility tracker tests and API tweaks

pmecho:
Adding unit tests for the visibility tracker along with some API tweaks that will allow for the model provider to still be a trailing lambda (if desired) or an unnamed parameter.

### DIFF
--- a/epoxy-integrationtest/build.gradle
+++ b/epoxy-integrationtest/build.gradle
@@ -42,6 +42,7 @@ android {
 
 dependencies {
   implementation rootProject.deps.kotlin
+  implementation rootProject.deps.androidAppcompat
   implementation rootProject.deps.androidRecyclerView
   implementation rootProject.deps.androidPagingComponent
   implementation project(':epoxy-adapter')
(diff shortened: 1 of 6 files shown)


epoxy-integrationtest/build.gradle
```diff
@@ -48,9 +48,11 @@ dependencies {
   implementation project(':epoxy-annotations')
   implementation project(':epoxy-databinding')
   implementation project(':epoxy-paging')
+  implementation project(':epoxy-viewbinder')
 
   kapt project(':epoxy-processor')
 
+  testImplementation rootProject.deps.androidAppcompat
```
elihart, line 55: could just do
```suggestion
  implementation rootProject.deps.androidAppcompat
```

elihart (approved):
Thanks a lot for the tests!

epoxy-integrationtest/build.gradle
pmecho: Ah yeah that makes more logical sense, even if it's not used outside tests. I'll change it up but make the commit manually so it's grouped with the other `implementation` declarations

PR #1172, opened 2021-04-20
Set packageid to com.airbnb.epoxy.viewbinder

eboudrant:
@pmecho, minor update for the package id in the new module


pmecho (approved):
Whoops, thanks for the fix!

PR #1180, opened 2021-05-08
Use kotlin dsl marker for model building receivers

elihart:
I had gotten annoyed by autocomplete suggesting all of our model names inside of building another model - it makes it hard to find the applicable properties of a model as well as enables buggy behavior.

Adding the kotlin dsl marker annotation to both the ModelCollector as well as generated models prevents both of these things.

This is a breaking change if inside of a model building function you reference a property in enclosing EpoxyController - to do so now requires an explicit `this`. While being safer and more readable that might require some people to make quite a few updates for this.


martymiller:
@elihart My team recently updated to this version, which required over 500 line changes of heavily verbose code.  Seeing that this breaking change was needed because "something annoyed you" seems like a good reason to start migrating away from Epoxy.

elihart:
@martymiller did you see the option to disable this change in https://github.com/airbnb/epoxy/pull/1185?

This wasn't just because it annoyed me, it was trying to fix a problem that I expected would cause friction for many other people. If you see something you would like changed please submit a PR to improve the project; after all, it's free, open source software you're using.

Regarding migrating away, please do move to Compose as it is simply better.

martymiller:
Yes, but I couldn't get this code to work.  Are you sure you should be calling all() here?  That requires a Boolean to be returned.  

I tried this in my `build.gradle.kts` and that didn't work either.  

```
  project.android.buildTypes.forEach { buildType ->
    buildType.javaCompileOptions.annotationProcessorOptions {
      argument("epoxyDisableDslMarker", "true")
    }
  }
```

PR #1182, opened 2021-05-11
AndroidX dependency updates and View Binder extension extractions

pmecho:
@eboudrant suggested pulling out the extension functions for the `EpoxyViewBinder` to clean up the code. This extracts those out as well as increases the versions on several AndroidX dependencies to the latest stable version.


elihart (approved):
Thanks!

PR #1186, opened 2021-05-13
not for merge

eboudrant:
Example to illustrate problem with `@DslMarker1 and this pattern : 
```
class Controller(val object : ...) ... {
  fun buildModels(...) {
   group {
      addSomething(...)
    }
  }
  open fun ModelCollector.addSomething(...) {
    model {
      onClick { object.doSomething(...) }
    }
  }
}
```

### DIFF
--- a/kotlinsample/src/main/java/com/airbnb/epoxy/kotlinsample/MainController.kt
+++ b/kotlinsample/src/main/java/com/airbnb/epoxy/kotlinsample/MainController.kt
@@ -1,0 +1,150 @@
+package com.airbnb.epoxy.kotlinsample
+
+import android.content.Intent
+import android.graphics.Color
+import android.util.Log
+import android.widget.Toast
+import com.airbnb.epoxy.EpoxyController
+import com.airbnb.epoxy.EpoxyModel
+import com.airbnb.epoxy.ModelCollector
+import com.airbnb.epoxy.group
+import com.airbnb.epoxy.kotlinsample.helpers.carouselNoSnapBuilder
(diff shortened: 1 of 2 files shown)


kotlinsample/src/main/java/com/airbnb/epoxy/kotlinsample/MainController.kt
```diff
@@ -0,0 +1,150 @@
...
+                }
+            }
+        }
+
+        decoratedLinearGroup {
+            id("epoxyModelGroupWithLayoutDsl")
+
+            this@MainController.addColoredSquareView(Color.DKGRAY)
+            this@MainController.addColoredSquareView(Color.GRAY)
+            this@MainController.addColoredSquareView(Color.LTGRAY)
```
eboudrant, line 68: The problem is here : 

```e: kotlinsample/src/main/java/com/airbnb/epoxy/kotlinsample/MainController.kt: (66, 33): 'fun ModelCollector.addColoredSquareView(color: Int): Unit' can't be called in this context by implicit receiver. Use the explicit one if necessary```

Comment (eboudrant) [reply kotlinsample/src/main/java/com/airbnb/epoxy/kotlinsample/MainController.kt:68]: It will work with 
```
        decoratedLinearGroup {
            val group = this
            id("epoxyModelGroupWithLayoutDsl")
            with (this@MainController) {
                group.addColoredSquareView(Color.DKGRAY)
                group.addColoredSquareView(Color.GRAY)
                group.addColoredSquareView(Color.LTGRAY)
            }
        }
```
but this is not ideal.

PR #1189, opened 2021-05-23
[Readme] Add ViewBinding model creation examples and usage

bigman212:
Why not? 
ViewBinding is kinda popular nowadays, I think this is going to be useful. 

I am not sure if I should add a wiki page about ViewBinding and Epoxy, but updating Readme is a good starting point!

All the code is taken from your samples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,6 +162,43 @@ A `HeaderModel_` class is generated that subclasses HeaderModel and implements t
 
 [More Details](https://github.com/airbnb/epoxy/wiki/ViewHolder-Models)
 
+#### From ViewBinding
+
+If you use ViewBinding in your project, you have two ways to create a model that can use your \*Binding class:
+1. Using annotation processing (like examples above)
+```kotlin
+@EpoxyModelClass(layout = R.layout.view_binding_holder_item)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -162,6 +162,43 @@ A `HeaderModel_` class is generated that subclasses HeaderModel and implements t
...
+    val title: String
+) : ViewBindingKotlinModel<DataClassViewBindingItemBinding>(R.layout.data_class_view_binding_item) {
+
+    override fun DataClassViewBindingItemBinding.bind() {
+        title.text = this@ItemViewBindingDataClass.title
+    }
+    
+}
+```
+where [ViewBindingKotlinModel](https://github.com/airbnb/epoxy/blob/master/kotlinsample/src/main/java/com/airbnb/epoxy/kotlinsample/helpers/ViewBindingKotlinModel.kt) - base class special for this case
```
bigman212, line 199: In ViewBindingKotlinModel there is `R.id.epoxy_viewbinding` that can be unknown for those who is going to copy-paste this base class.
Maybe I should add some info about it? 
Like "for correct usage of this base class, add <item name="epoxy_viewbinding" type="id" /> in res/xml/ids.xml in your project"

bigman212:
@elihart

PR #1192, opened 2021-05-27
Adding ComposeInterop example app with Mavericks

meggamind:
### Summary
- Adding 100 headerView (Epoxy ModelView) and Cmopose Text (wrapped around ComposeInterop)
- Mavericks state contains counter list representing counterValue to display for each of the 100 views 
- Clicking both epoxy HeaderView and composeInterop expoxy view increases only the counter value of that view by updating the state counter list, which calls `bind` of the view and updates the counter value of only that view.

### Known Bug
- Compose function inside composeinterop does not update on state change. (Will fix this in next PR)


epoxy-composeinterop-maverickssample/src/main/java/com/airbnb/epoxy/composeinterop/maverickssample/epoxyviews/HeaderView.kt
```diff
@@ -0,0 +1,63 @@
...
+import com.airbnb.paris.annotations.Style
+import com.airbnb.paris.annotations.Styleable
+import com.airbnb.paris.extensions.headerViewStyle
+import com.airbnb.paris.extensions.layoutHeight
+import com.airbnb.paris.extensions.layoutWidth
+
+@Styleable // Dynamic styling via the Paris library
+@ModelView(saveViewState = true)
+class HeaderView(context: Context?) : LinearLayout(context) {
+    private var title: TextView? = null
```
elihart, line 23: should be lateinit so they're not nullable

epoxy-composeinterop-maverickssample/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,5 @@
+<resources>
+    <string name="app_name">MavericksExample</string>
+    <!-- TODO: Remove or change this placeholder text -->
+    <string name="hello_blank_fragment">Hello blank fragment</string>
```
elihart, line 4: delete

elihart (approved):
@meggamind thanks, lgtm with just small cleanups

PR #1201, opened 2021-06-20
Adding key, implementing equals & hashcode for composableinterop

meggamind:
- Adding `key` as constructor params to `ComposeEpoxyModel`
- This `key` will be provided by the api user via ModelCollector extension `composableInterop`.
- The `key` contributes to the `equals` and `hashcode` functions of the EpoxyModel and hence a change in `key` causes the EpoxyModel to rebind, which in turn sets the content as the `composeFunction` with new values
- Added mavericks sample android tests in travis CI
- Running all android test throws error, which is out of the scope of this PR to resolve
```
Task :epoxy-modelfactorytest:mergeDexDebugAndroidTest FAILED
```

### Added androidTests to Verify recomposition when using composableInterop
**UsingMutableStateUpdatesEvenWithoutKey**
- **Uses**: `composableInterop`  with a Compose view without a key
- **Tests**: If a mutable state variable changes, it causes a recomposition in the above compose view

**ComposableInteropWithIntegerKey, ComposableInteropWithStringKey, ComposableInteropWithListKey, ComposableInteropWithDataClassKey**
- **Uses**: `composableInterop` with a Compose view with and without a key
- **Tests**: If a key of the values passed to composable is not provided, recomposition doesn't occur


epoxy-compose/src/main/java/com/airbnb/epoxy/ComposeInterop.kt
```diff
@@ -11,27 +11,60 @@ import androidx.compose.ui.viewinterop.AndroidView
...
+            code = 31 * code + it.hashCode()
+        }
+
+        return code
+    }
+
+    /**
+     * This key will contribute to calculate hashcode & equals
+     * */
+    fun keys(vararg keys: Any) {
```
elihart, line 48: seems like this should go in the constructor to enforce it is set. (it is lateinit right now which will crash if not set)

epoxy-compose/src/main/java/com/airbnb/epoxy/ComposeInterop.kt
```diff
@@ -11,27 +11,60 @@ import androidx.compose.ui.viewinterop.AndroidView
...
+        this.keys = arrayOf(*keys)
+    }
 }
 
 fun ModelCollector.composableInterop(
     id: String,
+    vararg keys: Any,
```
elihart, line 55: `keys` needs to be well documented with the purpose and usage guide

epoxy-composeinterop-maverickssample/build.gradle
```diff
@@ -0,0 +1,65 @@
+plugins {
+  id 'com.android.application'
+  id 'kotlin-android'
+}
```
elihart, line 4: don't include the mvrx changes in this PR to make it reviewable. either change the base branch or remove all of these files

epoxy-compose/src/main/java/com/airbnb/epoxy/ComposeInterop.kt
```diff
@@ -11,27 +11,60 @@ import androidx.compose.ui.viewinterop.AndroidView
...
         view.setContent(composeFunction)
     }
+
+    override fun equals(other: Any?): Boolean {
+        if (other === this) return true
+        if (other !is ComposeEpoxyModel) return false
+        if (this.hashCode() != other.hashCode()) return false
+
+        for (i in keys.indices) {
+            if (keys[i] != other.keys[i]) return false
```
elihart, line 29: this will crash if the keys array is different sizes. use a built in array equality/hashcode function

epoxy-compose/src/main/java/com/airbnb/epoxy/ComposeInterop.kt
```diff
@@ -11,27 +11,60 @@ import androidx.compose.ui.viewinterop.AndroidView
...
 
     override fun bind(view: ComposeView) {
         super.bind(view)
         view.setContent(composeFunction)
     }
+
+    override fun equals(other: Any?): Boolean {
+        if (other === this) return true
+        if (other !is ComposeEpoxyModel) return false
+        if (this.hashCode() != other.hashCode()) return false
```
elihart, line 26: I don't think you should be checking hashcode in equals

elihart (commented):
@meggamind this is a good start, but needs to be fleshed out.

Besides documentation it would be good to add unit tests to ensure that if keys change recomposition happens, and if keys don't change there isn't recomposition.

Can you also add a PR summary to describe the purpose of this and the design decisions made?

epoxy-composeinterop-maverickssample/src/main/java/com/airbnb/epoxy/composeinterop/maverickssample/MultiKeyComposeInteropFragment.kt
```diff
@@ -0,0 +1,325 @@
...
+            val updatedDataClass =
+                state.keyAsDataClass.copy(listInDataClass = state.keyAsDataClass.listInDataClass.toMutableList())
+
+            updatedDataClass.listInDataClass.add(2)
+
+            setState { copy(keyAsDataClass = updatedDataClass) }
+        }
+    }
+
+    val mutableCounter = mutableStateOf(0)
```
vinaygaba, line 75: optional: this could very well be a flow or a live data and could be converted into a Compose compatible state object using the `collectAs/observeAs` extension methods. This allows for clean separation between compose and the rest of our architecture.

vinaygaba, line 75: this should be a var

epoxy-composeinterop-maverickssample/src/androidTest/java/com/airbnb/epoxy/composeinterop/maverickssample/MultiKeyComposeInteropFragmentTest.kt
```diff
@@ -0,0 +1,89 @@
...
+
+@RunWith(AndroidJUnit4::class)
+class MultiKeyComposeInteropFragmentTest {
+    @get:Rule
+    val composeTestRule = createAndroidComposeRule<MainActivity>()
+
+    @Test
+    fun testComposableInInteropUsingMutableStateUpdatesEvenWithoutKey() {
+        for (i in 0..10) {
+            waitUntil("withoutInteropKey mutableState: $i, recomposingCount: ${i + 1}")
```
vinaygaba, line 20: The current logic that's using waitUntil waits for a given text to show up but it doesn't guarantee anything since there are no assertions being made. It can recompose 50 times after this for loop and we aren't ensuring that those recompositions aren't happening. Unfortunately, its not verifying what these tests were hoping to verify.

vinaygaba: Maybe I misinterpreted the logic. Feel free to correct me if you feel otherwise!

vinaygaba (commented):
Took a first stab at reviewing the PR. I think the tests aren't yet verifying recompositions. Additionally, it would be nice to have some tests that don't rely on Mavericks as that adds a new layer of complexity that we shouldn't assume all epoxy users to be using. You can add non mavericks tests in a follow up!

epoxy-composeinterop-maverickssample/src/androidTest/java/com/airbnb/epoxy/composeinterop/maverickssample/MultiKeyComposeInteropFragmentTest.kt
```diff
@@ -0,0 +1,152 @@
...
+            waitUntil(textToAssert)
+            checkIfTextIsDisplayed(textToAssert)
+
+            composeTestRule.onNodeWithText(textToAssert).performClick()
+        }
+
+        waitUntil("withoutInteropKey mutableState: 11, recomposingCount: 12")
+        checkIfTextIsDisplayed("withoutInteropKey mutableState: 11, recomposingCount: 12")
+
+        Thread.sleep(2000)
```
vinaygaba, line 31: using this is not ideal but okay for this iteration

epoxy-composeinterop-maverickssample/src/androidTest/java/com/airbnb/epoxy/composeinterop/maverickssample/MultiKeyComposeInteropFragmentTest.kt
```diff
@@ -0,0 +1,152 @@
...
+    fun testComposableInInteropUsingMutableStateUpdatesEvenWithoutKey() {
+        for (i in 0..10) {
+            val textToAssert = "withoutInteropKey mutableState: $i, recomposingCount: ${i + 1}"
+            waitUntil(textToAssert)
+            checkIfTextIsDisplayed(textToAssert)
+
+            composeTestRule.onNodeWithText(textToAssert).performClick()
+        }
+
+        waitUntil("withoutInteropKey mutableState: 11, recomposingCount: 12")
```
vinaygaba, line 28: synced offline about how using waitUntil isn't ideal. 

`Any condition that depends on Android's measure or draw (that is, measure or draw external to Compose) should use more general concept such as waitUntil()`

vinaygaba (approved):
Looks good for a first iteration. Thank you for having the patience and incorporating all the feedback given to you

PR #1217, opened 2021-08-03
Use the SPDX identifier for the license in the POM

### LINKED ISSUE #1216: Use SDPX identifier for license?
I was trying out [Licensee](https://github.com/cashapp/licensee), and I noticed that it couldn't detect what license Epoxy was using out of the box.

This could be resolved by using the [SPDX identifier](https://spdx.org/licenses/) for the license name (`Apache-2.0`), which is [recommended by the POM reference](https://maven.apache.org/pom.html#Licenses). Right now it looks like `Apache License 2.0` is being used: https://github.com/airbnb/epoxy/blob/e4768fbccae58fb3b0178b4566925731c8483965/gradle.properties#L8

It doesn't seem like it would be hard to change. If there's interest in this, I could put up a PR for it.

rashadsookram:
Fixes #1216.

I tested this change by running `./gradlew generatePomFileForMavenPublication` to generate the POM files, then running `find . -name '*pom*' -exec grep '<license>' -A1 {} \;` and checking that the new license name was used in all the files. The output looked like 👇 repeated for each POM file.

```
    <license>
      <name>Apache-2.0</name>
```


elihart:
Thanks!

PR #1219, opened 2021-08-07
[COMPOSE] Added proper dispose support to Compose interop

vinaygaba:
In this PR, I updated `ComposeEpoxyModel` to properly dispose the composition by explicitly disposing on `unBind` and setting the appropriate view composition strategy on `ComposeView`. The default view composition strategy is `DisposeOnDetachedFromWindow ` and that keeps Composition instances in memory until the RecyclerView is detached from the window.  To make the Compose support work in all scenarios related to recyclerview's, it's necessary to use the DisposeOnViewTreeLifecycleDestroyed strategy.

@elihart @meggamind

### DIFF
--- a/blessedDeps.gradle
+++ b/blessedDeps.gradle
@@ -51,7 +51,7 @@ rootProject.ext.PARIS_VERSION = "1.7.3"
 rootProject.ext.ROBOLECTRIC_VERSION = "4.5.1"
 rootProject.ext.SQUARE_JAVAPOET_VERSION = "1.13.0"
 rootProject.ext.SQUARE_KOTLINPOET_VERSION = "1.8.0"
-rootProject.ext.COMPOSE_VERSION = "1.0.0-beta06"
+rootProject.ext.COMPOSE_VERSION = "1.0.0-beta07"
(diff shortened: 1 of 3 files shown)


blessedDeps.gradle
```diff
@@ -51,7 +51,7 @@ rootProject.ext.PARIS_VERSION = "1.7.3"
 rootProject.ext.ROBOLECTRIC_VERSION = "4.5.1"
 rootProject.ext.SQUARE_JAVAPOET_VERSION = "1.13.0"
 rootProject.ext.SQUARE_KOTLINPOET_VERSION = "1.8.0"
-rootProject.ext.COMPOSE_VERSION = "1.0.0-beta06"
+rootProject.ext.COMPOSE_VERSION = "1.0.0-beta07"
```
vinaygaba, line 54: Using beta08 because I'm not sure if Epoxy can be upgraded to 1.5.1 yet. If its okay to do that, I'll also bump this up to the stable compose version -  `1.0.0 `

meggamind (approved):

PR #1225, opened 2021-08-24
Add EpoxyModel#prebind hook

pterhx:
This adds a hook that is called before bind (`EpoxyModel#prebind`) which lets subclasses capture view state before any changes are made, allowing us to create transition animations.

@elihart

### DIFF
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModel.java
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModel.java
@@ -98,6 +98,51 @@ public abstract class EpoxyModel<T> {
   }
 
   /**
+   * Hook that is called before {@link #bind(Object)}. This is similar to
+   * {@link GeneratedModel#handlePreBind(EpoxyViewHolder, Object, int)}, but is intended for
+   * subclasses of EpoxyModel to hook into rather than for generated code to hook into.
+   * Overriding prebind is useful to capture state before the view changes, e.g. for animations.
+   */
+  public void prebind(@NonNull T view) {
+
+  }
+
+  /**
+   * Similar to {@link #prebind(Object)}, but provides a non null model which was previously bound
+   * to this view. This will only be called if the model is used with an {@link EpoxyController}.
+   *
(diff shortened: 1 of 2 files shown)


epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyViewHolder.java
```diff
@@ -60,6 +60,14 @@ public void bind(@SuppressWarnings("rawtypes") EpoxyModel model,
       ((GeneratedModel) model).handlePreBind(this, objectToBind(), position);
     }
 
+    if (previouslyBoundModel != null) {
+      // noinspection unchecked
+      model.prebind(objectToBind(), previouslyBoundModel);
```
elihart, line 65: add a test please

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModel.java
```diff
@@ -97,6 +97,51 @@ public View buildView(@NonNull ViewGroup parent) {
...
+   *                             reused. This means that you only need to update the view to reflect
+   *                             the data that changed. If you are using a custom item animator then
+   *                             the view will be the same if the animator returns true in
+   *                             canReuseUpdatedViewHolder.
+   *                             <p>
+   *                             This previously bound model is taken as a payload from the diffing
+   *                             process, and follows the same general conditions for all
+   *                             recyclerview change payloads.
+   */
+  public void prebind(@NonNull T view, @NonNull EpoxyModel<?> previouslyBoundModel) {
```
elihart, line 141: I think I prefer a single prebind function with a nullable previous model

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyViewHolder.java
pterhx: done!

epoxy-adapter/src/main/java/com/airbnb/epoxy/EpoxyModel.java
pterhx: done!

Issue #1199, opened 2021-06-14
Crash on Android 12


woodii:
According to Firebase Crashlytics users with Google Pixel Devices and Android 11 are experiencing crashes every time they start up our app.
I belive that that data is not correct and its actually Android 12 Beta as a few users stated that they can't start the app and I can't see any A12 non fatal errors logged aswell.

Unfortunately I can't reproduce the error on emulators or physical devices (including Pixel devices) running Android 11 or 12.

```
Fatal Exception: com.airbnb.epoxy.IllegalEpoxyUsage: You must set an id on a model before adding it. Use the @AutoModel annotation if you want an id to be automatically generated for you.
       at com.airbnb.epoxy.EpoxyController.addInternal(EpoxyController.java:502)
       at com.airbnb.epoxy.EpoxyModel.addTo(EpoxyModel.java:337)
```

The error is a bit missleading as I am setting an id via the builder pattern with `.id("***")` before calling `.addTo(this)`
But I have no idea what the real cause is.

epoxy Version: 4.6.1 (it's also happening on older versions e.g. 4.0.0-beta6)

johnpaulmckean:
I've been seeing this as well.

elihart:
Hm, we're not seeing this at all ourselves and I don't see how this code path could be affected by the OS version or how this would be possible if you are setting the id first. If you can give me more information that would help

johnpaulmckean:
Details I have from Bugsnag:
```
OS Versions: 11 (but based on users who reach out, it is 12 beta)
Manufacturers: Google
Models: [Pixel 3 XL, Pixel 3a XL, Pixel 4a, Pixel 5]
```

woodii:
Unfortunately there is nothing special about that epoxy usage.
The controller is placed inside a flexbox-layout in a fragment.
The crashing controller is one of many on the screen but probably the first one with models as the other ones rely on network-data.

More interestingly the crash occurred on one of our devices as well but not immediately after installing the app. It happend half a week later on appstart.

Just to emphasise the statistics above:
98% Android 11, and 2% Android 12 (imo its only 12 and just wrong reporting)
And distributed across various Pixel devices.

Just let me know if I can provide any specific information.

anurag-goyal:
To add to woodii's point. We are also experiencing the same issue on my app and the common thread is the Flexbox layout. We are also using the controller to places some views dynamically based on the server response.

thagikura:
[...]

PR #1234, opened 2021-09-24
https://github.com/airbnb/epoxy/issues/1199

blah1234:
This is for the crash on Android 12 reported in https://github.com/airbnb/epoxy/issues/1199 and also here: https://issuetracker.google.com/issues/197818595

- temporary workaround (https://issuetracker.google.com/issues/197818595#comment20) suggested by google engineer:
- In generated kotlin extension functions, store the created epoxy model in a local field before passing the object to ModelCollector.add()

### DIFF
--- a/epoxy-processor/src/main/java/com/airbnb/epoxy/processor/KotlinModelBuilderExtensionWriter.kt
+++ b/epoxy-processor/src/main/java/com/airbnb/epoxy/processor/KotlinModelBuilderExtensionWriter.kt
@@ -106,14 +106,15 @@ internal class KotlinModelBuilderExtensionWriter(
             addModifiers(KModifier.INLINE)
             addModifiers(if (constructorIsNotPublic) KModifier.INTERNAL else KModifier.PUBLIC)
 
-            addStatement("add(")
+            addStatement("val model = ")
             beginControlFlow(
                 "%T(${params.joinToString(", ") { it.name }}).apply",
                 modelClass
             )
             addStatement("modelInitializer()")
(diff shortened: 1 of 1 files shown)


elihart (approved):

rossbacher:
@blah1234 did you verify that this fixes the issue?

blah1234:
> @blah1234 did you verify that this fixes the issue?

We're going to push to our beta channel and see if the crashlytics incidence reports stop with this fix. Thanks so much for merging and following up!

rossbacher:
> > @blah1234 did you verify that this fixes the issue?
> 
> We're going to push to our beta channel and see if the crashlytics incidence reports stop with this fix. Thanks so much for merging and following up!

From looking at how the code gets compiled and how those extension functions create the issue I doubt this will work but please keep us updated. We are also trying a few things on our side if there is any way to work around this.

One way that works for sure is to disable optimization in r8. However you will be hot with an app size impact.

rossbacher:
@blah1234 I added some thoughts to https://github.com/airbnb/epoxy/issues/1199 do you have any update? I'll put a PR up to revert this, it does do no harm but it is also not necessary.

blah1234:
> @blah1234 I added some thoughts to #1199 do you have any update? I'll put a PR up to revert this, it does do no harm but it is also not necessary.

I'm out of the office this week, actually, so I've messaged my team to respond to this thread if they were able to do a push to the beta channel.

If not, I'll investigate and respond when I get back into the office next week. Thanks!

PR #1244, opened 2021-10-20
KSP support

elihart:
This adds support for Kotlin Symbol Processing, while maintaining backwards compatibility with java annotation processing via the xprocessing library from Room.

This includes a major version bump to 5.0.0 because there may be slight behavior differences with KSP, especially for generic types in generated code. For example, if you previously had an epoxy attribute in java source code with a raw type it may now appear in the generated code with a wildcard type, which may require tweaking the type that is passed to the model.

Additionally, some type checking was improved, for example more accurate validation of proper equals and hashcode implementations.

To use Epoxy with KSP, simply apply it with the ksp gradle plugin instead of kapt (https://github.com/google/ksp/blob/main/docs/quickstart.md). See the new epoxy-kspsample module for an example.

Note that unfortunately the databinding processor does NOT support KSP, simply because Android databinding itself uses KAPT and KSP cannot currently depend on KAPT sources. The code changes are in place to enable KSP with databinding once the databinding plugin from Android supports KSP (although this is unlikely) - alternatively it may be possible to configure the KSP plugin to run after KAPT and depend on its outputs (you're on your own if you want to try that).

Also, parallel processing support was removed because it is not compatible with KSP.

This PR is split into commits for "easier" reviewing, as much as is possible. The processor changes are the crux of the change. There is also a new ksp sample, and a large number of new tests using kotlin compile testing to test both ksp and kapt modes, also now with kotlin sources as inputs and outputs. The testing setup was refactored so that the existing tests would run on all processor types. I have also verified that this compiles correctly in our Airbnb app with all screenshot and interaction tests passing, so I have high confidence in the accuracy of the changes.


elihart:
I've already pushed this as a 5.0.0-beta01 release if anyone wants to try it

chrisjenx:
@elihart the ksp example doesn't use KSP fyi, and when you switch over to ksp the models aren't generated, not sure if that was an accident, but ksp isn't detecting the processor in these versions yet :(

elihart:
@chrisjenx thanks for the heads up, I was switching between kapt and ksp in that module to test. KSP does work if you apply the plugin, but the IDE may not show the references correctly because KSP does not support IDE code gen detection from kotlin sources (not automatically at least, it needs to be configured)

chrisjenx:
Tested on a couple of our modules, checked the generated sources and they
are missing. We're using deeplink dispatch and Moshi KSP, this is limited
to epoxy. Lemme check the binary, wondering if this is a packaging issue...

elihart:
What version are you on? I fixed the sample, and verified that `epoxy-kspsample:assembleDebug` succeeds locally and generates ksp sources

image

PR #1273, opened 2022-01-26
[paging3] Expose modelCache

mike-ghes:
Intended to support two use cases without resorting to reflection:

### "Load more" button

This was requested by our designers to leave room for footer content rather than infinite scrolling. Disabling automatic loading can be done just by removing call to super. And to load the next page, we just use `modelCache.loadAround()`:
```
override fun onModelBound(holder: EpoxyViewHolder, boundModel: EpoxyModel<*>, position: Int, previouslyBoundModel: EpoxyModel<*>?) {
    // Disable automatically loading pages by removing call to super
    // super.onModelBound(holder, boundModel, position, previouslyBoundModel)
}

override fun buildItemModel(currentPosition: Int, item: I?): EpoxyModel<*> {
    if (item == null) {
        return LoadMoreBindingModel_().apply {
            id("load more")
            onClick { _ -> modelCache.loadAround(currentPosition) }
        }
    }
    ...
}
```

### Invalidate item models

An item model may depends on both paging data from remote and some view state. We're able to manually rebuild EpoxyModels for individual items using the `modelCache.updateCallback.onChanged()`. 
```
var seletedItemIds = emptySet()

override fun buildItemModel(currentPosition: Int, item: ItemType?): EpoxyModel<*> {
    return ItemBindingModel_().apply {
        id(item.id)
        selected(seletedItemIds.contains(item.id))
        onClick { _ -> 
            selectedItemIds += item.id
            modelCache.updateCallback.onChanged(currentPosition, 1, null) 
        }
    }
}
```

### DIFF
--- a/epoxy-paging3/src/main/java/com/airbnb/epoxy/paging3/PagedDataModelCache.kt
+++ b/epoxy-paging3/src/main/java/com/airbnb/epoxy/paging3/PagedDataModelCache.kt
@@ -32,7 +32,7 @@ import kotlinx.coroutines.flow.Flow
  * that thread we allow a special case of cache modification when a new list is being submitted,
  * and all cache access is marked with @Synchronize to ensure safety when this happens.
  */
-internal class PagedDataModelCache<T : Any>(
+class PagedDataModelCache<T : Any>(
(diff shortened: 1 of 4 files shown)


elihart (commented):
Fine with me 👍

elihart (approved):

PR #1275, opened 2022-02-03
Add notifyModelChanged Payload

jakchang01:
The base code has no notifyModelChanged with payload, but sometimes we need to adopt payload.


elihart:
what is the specific use case? With epoxy controller any changes to model properties are already automatically delivered to the model in a payload when models are rebuilt with changes

PR #1292, opened 2022-04-27
Add support for Async Layout inflation with EPoxy

varungu:
EPoxy currently supports only synchronous layout inflation with most of the time during first layout is spent on the inflation of the views.

To make the first layout faster, we should support asynchronous layout inflation. With this approach, the application developer can make heavy views (like video player) asynchronous and improve the page load time

https://user-images.githubusercontent.com/2263278/165597664-3e3355bf-9296-4f78-9ab5-ef4f832099c7.mov

### DIFF
--- a/epoxy-adapter/src/main/java/com/airbnb/epoxy/BaseEpoxyAdapter.java
+++ b/epoxy-adapter/src/main/java/com/airbnb/epoxy/BaseEpoxyAdapter.java
@@ -108,6 +108,11 @@ public abstract class BaseEpoxyAdapter
 
   @Override
   public void onBindViewHolder(EpoxyViewHolder holder, int position, List<Object> payloads) {
+    executeWhenInflated(holder,
+        () -> onBindViewHolderInternal(holder, holder.getBindingAdapterPosition(), payloads));
+  }
+
+  private void onBindViewHolderInternal(EpoxyViewHolder holder, int position, List<Object> payloads) {
     EpoxyModel<?> modelToShow = getModelForPosition(position);
 
     EpoxyModel<?> previouslyBoundModel = null;
@@ -209,15 +214,19 @@ public abstract class BaseEpoxyAdapter
   @CallSuper
   @Override
   public void onViewAttachedToWindow(EpoxyViewHolder holder) {
-    //noinspection unchecked,rawtypes
-    ((EpoxyModel) holder.getModel()).onViewAttachedToWindow(holder.objectToBind());
+    executeWhenInflated(holder,
+        () ->
+            //noinspection unchecked,rawtypes
+            ((EpoxyModel) holder.getModel()).onViewAttachedToWindow(holder.objectToBind()));
   }
 
   @CallSuper
   @Override
   public void onViewDetachedFromWindow(EpoxyViewHolder holder) {
-    //noinspection unchecked,rawtypes
-    ((EpoxyModel) holder.getModel()).onViewDetachedFromWindow(holder.objectToBind());
+    executeWhenInflated(holder,
+        () ->
+            //noinspection unchecked,rawtypes
+            ((EpoxyModel) holder.getModel()).onViewDetachedFromWindow(holder.objectToBind()));
   }
 
   public void onSaveInstanceState(Bundle outState) {
@@ -339,4 +348,12 @@ public abstract class BaseEpoxyAdapter
   }
 
   //endregion
+
+  protected void executeWhenInflated(EpoxyViewHolder holder, Runnable runnable) {
+    if (holder.itemView instanceof AsyncFrameLayout) {
+      ((AsyncFrameLayout)holder.itemView).executeWhenInflated(runnable);
+    } else {
+      runnable.run();
+    }
+  }
 }
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,4 @@
-VERSION_NAME=5.0.0-beta05
+VERSION_NAME=5.0.0-beta06
 GROUP=com.airbnb.android
 POM_DESCRIPTION=Epoxy is a system for composing complex screens with a ReyclerView in Android.
 POM_URL=https://github.com/airbnb/epoxy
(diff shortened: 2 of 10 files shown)


epoxy-adapter/src/main/java/com/airbnb/epoxy/AsyncInflatedView.kt
```diff
@@ -0,0 +1,43 @@
...
+        }
+        pendingRunnables.clear()
+    }
+
+    /**
+     * executeWhenInflated method is called by EPoxy to execute a runnable that depend on the
+     * inflated view. If the view is already inflated, the runnable will immediately run,
+     * otherwise it is added to the list of pending runnables.
+     */
+    fun executeWhenInflated(runnable: Runnable) {
```
pmecho, line 36: Do we need a way of clearing these pending executions when the view is recycled?

epoxy-adapter/src/main/java/com/airbnb/epoxy/BaseEpoxyAdapter.java
```diff
@@ -339,4 +348,12 @@ public boolean isStickyHeader(int position) {
   }
 
   //endregion
+
+  protected void executeWhenInflated(EpoxyViewHolder holder, Runnable runnable) {
+    if (holder.itemView instanceof AsyncInflatedView) {
```
pmecho, line 353: @varungu Have we considered making the async layout a property of the `EpoxyModel` or `EpoxyViewHolder` so we can avoid needing to do an `instanceof` check here?

epoxy-adapter/src/main/java/com/airbnb/epoxy/AsyncFrameLayout.kt
```diff
@@ -0,0 +1,17 @@
+package com.airbnb.epoxy
+
+import android.content.Context
+import android.util.AttributeSet
+import android.widget.FrameLayout
+
+/**
+ * Base class to support Async layout inflation with EPoxy.
```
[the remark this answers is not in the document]
varungu: Done.

epoxy-adapter/src/main/java/com/airbnb/epoxy/AsyncInflatedView.kt
```diff
@@ -0,0 +1,43 @@
+package com.airbnb.epoxy
+
+/**
+ * Interface to support Async layout inflation with EPoxy
```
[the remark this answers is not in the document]
varungu: Done.

epoxy-adapter/src/main/java/com/airbnb/epoxy/AsyncInflatedView.kt
```diff
@@ -0,0 +1,43 @@
...
+     * pendingRunnables keep a list of runnables, in order, that are waiting for the view to be
+     * inflated.
+     */
+    var pendingRunnables : ArrayList<Runnable>
+
+    /**
+     * onInflationComplete method that MUST be called after the view is asyncronously inflated.
+     * It runs all pending runnables waiting for view inflation.
+     */
+    fun onInflationComplete() {
```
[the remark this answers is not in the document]
varungu: I think so, we should keep it optional to keep the library as small as possible.

gradle.properties
```diff
@@ -1,4 +1,4 @@
-VERSION_NAME=5.0.0-beta05
+VERSION_NAME=5.0.0-beta06
```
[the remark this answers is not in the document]
varungu: Done

epoxy-adapter/src/main/java/com/airbnb/epoxy/AsyncInflatedView.kt
varungu: Good catch. Done.

epoxy-adapter/src/main/java/com/airbnb/epoxy/BaseEpoxyAdapter.java
varungu: I have a TODO for this in the description. It is not very straight forward to make it a property. Biggest problem is all views will need to extend from this interface, that we don't want to do. If we don't enforce the extension, we will need the check in addition to the property.

My plan is to test this change before putting effort into adding a property.

pmecho: @varungu I just realized an additional concern with this approach. On each bind and attach/detach from the window this will do an object allocation of the runnable which may cause some performance hits for those not using async layouts. I think we'd probably want to better integrate this into the model or view holder before merging.

varungu: We can separate this and create runnable only for AsyncViews, would that resolve your concern?

varungu: Done.

epoxy-adapter/src/main/java/com/airbnb/epoxy/BaseEpoxyAdapter.java
```diff
@@ -108,6 +108,15 @@ public void onBindViewHolder(EpoxyViewHolder holder, int position) {
 
   @Override
   public void onBindViewHolder(EpoxyViewHolder holder, int position, List<Object> payloads) {
+    if (holder.itemView instanceof AsyncInflatedView) {
```
vinaygaba, line 111: I think Peter might've mentioned this earlier and I'm not completely familiar with the internals of Epoxy but is there a way where we could avoid these checks? We discussed an implementation where we potentially create an `AsycEpoxyModel` class (similar to how we handled it for Compose interoperability - https://github.com/airbnb/epoxy/blob/master/epoxy-compose/src/main/java/com/airbnb/epoxy/ComposeInterop.kt)

elihart:
@varungu @vinaygaba can we merge this?

vinaygaba:
@elihart I think Varun is running a test with these changes.

PR #1309, opened 2022-08-10
[ComposeInterop] EpoxyInterop now handles @AfterPropSet ModelView methods

f-noriega:
Currently `EpoxyInterop` won't call any `@AfterPropSet` method in the view. This fixes that by calling `handlePostBind` after the view has been bound.

Modified example to show alternating title colors with `@AfterPropSet fun changeTitleColor()`

Alternating title color.

| Before | After |
|---|---|
| image | image |


vinaygaba (approved):
Great find!

PR #1315, opened 2022-09-23
[ComposeInterop] Exposes Compose Epoxy Model generation

pete32:
There are instances where we need to have finer control over the Compose generated Epoxy model before it's added, so this change exposes the function which generates the model.

This new function can be used directly instead of `composeInterop {}` when want to override Epoxy model parameters on the Compose generated model. Eg.  overriding row spans like: model.spanSizeOverride(3).

### DIFF
--- a/epoxy-compose/src/main/java/com/airbnb/epoxy/ComposeInterop.kt
+++ b/epoxy-compose/src/main/java/com/airbnb/epoxy/ComposeInterop.kt
@@ -69,14 +69,14 @@ fun ModelCollector.composableInterop(
     vararg keys: Any,
     composeFunction: @Composable () -> Unit
 ) {
-    add(generateComposeEpoxyModel(id, *keys, composeFunction = composeFunction))
+    add(composeEpoxyModel(id, *keys, composeFunction = composeFunction))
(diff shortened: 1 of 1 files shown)


epoxy-compose/src/main/java/com/airbnb/epoxy/ComposeInterop.kt
```diff
@@ -69,11 +69,21 @@ fun ModelCollector.composableInterop(
...
-        }
-    )
+    add(generateComposeEpoxyModel(id, *keys, composeFunction = composeFunction))
+}
+
+/**
+ * [generateComposeEpoxyModel] can be used directly in cases where more control over the epoxy model
+ * is needed. Eg. When the epoxy model needs to be modified before it's added.
+ */
+fun generateComposeEpoxyModel(
```
vinaygaba, line 79: you can just call it `composeEpoxyModel`

pete32: Updated the name

vinaygaba (approved):

PR #1316, opened 2022-09-29
Prepare for release 5.0.0

elihart:
# 5.0.0
This adds support for Kotlin Symbol Processing, while maintaining backwards compatibility with java annotation processing via the xprocessing library from Room.

This includes a major version bump to 5.0.0 because there may be slight behavior differences with KSP, especially for generic types in generated code. For example, if you previously had an epoxy attribute in java source code with a raw type it may now appear in the generated code with a wildcard type, which may require tweaking the type that is passed to the model.

Additionally, some type checking was improved, for example more accurate validation of proper equals and hashcode implementations.

To use Epoxy with KSP, simply apply it with the ksp gradle plugin instead of kapt (https://github.com/google/ksp/blob/main/docs/quickstart.md). See the new epoxy-kspsample module for an example.

Note that unfortunately the databinding processor does NOT support KSP, simply because Android databinding itself uses KAPT and KSP cannot currently depend on KAPT sources. The code changes are in place to enable KSP with databinding once the databinding plugin from Android supports KSP (although this is unlikely) - alternatively it may be possible to configure the KSP plugin to run after KAPT and depend on its outputs (you're on your own if you want to try that).

Also, parallel processing support was removed because it is not compatible with KSP.

We have also added easy interop with Jetpack Compose via functions in the `epoxy-composeinterop` artifact.
See the epoxy-composesample module for example usage.


eboudrant:
@elihart, with KSP, does the option `epoxyDisableDslMarker` not supported anymore? After updating to KSP we have lot of error related to the DSL marker.

elihart:
Strange, we have epoxyDisableDslMarker enabled with ksp and haven't had a problem. are you sure you are setting annotation options correctly with ksp?

eboudrant:
Ok, good to know it should still work, I'll troubleshoot.

We're seeing this in the log : 
```
> Task :ui:epoxy-models:api:kaptDebugKotlin
warning: The following options were not recognized by any processor: '[epoxyDisableDslMarker, requireAbstractEpoxyModels, validateEpoxyModelUsage, enableParallelEpoxyProcessing, requireHashCodeInEpoxyModels, kapt.kotlin.generated, logEpoxyTimings]'
```
Not sure if this is kapt or ksp output, looks like kapt.

We're using the argument in a convention plugin, it was working with katp, according the `epoxy-kspsample` it should work with ksp too 🤔 

```
            with(checkNotNull(androidExtension)) {
                buildTypes.configureEach {
                    javaCompileOptions.annotationProcessorOptions.arguments.let { arguments ->
                        arguments["validateEpoxyModelUsage"] = "${name.contains("debug")}"
                        arguments["epoxyDisableDslMarker"] = "true"
                        arguments["requireHashCodeInEpoxyModels"] = "true"
                        arguments["requireAbstractEpoxyModels"] = "true"
                        arguments["logEpoxyTimings"] = "true"
                    }
                }
            }
```

eboudrant:
If I use this code in our convention plugin then it works 🤷 
```
            project.configure<KspExtension> {
                arg("validateEpoxyModelUsage", "${name.contains("debug")}")
                arg("epoxyDisableDslMarker", "true")
                arg("requireHashCodeInEpoxyModels", "true")
                arg("requireAbstractEpoxyModels", "true")
                arg("logEpoxyTimings", "true")
            }
```

elihart:
The `javaCompileOptions.annotationProcessorOptions` does not work to pass arguments to ksp. like you said, the `ksp` block needs to be used instead https://kotlinlang.org/docs/ksp-quickstart.html#pass-options-to-processors

in the sample you mentioned, i probably just forgot to uncomment the ksp options when switching between kapt and ksp set ups 
https://github.com/airbnb/epoxy/blob/master/epoxy-kspsample/build.gradle#L42

eboudrant:
I see, and the Epoxy timing I was seeing in the sample build console were actually coming from `epoxy-adapter`, not the sample.

PR #1321, opened 2022-10-21
Please allow Kotlin classes to be subclassed with `open` keyword

blah1234:
I am upgrading from 4.2.0.  My existing codebase makes use of subclassing and method overrides.  With the refactor from java to kotlin, the following need to be marked `open` in order for my codebase to compile.

https://kotlinlang.org/docs/inheritance.html


elihart (approved):
fine with me 👍

PR #1322, opened 2022-11-01
Adding option to provide span size for compose epoxy interop

pranayairan:
Currently there is no option to provide span size to compose epoxy interop. This creates issues when the default layout manager is grid but you want to override the epoxy model behavior at runtime to change the span size. By default the span size is always set to 1 if not overridden. 

I added a sample fragment showing how to use the new changes. 

without_override
with_override


pranayairan:
This change is not required any more, we can use changes introduced in this PR to get epoxy models and customize there behavior https://github.com/airbnb/epoxy/pull/1315

PR #1327, opened 2022-11-15
Fix click listener kapt bug 

elihart:
This bug incorrectly generates click listener code.

Also cleans up how publishing is done

### DIFF
--- a/epoxy-processor/src/main/java/com/airbnb/epoxy/processor/Type.kt
+++ b/epoxy-processor/src/main/java/com/airbnb/epoxy/processor/Type.kt
@@ -43,9 +43,11 @@ class Type(val xType: XType, memoizer: Memoizer) {
                     xType.isTypeOf(CharSequence::class) || xType.isTypeOf(String::class) -> StringOrCharSequence
                     xType.typeName == ClassNames.EPOXY_STRING_ATTRIBUTE_DATA -> StringAttributeData
                     // We don't care about nullability for the purposes of type checking
-                    nonNullType == memoizer.viewOnClickListenerType -> ViewClickListener
-                    nonNullType == memoizer.viewOnLongClickListenerType -> ViewLongClickListener
-                    nonNullType == memoizer.viewOnCheckChangedType -> ViewCheckedChangeListener
+                    // Note, == does not work for type comparisons when comparing types between classpath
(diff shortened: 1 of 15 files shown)


publishing.gradle
```diff
@@ -0,0 +1,29 @@
+// Sets up publishing of release artifacts.
+// Note: Keep this script in sync across all airbnb open source projects.
+apply plugin: 'com.vanniktech.maven.publish'
+
+
+// https://vanniktech.github.io/gradle-maven-publish-plugin/other/
+publishing {
```
vinaygaba, line 7: Thank you for this file, I ended up using this for Showkase publishing as well 🙏

epoxy-processor/src/main/java/com/airbnb/epoxy/processor/Type.kt
```diff
@@ -43,9 +43,11 @@ class Type(val xType: XType, memoizer: Memoizer) {
                     xType.isTypeOf(CharSequence::class) || xType.isTypeOf(String::class) -> StringOrCharSequence
                     xType.typeName == ClassNames.EPOXY_STRING_ATTRIBUTE_DATA -> StringAttributeData
                     // We don't care about nullability for the purposes of type checking
-                    nonNullType == memoizer.viewOnClickListenerType -> ViewClickListener
-                    nonNullType == memoizer.viewOnLongClickListenerType -> ViewLongClickListener
-                    nonNullType == memoizer.viewOnCheckChangedType -> ViewCheckedChangeListener
+                    // Note, == does not work for type comparisons when comparing types between classpath
```
vinaygaba, line 46: any chance you could maybe describe the bug a bit more in the PR description. Just helps me understand a bit better. I can follow that the equals check wasn't what was needed but didn't understand what the difference in the generated code looked like

vinaygaba: is there any test case that could be added?

vinaygaba (approved):

epoxy-processor/src/main/java/com/airbnb/epoxy/processor/Type.kt
elihart: hard to test because this happens when they are in different modules

PR #1337, opened 2023-01-18
Resolve unchecked call warning for WrappedEpoxyModelClickListener

SubhrajyotiSen:
Generated classes that contain a View.ClickListener attribute has a method similar to the following:

```java
public DateViewHolder_ clickListener(
      @NonNull final OnModelClickListener<DateViewHolder_, ViewBindingHolder> clickListener) {
    onMutation();
    if (clickListener == null) {
      super.setClickListener(null);
    }
    else {
      super.setClickListener(new WrappedEpoxyModelClickListener(clickListener));
    }
    return this;
  }
```

The code `new WrappedEpoxyModelClickListener(clickListener)` causes a warning similar to 

```bash
 warning: [unchecked] unchecked call to WrappedEpoxyModelClickListener(OnModelClickListener<T,V>) as a member of the raw type WrappedEpoxyModelClickListener
            super.setClickListener(new WrappedEpoxyModelClickListener(clickListener));
                                   ^
  where T,V are type-variables:
    T extends EpoxyModel<?> declared in class WrappedEpoxyModelClickListener
    V extends Object declared in class WrappedEpoxyModelClickListener
```

This PR resolves the warning by writing `super.setClickListener(new WrappedEpoxyModelClickListener<>(clickListener));` to the generated code instead of `super.setClickListener(new WrappedEpoxyModelClickListener(clickListener));`

Also, `XTypeElement.superType` is Deprecated. Hence replacing it with `XTypeElement.superClass`

### DIFF
--- a/epoxy-modelfactorytest/src/test/resources/AllTypesModelViewModel_.java
+++ b/epoxy-modelfactorytest/src/test/resources/AllTypesModelViewModel_.java
@@ -528,7 +528,7 @@ public class AllTypesModelViewModel_ extends EpoxyModel<AllTypesModelView> imple
       this.onClickListener_OnClickListener = null;
     }
     else {
-      this.onClickListener_OnClickListener = new WrappedEpoxyModelClickListener(onClickListener);
+      this.onClickListener_OnClickListener = new WrappedEpoxyModelClickListener<>(onClickListener);
     }
     return this;
   }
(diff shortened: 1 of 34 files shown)


SubhrajyotiSen:
@elihart Tagging you here since I'm unable to request reviews on this PR. I hope that's okay.

elihart:
Thanks for the fix!

PR #1347, opened 2023-05-09
epoxy-kspsample : use ksp block to specify arguments

eboudrant:
As discussed here https://github.com/airbnb/epoxy/pull/1316#issuecomment-1540983979


elihart (approved):
thanks!

PR #1368, opened 2023-11-16
Dispose composition of `ComposeView` when unbinding `ComposeEpoxyModel`

ubuntudroid:
Right now the composition of `ComposeView` in `ComposeEpoxyModel` isn't properly disposed. This leads to issues where e.g. when navigating to away from the screen where the `ComposeEpoxyModel` is rendered, and then coming back, the model isn't rendered anymore.

This PR fixes this by calling `ComposeView.disposeComposition()` when unbinding the `ComposeView` from the model.

### DIFF
--- a/epoxy-compose/src/main/java/com/airbnb/epoxy/ComposeInterop.kt
+++ b/epoxy-compose/src/main/java/com/airbnb/epoxy/ComposeInterop.kt
@@ -46,6 +46,11 @@ class ComposeEpoxyModel(
         view.setContent(composeFunction)
     }
 
+    override fun unbind(view: ComposeView) {
+        super.unbind(view)
+        view.disposeComposition()
+    }
+
     override fun equals(other: Any?): Boolean {
         if (other === this) return true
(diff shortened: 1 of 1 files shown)


StephenVinouze:
Thxs a lot for your investigation @ubuntudroid! I've just upgraded the lib to benefit from the interop wrapper and I couldn't figure out why some items weren't rendered anymore after navigating away then back to the screen. As if the recomposition didn't get triggered. Since it's not systematic, not easy to debug. Unbinding does the trick!
Would love to see this merge to use the compose dependency. It's ok as the code is contained within a class though.

StephenVinouze:
Also, I understand the support isn't as expected when it comes to Compose since LazyList should be a natural replacement candidate. But we still appreciate the support, especially with compose interop, as migrating large codebase raises bigger issues. And LazyList at its current state does not support everything you can do with a RecyclerView (and Epoxy by extension). I'm thinking about item animations (move is doable but not add/remove although it should come in Compose 1.7) + MotionLayout interop too. Thxs a lot for keeping the support :)

